PR #2, opened 2018-09-17
Make rsyslog__weight_map consistent.

nhosoi:
Add a missing variable rsyslog__capabilities to README.

### DIFF
--- a/roles/rsyslog/README.md
+++ b/roles/rsyslog/README.md
@@ -130,6 +130,8 @@ rsyslog__user: root
 - `rsyslog__viaq` : When 'True' rsyslog role will deploy the viaq configuration set.  In this case, rsyslog works as a collector of OpenShift logs, normalizes them, then sends to the ElasticSearch.  Default to 'False'.
 - `rsyslog__example` : When 'True' rsyslog role will deploy the example configuration set.  Default to 'False'.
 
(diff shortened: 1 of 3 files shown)


roles/rsyslog/README.md
```diff
@@ -130,6 +130,8 @@ rsyslog__user: root
 - `rsyslog__viaq` : When 'True' rsyslog role will deploy the viaq configuration set.  In this case, rsyslog works as a collector of OpenShift logs, normalizes them, then sends to the ElasticSearch.  Default to 'False'.
 - `rsyslog__example` : When 'True' rsyslog role will deploy the example configuration set.  Default to 'False'.
 
+- `rsyslog__capabilities` : List of capabilities to configure.  [ 'network', 'remote-files', 'tls', 'viaq', 'viaq-k8s' ] are predefined.
```
portante: Seems like these two changes are orthogonal, is it worth having separate commits?

portante: So I am just reviewing these roles to see if we could use them, so not so sure yet if I understand everything to really comment well.  It just seems that the README change is independent of the other change.

Having to commits in one PR is okay, so that when they are merged you can see the two different changes.  We use the "rebase and merge" option often so that we can maintain two commits together on top of `master`, and use "squash and merge" in other cases where they commits are related and really should land as one after a review.  That kind of policy is something this repo needs to arrive at from the community.

Issue #6, opened 2018-10-04
Should be an "or" condition instead of "and"


sradco:
https://github.com/linux-system-roles/logging/blob/a72b50f6f686fb071d84d62e78bd6a6df456f79a/roles/rsyslog/tasks/main.yaml#L134

@nhosoi Hi, Should the condition here be with an "or" instead of "and"?

sradco:
Doesn't  item.state|d() mean default is empty? I think for is defined it should be item.state is defined.

nhosoi:
Thank you, @sradco !  Definitely, there was a bug...
Could your review #7 including this commit?
https://github.com/linux-system-roles/logging/pull/7/commits/cd28ff8473fbf8e073adce00093a78f35a2152c2
I tested starting with:
```
rsyslog__example: true
rsyslog__capabilities: [ 'network', 'remote-files', 'tls' ]
```
to
```
rsyslog__viaq: true
rsyslog__capabilities: [ 'viaq', 'viaq-k8s' ]
```
to
```
rsyslog__viaq: true
rsyslog__capabilities: [ 'viaq' ]
```
then none.  I verified that by running the ansible command, necessary files are created and unnecessary files are removed based on the combination of rsyslog config type and capabilities.

nhosoi:
@sradco , is it ok to close this issue?

sradco:
Fixed.

PR #7, opened 2018-10-04
Bug Fixes

nhosoi:
Thanks to @sradco for reviews.  These are fixes based on her comments.

README.md
- Sample inventory file
  - Replacing deprecated ansible_ssh_user with ansible_user
  - Having ${RSYSLOG_ANSIBLE_SSH_USER} for ansible_user in the inventory
    file example is not appropriate. Replacing it with more specific string.
- Sample vars.yaml
  - Adding elasticsearch_server_host and elasticsearch_server_port.
  - Adding a section for copying custom config files by rsyslog__custom_config_files.

defaults/main.yaml
- Setting rsyslog__capabilities to [].  Setting capabilities to add a capability.
- Using 'viaq' capability explicitly.
- Deploying crio.rulebase only when 'viaq-k8s' is in rsyslog__capabilities.

### DIFF
--- a/roles/rsyslog/defaults/main.yaml
+++ b/roles/rsyslog/defaults/main.yaml
@@ -35,7 +35,8 @@ rsyslog__work_dir: '/var/lib/rsyslog'
 # .. envvar:: rsyslog__capabilities
 #
 # List of different capabilities to configure. See :ref:`rsyslog__capabilities`
-rsyslog__capabilities: [ 'network', 'remote-files', 'tls', 'viaq', 'viaq-k8s' ]
+# rsyslog__capabilities: [ 'network', 'remote-files', 'tls', 'viaq', 'viaq-k8s' ]
+rsyslog__capabilities: []
 
 # .. envvar:: rsyslog__unprivileged
 #
@@ -818,6 +819,9 @@ rsyslog__conf_viaq_main_modules:
 
   - name: 'viaq_main'
     type: 'modules'
+    state: "{{ 'present'
+               if ('viaq' in rsyslog__capabilities)
+               else 'absent' }}"
     sections:
 
       - options: |-
@@ -849,6 +853,9 @@ rsyslog__conf_viaq_mmk8s:
   - name: 'mmk8s'
     type: 'modules'
     path: '{{rsyslog__viaq_config_dir }}'
+    state: "{{ 'present'
+               if ('viaq-k8s' in rsyslog__capabilities)
+               else 'absent' }}"
     sections:
 
       - options: |-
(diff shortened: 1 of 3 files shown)


nhosoi:
@sradco , @richm , could you please review this pr?  Thanks!

roles/rsyslog/defaults/main.yaml
```diff
@@ -872,15 +879,15 @@ rsyslog__conf_viaq_mmk8s:
...
-        state: '{{ "present"
-               if ("viaq-k8s" in rsyslog__capabilities)
-               else "absent" }}'
 
 rsyslog__conf_viaq_elasticsearch:
 
   - name: 'elasticsearch'
     type: 'output'
     path: '{{rsyslog__viaq_config_dir }}'
+    state: "{{ 'present'
```
richm: So ovirt will use the `viaq` capability?
There will be other non-viaq users of the elasticsearch output

roles/rsyslog/defaults/main.yaml
```diff
@@ -1299,6 +1309,9 @@ rsyslog__rulebase_viaq_k8s_filename:
     filename: 'k8s_filename.rulebase'
     nocomment: 'true'
     path: '{{rsyslog__viaq_config_dir }}'
+    state: "{{ 'present'
+               if ('viaq' in rsyslog__capabilities)
```
richm: this file is only used with `viaq-k8s`

roles/rsyslog/defaults/main.yaml
```diff
@@ -1323,6 +1339,9 @@ rsyslog__rulebase_viaq_k8s_container_name:
     filename: 'k8s_container_name.rulebase'
     nocomment: 'true'
     path: '{{rsyslog__viaq_config_dir }}'
+    state: "{{ 'present'
+               if ('viaq' in rsyslog__capabilities)
```
richm: this file is only used with `viaq-k8s`

roles/rsyslog/defaults/main.yaml
nhosoi: Shall we add another capability like `elasticsearch` to distinguish from viaq?

roles/rsyslog/defaults/main.yaml
nhosoi: Oops.  Thanks!

roles/rsyslog/defaults/main.yaml
nhosoi: Thanks, again.

roles/rsyslog/defaults/main.yaml
richm: hmm - actually, this elasticsearch output is specific to viaq `rsyslog__conf_viaq_elasticsearch` - so it should be specific to viaq - we'll have to add another elasticsearch output that is more generic in the future

nhosoi: Ok.  I'll leave this output plugin as is for now and open an issue with your comment.  Thanks.

nhosoi:
Thanks, @richm !
I'm cleaning up some confusions pointed by @sradco .  Please arrow me one more commit.
https://github.com/linux-system-roles/logging/issues/6

Issue #12, opened 2018-10-29
Issues during installation of rsyslog example role on fedora 28


sradco:
I run the rsyslog role on Fedora28.
I run it with the example rsyslog vars.yaml.
```
rsyslog__enabled: true
# install example packages & config files
rsyslog__example: true
rsyslog__capabilities: [ 'network', 'remote-files', 'tls' ]
rsyslog__forward: [ '' ]
rsyslog__group: root
rsyslog__user: root
```
I had a few issues:

1. I was missing the python2-libselinux package
TASK [rsyslog : Generate main rsyslog configuration] *****************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "checksum": "6d5d22e905ff75f31050d24cc0adb3ca18f57430", "msg": "Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!"}

2. TASK [rsyslog : Install/Update required packages] ********************************************************************************************************************************************
[DEPRECATION WARNING]: Invoking "yum" only once while using a loop via squash_actions is deprecated. Instead of using a loop to supply multiple items and specifying `name: {{ item }}`, 
please use `name: [u'{{ rsyslog__base_packages }}', u'{{ rsyslog__viaq_prereq_packages if rsyslog__viaq|bool else [] }}', u'{{ rsyslog__viaq_packages if rsyslog__viaq|bool else [] }}', u'{{
 rsyslog__tls_packages if rsyslog__pki|bool else [] }}', u'{{ rsyslog__packages }}']` and remove the loop. This feature will be removed in version 2.11. Deprecation warnings can be disabled
 by setting deprecation_warnings=False in ansible.cfg.

3. TASK [rsyslog : Moving the contents of /etc/rsyslog.d to the backup dir] *********************************************************************************************************************
 [WARNING]: Consider using the unarchive module rather than running tar.  If you need to use command because unarchive is insufficient you can add warn=False to this command task or set
command_warnings=False in ansible.cfg to get rid of this message.

@richm @nhosoi

nhosoi:
Regarding 1:
[...]

Issue #15, opened 2018-10-29
Replace double underlines(__) with only a single underline (_)


sradco:
I think we should replace the double underlines(__) to have only a single underline (_).
It is not standard in the linux-system-roles. Is there any special reason this was left as is?

@richm @nhosoi @pcahyna

nhosoi:
Please see https://github.com/linux-system-roles/logging/pull/20 for this issue, as well.

Issue #13, opened 2018-10-29
rsyslog - Don't move all confs to backup  by default when rerunning rsyslog role


sradco:
Currently when running rsyslog role we move all conf files to a backup dir and deploy new confs.

We may want to adopt this kind of approach: https://github.com/linux-system-roles/selinux#purge-local-modifications.

That is only purge the files if requested by the user.
This should allow rerunning of the role and have it immutable.

@richm @nhosoi @nkinder

nhosoi:
Please take a look at https://github.com/linux-system-roles/logging/pull/20?

Does "Defining rsyslog_purge_original_conf to manage pre-existing config files to keep or purge" satisfy the needs?

Issue #8, opened 2018-10-06
[RFE] rsyslog - need to add a generic elasticsearch output


nhosoi:
Comment by @richm :
> actually, this elasticsearch output is specific to viaq rsyslog__conf_viaq_elasticsearch - so it should be specific to viaq - we'll have to add another elasticsearch output that is more generic in the future

nhosoi:
@sradco, 
This is the first cut for this issue (e.g., no doc has been added yet).
Do you think this satisfies the needs in RHV?
https://github.com/linux-system-roles/logging/pull/20/commits/29e3353d2860574f9c69baf16f9d380cb73996f6

sradco:
I see you updated rsyslog_elastcsearch to rsyslog_elastcsearch_viaq. I think this RFE is still relevant since this does not handle generics elasticsearch output.
@nhosoi @richm

sradco:
Fixed in PR #27.

PR #20, opened 2018-10-30
Updating the rsyslog roles based on the reviews by @sradco

nhosoi:
1. Fixing "DEPRECATION WARNING" from the TASK "Install/Update required packages"
    Avoid using loop, devide each package array and assign it to its own task.
    (Unfortunately, it made the code uglier...)

2. Defining rsyslog_purge_original_conf to manage pre-existing config files to keep or purge.

3. Adding 'libselinux-python' to rsyslog_base_packages to avoid the installation
    failure due to the missing package.  
    (Not sure this is the right package to be in the list.)

4. Replacing rsyslog__ with rsyslog_

### DIFF
--- a/roles/rsyslog/README.md
+++ b/roles/rsyslog/README.md
@@ -1,8 +1,31 @@
+linux-system-roles-rsyslog
+==========================
+
 # Guidelines for Using Rsyslog Ansible Roles
 
 This is the ansible rsyslog roles to deploy configuration files.
 
-## How to Deploy Configuration Files Using Rsyslog Ansible Roles
+Table of Contents
+=================
+
+<!--ts-->
+   * [linux-system-roles-rsyslog](#linux-system-roles-rsyslog)
+   * [Table of contents](#table-of-contents)
+   * [Deploy Rsyslog Configuration Files](#deploy)
+      * [Inventory File](#inventory-file)
+      * [vars.yaml](#vars)
+      * [playbook.yaml](#playbook)
+   * [Deployed Results](#results)
+      * [rsyslog_viaq](#rsyslog_viaq)
+      * [rsyslog_example](#rsyslog_example)
+   * [Variables in vars.yaml](#variables)
+      * [Common sub-variables](#common-variables)
+      * [Viaq sub-variables](#viaw)
+   * [Contents of Roles](#roles)
+<!--te-->
+
+Deploy Rsyslog Configuration Files
+==================================
 
 Typical ansible-playbook command line.
 
@@ -10,7 +33,9 @@ Typical ansible-playbook command line.
 
 Two files - inventory_file and vars.yaml - in the command line is to be updated by the user.
 
-1. inventory_file is used to specify the hosts to deploy the configuration files.
+Inventory File
+--------------
+inventory_file is used to specify the hosts to deploy the configuration files.
 
    Sample inventory file for the es-ops enabled case
 ```
@@ -21,53 +46,96 @@ localhost ansible_user=YOUR_ANSIBLE_USER openshift_logging_use_ops=True
 localhost ansible_user=YOUR_ANSIBLE_USER openshift_logging_use_ops=True
 ```
 
-2. vars.yaml stores variables which are passed to ansible to control the tasks.
+vars.yaml
+---------
+1. vars.yaml stores variables which are passed to ansible to control the tasks.
 
-   Currently, this rsyslog roles support 2 deployment variables, rsyslog__viaq and rsyslog__example.  I.e., there are 3 sets of deployment - rsyslog__viaq: true, rsyslog__example: true, and both false.  Theoretically, both true could be set and the specified configuration files are deployed, but rsyslog does not work properly with the configuration.
+   Currently, this rsyslog roles support 2 deployment variables, rsyslog_viaq and rsyslog_example.  I.e., there are 3 sets of deployment - rsyslog_viaq: true, rsyslog_example: true, and both false.  Theoretically, both true could be set and the specified configuration files are deployed, but rsyslog does not work properly with the configuration.
 
    See the variables section for each variable.
 
    Sample vars.yaml file for the viaq case.
 ```
-rsyslog__enabled: true
+rsyslog_enabled: true
 # install viaq packages & config files
-rsyslog__viaq: true
-rsyslog__capabilities: [ 'viaq' ]
-rsyslog__group: root
-rsyslog__user: root
+rsyslog_viaq: true
+rsyslog_capabilities: [ 'viaq' ]
+rsyslog_group: root
+rsyslog_user: root
 elasticsearch_server_host: es_hostname
 elasticsearch_server_port: 9200
 ```
 
-2.1  vars.yaml to configure to handle the inputs from openshift containers
+2. vars.yaml to configure to handle the inputs from openshift containers
 ```
-rsyslog__enabled: true
+rsyslog_enabled: true
 # install viaq packages & config files
-rsyslog__viaq: true
-rsyslog__capabilities: [ 'viaq', 'viaq-k8s' ]
-rsyslog__group: root
-rsyslog__user: root
-# If 'viaq-k8s' is in rsyslog__capabilities, logging_mmk8s_* need to be specified.
-logging_mmk8s_token: "{{rsyslog__viaq_config_dir}}/mmk8s.token"
-logging_mmk8s_ca_cert: "{{rsyslog__viaq_config_dir}}/mmk8s.ca.crt"
-# If use_omelasticsearch_cert is True, logging_elasticsearch_* need to be specified.
+rsyslog_viaq: true
+rsyslog_capabilities: [ 'viaq', 'viaq-k8s' ]
+rsyslog_group: root
+rsyslog_user: root
+# If 'viaq-k8s' is in rsyslog_capabilities, logging_mmk8s_* need to be specified.
+logging_mmk8s_token: "{{rsyslog_viaq_config_dir}}/mmk8s.token"
+logging_mmk8s_ca_cert: "{{rsyslog_viaq_config_dir}}/mmk8s.ca.crt"
+# If use_omelasticsearch_cert is True, ca_cert, cert and key in rsyslog_elasticsearch_viaq needs to be set.
 use_omelasticsearch_cert: True
-logging_elasticsearch_ca_cert: "{{rsyslog__viaq_config_dir}}/es-ca.crt"
-logging_elasticsearch_cert: "{{rsyslog__viaq_config_dir}}/es-cert.pem"
-logging_elasticsearch_key: "{{rsyslog__viaq_config_dir}}/es-key.pem"
-
+openshift_logging_use_ops: True
+rsyslog_elasticsearch_viaq:
+  - name: viaq-elasticsearch
+    server_host: logging-es
+    server_port: 9200
+    index_prefix: project.
+    ca_cert: "{{rsyslog_viaq_config_dir}}/es-ca.crt"
+    cert: "{{rsyslog_viaq_config_dir}}/es-cert.pem"
+    key: "{{rsyslog_viaq_config_dir}}/es-key.pem"
+  - name: viaq-elasticsearch-ops
+    server_host: logging-es-ops
+    server_port: 9200
+    index_prefix: .operations.
+    ca_cert: "{{rsyslog_viaq_config_dir}}/es-ca.crt"
+    cert: "{{rsyslog_viaq_config_dir}}/es-cert.pem"
+    key: "{{rsyslog_viaq_config_dir}}/es-key.pem"
 ```
-
-2.2  vars.yaml to configure custom config files.
-
-   To include existing config files in the new ansible deployment, add the paths to rsyslog__custom_config_files as follows.  The specified files are copied to /etc/rsyslog.d.
+The key-value pairs in rsylog_elasticsearch_viaq are used to configure rsyslog to send the logs to the Openshift Aggregated Logging ElasticSearch.  The elements are used in the output elasticsearch configuration 30-elasticsearch.conf as follows (note: not following the abstract syntax):
 ```
-rsyslog__enabled: true
+if index_prefix starts with "project." then {
+  action( 
+      type="omelasticsearch"
+      name="viaq-elasticsearch"
+      server="logging-es"
+      serverport="9200"
+      ....
+      tls.cacert="/etc/rsyslog.d/viaq/es-ca.crt"
+      tls.mycert="/etc/rsyslog.d/viaq/es-cert.pem"
+      tls.myprivkey="/etc/rsyslog.d/viaq/es-key.pem"
+  )
+} else {
+  action( 
+      type="omelasticsearch"
+      name="viaq-elasticsearch-ops"
+      server="logging-es-ops"
+      serverport="9200"
+      ....
+      tls.cacert="/etc/rsyslog.d/viaq/es-ca.crt"
+      tls.mycert="/etc/rsyslog.d/viaq/es-cert.pem"
+      tls.myprivkey="/etc/rsyslog.d/viaq/es-key.pem"
+  )
+}
+```
+The order of the list in rsyslog_elasticsearch_viaq is important.  The first item is in the first if clause with the index_prefix value and the last item is in the else clause.  Elements in between will be placed with else if clause.
+
+
+3. vars.yaml to configure custom config files.
+
+   To include existing config files in the new ansible deployment, add the paths to rsyslog_custom_config_files as follows.  The specified files are copied to /etc/rsyslog.d.
+```
+rsyslog_enabled: true
 ....
-rsyslog__custom_config_files: [ '/path/to/custom_A.conf', '/path/to/custom_B.conf' ]
+rsyslog_custom_config_files: [ '/path/to/custom_A.conf', '/path/to/custom_B.conf' ]
 ```
 
-3. playbook.yaml
+playbook.yaml
+-------------
 
 ```
 - name: install and configure rsyslog on the nodes
@@ -77,7 +145,11 @@ rsyslog__custom_config_files: [ '/path/to/custom_A.conf', '/path/to/custom_B.con
       tags: [ 'role::rsyslog' ]
 ```
 
-   Once ansible-playbook is run with rsyslog__viaq: true, the following configuration files will be deployed.
+Deployed Results
+================
+rsyslog_viaq
+------------
+Once the command-line ansible-playbook is run with rsyslog_viaq: true, the following configuration files will be deployed.
 
 ```
 /etc/rsyslog.conf
@@ -100,18 +172,20 @@ rsyslog__custom_config_files: [ '/path/to/custom_A.conf', '/path/to/custom_B.con
                     mmk8s.token
 ```
 
-   Sample vars.yaml file for the example case
+Sample vars.yaml file for the viaq case
 
 ```
-rsyslog__enabled: true
+rsyslog_enabled: true
 # install example packages & config files
-rsyslog__example: true
-rsyslog__capabilities: [ 'network', 'remote-files', 'tls' ]
-rsyslog__group: root
-rsyslog__user: root
+rsyslog_viaq: true
+rsyslog_capabilities: [ 'viaq', 'viaq-k8s' ]
+rsyslog_group: root
+rsyslog_user: root
 ```
 
-   Once ansible-playbook is run with rsyslog__example: true, the following configuration files will be deployed.
+rsyslog_example
+---------------
+Once ansible-playbook is run with rsyslog_example: true and rsyslog_pki: true, the following configuration files will be deployed.
 
 ```
 /etc/rsyslog.conf
@@ -119,6 +193,7 @@ rsyslog__user: root
                05-common-defaults.conf
                10-local-modules.conf
                10-network-modules.conf
+               20-remote-forward.system
                20-templates.conf
                40-cron.system
                40-dynamic-cron.remote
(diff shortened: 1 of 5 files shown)


sradco:
I believe it might be possible to use the |flatten filter.

- name: Install/Update required packages
  yum:
    name: ['{{ rsyslog__base_packages|flatten }}', '{{ rsyslog__viaq_prereq_packages|flatten if rsyslog__viaq|bool else [] }}', '{{ rsyslog__viaq_packages|flatten if rsyslog__viaq|bool else [] }}', '{{ rsyslog__tls_packages|flatten if rsyslog__pki|bool else [] }}', '{{ rsyslog__packages|flatten }}']
    state: 'latest'
  when: rsyslog__enabled|bool and not use_rsyslog_image|default(False)|bool

@nhosoi

sradco:
We need to add documentation of rsyslog_purge_original_conf and I'm not sure id backing to tempdir is ok. @richm @nhosoi @pcahyna @nkinder

richm:
>  Should " unset $!__REALTIME_TIMESTAMP; "
> also change to "unset $!_REALTIME_TIMESTAMP; " ?

No.  These are journald fields.  Note that some journald fields begin with double underscores `__` and some begin with single underscore `_` and some have no underscore at all.  That cannot be changed.

nhosoi:
@sradco , 
> We need to add documentation of rsyslog_purge_original_conf.

> By default, the Rsyslog configuration files are applied on top of pre-existing configuration files. To purge local files prior to setting new ones, set rsyslog_purge_original_conf variable to 'True', it will move all Rsyslog configuration files to a backup directory before deploying the new configuration files. Defaults to 'False'.

Thanks!  /lgtm  I'm adding it to README.

> I'm not sure if backing to tempdir is ok.

Do you think we'd better specify some more specific name?  How about /tmp/rsyslog-backup-XXXXXX?

nhosoi:
Still debugging, but updating "Install/Update required packages"
```
- name: Install/Update required packages
  yum:
    name: ['{{rsyslog_base_packages|flatten}}','{{rsyslog_viaq_prereq_packages|flatten if rsyslog_viaq|bool else []}}','{{rsyslog_viaq_packages|flatten if rsyslog_viaq|bool else []}}','{{rsyslog_tls_packages|flatten if rsyslog_pki|bool
 else []}}','{{ rsyslog_packages|flatten }}']
    state: 'latest'
  when: rsyslog_enabled|bool and not use_rsyslog_image|default(False)|bool
```
fails the task as follows...
```
fatal: [momo7.localdomain]: FAILED! => {
    "changed": false, 
    "module_stderr": "Shared connection to momo7.localdomain closed.\r\n", 
    "module_stdout": "Traceback (most recent call last):
      File \"/home/nhosoi/.ansible/tmp/ansible-tmp-1541009377.08-165807658319414/AnsiballZ_yum.py\", line 113, in <module> _ansiballz_main()
      File \"/home/nhosoi/.ansible/tmp/ansible-tmp-1541009377.08-165807658319414/AnsiballZ_yum.py\", line 105, in _ansiballz_main
      invoke_module(zipped_mod, temp_path, ANSIBALLZ_PARAMS)
      File \"/home/nhosoi/.ansible/tmp/ansible-tmp-1541009377.08-165807658319414/AnsiballZ_yum.py\", line 48, in invoke_module
      imp.load_module('__main__', mod, module, MOD_DESC)
      File \"/tmp/ansible_yum_payload_8i33FM/__main__.py\", line 1526, in <module>
      File \"/tmp/ansible_yum_payload_8i33FM/__main__.py\", line 1521, in main
      File \"/tmp/ansible_yum_payload_8i33FM/__main__.py\", line 375, in __init__
      File \"/tmp/ansible_yum_payload_8i33FM/ansible_yum_payload.zip/ansible/module_utils/yumdnf.py\", line 79, in __init__
      AttributeError: 'list' object has no attribute 'strip'", 
    "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error", 
    "rc": 1
}
```

roles/rsyslog/tasks/main.yaml
```diff
@@ -1,41 +1,47 @@
...
+  yum:
+    name: '{{rsyslog_tls_packages|flatten}}'
+    state: 'latest'
+  when: rsyslog_enabled|bool and rsyslog_pki|bool and not use_rsyslog_image|default(False)|bool
+
+- name: Install/Update required viaq packages
+  yum:
+    name: '{{rsyslog_viaq_prereq_packages|flatten + rsyslog_viaq_packages|flatten}}'
+    state: 'latest'
+  when: rsyslog_enabled|bool and rsyslog_viaq|bool and not use_rsyslog_image|default(False)|bool
```
nhosoi: @sradco , I could not replace it with one task as you suggested (conditions with "rsyslog_viaq|bool" and "rsyslog_pki|bool" are not allowed in {{ ... }}.   Hope this is acceptable...

sradco:
Hi, 

I managed to get the yum working with this

- name: Install/Update required packages
  yum:
    name: "{{ rsyslog__base_packages|flatten }}+{{ rsyslog__viaq_prereq_packages|flatten if rsyslog__viaq else []}}+{{ rsyslog__viaq_packages|flatten if rsyslog__viaq else [] }}+{{ rsyslog__tls_packages|flatten if rsyslog__pki else [] }}+{{ rsyslog__packages|flatten }}"
    state: 'latest'
  when: rsyslog__enabled|bool and not use_rsyslog_image|default(False)|bool


@nhosoi Can you please review?

nhosoi:
Thanks, @sradco !
I thought I also tried the case, but obviously I didn't... :p  Yes, it worked for me, too.  Let me update this pr...

roles/rsyslog/defaults/main.yaml
```diff
@@ -977,31 +981,36 @@ rsyslog_conf_viaq_elasticsearch:
...
-              {% if use_omelasticsearch_cert | default(False) %}
-                  tls.cacert="{{logging_elasticsearch_ca_cert}}"
-                  tls.mycert="{{logging_elasticsearch_cert}}"
-                  tls.myprivkey="{{logging_elasticsearch_key}}"
-              {% endif %}
+          {% else %}
+          {% if res.name|default("") == "primary" %}
+              if $.viaq_index_prefix startswith "{{ res.index_prefix }}" then {
+          {% else %}
+              elif $.viaq_index_prefix startswith "{{ res.index_prefix }}" then {
```
richm: I don't think rsyslog rainierscript has an `elif` keyword

nhosoi:
Hi @sradco,

I committed 2 patches.  Could you please take a look?
1) Updated the README and now it has the Table of Contents.
https://github.com/nhosoi/logging/tree/working/roles/rsyslog#linux-system-roles-rsyslog
2) Extended the omelasticsearch to allow more than 2 servers.  (As I noted in the commit comments, since "else" is missing, I'm a bit uncomfortable with this patch...)
https://github.com/linux-system-roles/logging/pull/20/commits/1fb07e884036dfaddfb99869e847fe21fcfd3652

roles/rsyslog/defaults/main.yaml
nhosoi: Oops.  Thanks, @richm .  I mixed up with the ansible syntax... :p
Then, all we could do is like this?
```
if $.viaq_index_prefix startswith "projects." then {
}
if $.viaq_index_prefix startswith ".operations." then {
}
...
```

nhosoi:
@sradco,
> I meant , you removed
> "Create required system group." task
Ahhhh, you are right.  I've resurrected the task.  Thanks!!

nhosoi:
Thanks, @richm !  jinja2 provides `loop` to tell us the index as well as the length of the for loop, which allows us to know when to use `if` or `else`. :)

roles/rsyslog/defaults/main.yaml
```diff
@@ -959,7 +959,7 @@ rsyslog_conf_viaq_elasticsearch:
                   set $.viaq_index_name = exec_template("prefix_index_template");
               }
           {% for res in rsyslog_elasticsearch %}
-          {% if res.index_prefix|default("") == "" %}
+          {% if loop.length == 1 %}
```
sradco: Can you please explain what is this loop.length ==1 used for?  I'm probably missing something in the logic ... @nhosoi

roles/rsyslog/defaults/main.yaml
```diff
@@ -948,36 +961,51 @@ rsyslog__conf_viaq_elasticsearch:
...
+              if $.viaq_index_prefix startswith "{{ res.index_prefix }}" then {
+                set $.es_host="{{ res.server_host | default('logging-es') }}"
+                set $.es_port="{{ res.server_port | default(9200) }}"
+          {% if use_omelasticsearch_cert | default(False) %}
+                set $.es_ca_cert="{{ res.ca_cert }}"
+                set $.es_cert="{{ res.cert }}"
+                set $.es_key="{{ res.key }}"
           {% endif %}
+          {% elif loop.last %}
+              else {
```
sradco: Do we assume here that the last elasticsearch in the `rsyslog_elasticsearch` list is the one all "miscellaneous" logs go to? Why not the first 'Default` es?
If we do assume that, we should mention it in the documentation very clearly...

What will be the index prefix of the logs that go to es-ops ?

We should add the index_prefix parameter to the README.

@nhosoi @richm

roles/rsyslog/README.md
```diff
@@ -21,53 +46,56 @@ localhost ansible_user=YOUR_ANSIBLE_USER openshift_logging_use_ops=True
...
 # install viaq packages & config files
-rsyslog__viaq: true
-rsyslog__capabilities: [ 'viaq' ]
-rsyslog__group: root
-rsyslog__user: root
+rsyslog_viaq: true
+rsyslog_capabilities: [ 'viaq' ]
+rsyslog_group: root
+rsyslog_user: root
 elasticsearch_server_host: es_hostname
```
sradco: Need to update here to the rsyslog_elasticsearch parameter

rsyslog_elasticsearch:
  - name: default
    server_host: es_hostname
    server_port: 9200
    ca_cert: "{{rsyslog_viaq_config_dir}}/es-ca.crt"
    cert: "{{rsyslog_viaq_config_dir}}/es-cert.pem"
    key: "{{rsyslog_viaq_config_dir}}/es-key.pem"

roles/rsyslog/defaults/main.yaml
```diff
@@ -948,36 +961,51 @@ rsyslog__conf_viaq_elasticsearch:
...
+          {% if loop.length == 1 %}
+              set $.es_host="{{ res.server_host | default('logging-es') }}"
+              set $.es_port="{{ res.server_port | default(9200) }}"
+            {% if use_omelasticsearch_cert | default(False) %}
+              set $.es_ca_cert="{{ res.ca_cert }}"
+              set $.es_cert="{{ res.cert }}"
+              set $.es_key="{{ res.key }}"
+            {% endif %}
+          {% else %}
+          {% if loop.first %}
```
sradco: Can we remove the above  {% if loop.length == 1 %}  section from lines 965 till 973 and instead update the loop conditions in this line (974) and line 983 to:

{% if loop.first and  loop.length > 1  %}
{% elif loop.last and  loop.length > 1 %}

?

@nhosoi @richm

roles/rsyslog/defaults/main.yaml
```diff
@@ -989,32 +1017,30 @@ rsyslog__conf_viaq_elasticsearch:
...
                   retryruleset="try_es"
                   usehttps="on"
-              {% if use_omelasticsearch_cert | default(False) %}
-                  tls.cacert="{{logging_elasticsearch_ca_cert}}"
-                  tls.mycert="{{logging_elasticsearch_cert}}"
-                  tls.myprivkey="{{logging_elasticsearch_key}}"
-              {% endif %}
-              )
-              }
+          {% if use_omelasticsearch_cert | default(False) %}
```
sradco: Shouldn't this line and the 3 lines below it be removed since they were moved inside the loop? @nhosoi

roles/rsyslog/README.md
```diff
@@ -21,53 +46,56 @@ localhost ansible_user=YOUR_ANSIBLE_USER openshift_logging_use_ops=True
...
-rsyslog__group: root
-rsyslog__user: root
-# If 'viaq-k8s' is in rsyslog__capabilities, logging_mmk8s_* need to be specified.
-logging_mmk8s_token: "{{rsyslog__viaq_config_dir}}/mmk8s.token"
-logging_mmk8s_ca_cert: "{{rsyslog__viaq_config_dir}}/mmk8s.ca.crt"
+rsyslog_viaq: true
+rsyslog_capabilities: [ 'viaq', 'viaq-k8s' ]
+rsyslog_group: root
+rsyslog_user: root
+# If 'viaq-k8s' is in rsyslog_capabilities, logging_mmk8s_* need to be specified.
```
sradco: Need to update here to the rsyslog_elasticsearch parameter
```
rsyslog_elasticsearch:
  - name: default
    server_host: es_hostname
    server_port: 9200
    # If use_omelasticsearch_cert is True, logging_elasticsearch_* need to be specified.
    logging_elasticsearch_ca_cert: "{{rsyslog_viaq_config_dir}}/es-ca.crt"
    logging_elasticsearch_cert: "{{rsyslog_viaq_config_dir}}/es-cert.pem"
    logging_elasticsearch_key: "{{rsyslog_viaq_config_dir}}/es-key.pem"
```

roles/rsyslog/README.md
```diff
@@ -21,53 +46,56 @@ localhost ansible_user=YOUR_ANSIBLE_USER openshift_logging_use_ops=True
...
+logging_mmk8s_token: "{{rsyslog_viaq_config_dir}}/mmk8s.token"
+logging_mmk8s_ca_cert: "{{rsyslog_viaq_config_dir}}/mmk8s.ca.crt"
 # If use_omelasticsearch_cert is True, logging_elasticsearch_* need to be specified.
```
sradco: Need to update here to the rsyslog_elasticsearch parameter
```
rsyslog_elasticsearch:
  - name: default
    server_host: es_hostname
    server_port: 9200
    # If use_omelasticsearch_cert is True, logging_elasticsearch_* need to be specified.
    ca_cert: "{{rsyslog_viaq_config_dir}}/es-ca.crt"
    cert: "{{rsyslog_viaq_config_dir}}/es-cert.pem"
    key: "{{rsyslog_viaq_config_dir}}/es-key.pem"
```

roles/rsyslog/defaults/main.yaml
```diff
@@ -948,36 +961,51 @@ rsyslog__conf_viaq_elasticsearch:
               } else {
                   set $.viaq_index_name = exec_template("prefix_index_template");
               }
-          {% if openshift_logging_use_ops | default(False) %}
-              if $.viaq_index_prefix startswith "project." then {
+          {% for res in rsyslog_elasticsearch %}
```
sradco: Should this be rsyslog_elasticsearch or perhaps better to use rsyslog_viaq ? or rsyslog_elasticsearch_viaq?

@nhosoi @richm

roles/rsyslog/defaults/main.yaml
richm: > Do we assume here that the last elasticsearch in the rsyslog_elasticsearch list is the one all "miscellaneous" logs go to?

Yes

> Why not the first 'Default` es?

We are getting rid of the `default` keyword.

> If we do assume that, we should mention it in the documentation very clearly...

We should document that the elasticsearch outputs will be processed in the order specified, and the last one is the default or "catch-all" output with no condition.

roles/rsyslog/defaults/main.yaml
richm: I think we should make it so that the values in `rsyslog_elasticsearch` can be used for any/all elasticsearch outputs - `rsyslog_elasticsearch` and `rsyslog_viaq` and `rsyslog_elasticsearch_viaq`.

roles/rsyslog/defaults/main.yaml
nhosoi: > We should document that the elasticsearch outputs will be processed in the order specified, and the last one is the default or "catch-all" output with no condition.

Will do.  Thanks.

roles/rsyslog/defaults/main.yaml
nhosoi: ???  Am I supposed to replace `rsyslog_elasticsearch` with `rsyslog_elasticsearch_viaq` since the current config has viaq specific setting, e.g., template="viaq_template"?  So far, we do not have a "generic" elasticsearch output for the non-viaq case as in https://github.com/linux-system-roles/logging/issues/8.  Once it is implemented, we should use `rsyslog_elasticsearch` variable?  Is that the suggestion?

richm: Something like that, yes.  I think we can make a more generic rsyslog_elasticsearch

roles/rsyslog/defaults/main.yaml
```diff
@@ -948,36 +961,51 @@ rsyslog__conf_viaq_elasticsearch:
...
+          {% if loop.first %}
+              if $.viaq_index_prefix startswith "{{ res.index_prefix }}" then {
+                set $.es_host="{{ res.server_host | default('logging-es') }}"
+                set $.es_port="{{ res.server_port | default(9200) }}"
+          {% if use_omelasticsearch_cert | default(False) %}
+                set $.es_ca_cert="{{ res.ca_cert }}"
+                set $.es_cert="{{ res.cert }}"
+                set $.es_key="{{ res.key }}"
           {% endif %}
+          {% elif loop.last %}
```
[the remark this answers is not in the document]
sradco: What will happen in case user did not specify an index_prefix and it is not the last elasticsearch?
Do we want to fail in this case?

@richm @nhosoi @pcahyna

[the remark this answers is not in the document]
richm: @sradco yes

roles/rsyslog/README.md
richm: Using the loop index allows us to get rid of the `name` field for the `default` value, to know how to set up the `if`/`else if`/`else` conditions.

But - rsyslog actions have a `name` field https://www.rsyslog.com/doc/v8-stable/configuration/actions.html#general-action-parameters - we will need to use this for stats and rsyslog monitoring e.g. if you have several elasticsearch outputs, you will need to name them, so you can look at a prometheus/grafana dashboard and tell that elasticsearch-ops is having problems but not elasticsearch-apps.

sradco: Where is it used in the code to reflect that? @richm

richm: @sradco something like this:
```
action(
  type="omelasticsearch"
  name="elasticsearch-apps"
...
```

roles/rsyslog/defaults/main.yaml
nhosoi: I'm afraid action is not that dynamic to take variables as a param value... (;_;)
E.g., https://github.com/rsyslog/rsyslog/issues/1149
Note: template seems ok, but not action.   Obviously, the variable param takes a variable, though.
` action(type="mmnormalize" ruleBase="/etc/rsyslog.d/viaq/parse_json.rulebase" variable="$!log")`
I'm putting action in each if/else if clause...

richm: @nhosoi not sure what you mean - not sure what you were trying to do

nhosoi: This 30-elasticsearch.conf fails with an error [1].  If the serverport param is set to `serverport="9200"`, the error disappears.  First, I thought it could be a type mismatch issue, e.g., string vs integer.  But now I'm suspecting the variable may not be evaluated in the action parameter...  I could be wrong but the issue rsyslog/rsyslog#1149 mentions dynamically updating action may cause thread unsafe situation, thus not supported?>>
```
    set $.es_host = "logging-es";
    set $.es_port = "9200";
    action(
        type="omelasticsearch"
        server="$.es_server"
        serverport="$.es_port"
        template="viaq_template"
        searchIndex="index_template"
        ...
        usehttps="on"
    )
```
[1]
```
Nov 05 18:37:50 ip-172-18-13-50.ec2.internal rsyslogd[24987]: error during parsing file /etc/rsyslog.d/viaq/30-elasticsearch.conf, on or before line 76: parameter 'serverport' is not a proper number [v8.37.0-6.2.el7 try http://www.rsyslog.com/e/2207 ]
Nov 05 18:37:50 ip-172-18-13-50.ec2.internal rsyslogd[24987]: error during parsing file /etc/rsyslog.d/viaq/30-elasticsearch.conf, on or before line 76: errors occured in file '/etc/rsyslog.d/viaq/30-elasticsearch.conf' around line 76 [v8.37.0-6.2.el7 try http://www.rsyslog.com/e/2207 ]
```
What I'm trying to do is very simple - just going back to the previous code, where we don't use the rsyslog local varialbe but put them in action param directly.
```
if $.viaq_index_prefix startswith "project." then {
  action( server="logging-es" serverport="9200" ...)
} else if $.viaq_index_prefix startswith ".operations." then {
  action( server="logging-es-ops" serverport="9200" ...)
} else {
  action( server="logging-es-others" serverport="9200" ...)
}
```

richm: Yes - that's the only way you can handle it.

roles/rsyslog/defaults/main.yaml
nhosoi: I think this is ok.  My test passed. :)  The diff might be unclear; the real code looks like this.
```
          {% else %}
              else if $.viaq_index_prefix startswith "{{ res.index_prefix }}" then {
                action(
                    type="omelasticsearch"
                    name="{{ res.name | default('viaq-elasticsearch') }}"
                    server="{{ res.server_host | default('logging-es') }}"
                    serverport="{{ res.server_port | default(9200) | int }}"
                    template="viaq_template"
                    searchIndex="index_template"
                    dynSearchIndex="on"
                    searchType="com.redhat.viaq.common"
                    bulkmode="on"
                    writeoperation="create"
                    bulkid="id_template"
                    dynbulkid="on"
                    retryfailures="on"
                    retryruleset="try_es"
                    usehttps="on"
            {% if use_omelasticsearch_cert | default(False) %}
                    tls.cacert="{{ res.ca_cert }}"
                    tls.mycert="{{ res.cert }}"
                    tls.myprivkey="{{ res.key }}"
            {% endif %}
                )
          {% endif %}
```

roles/rsyslog/defaults/main.yaml
nhosoi: Ok.  Thanks!

roles/rsyslog/README.md
```diff
@@ -100,25 +172,28 @@ rsyslog__custom_config_files: [ '/path/to/custom_A.conf', '/path/to/custom_B.con
...
 ```
-rsyslog__enabled: true
+rsyslog_enabled: true
 # install example packages & config files
-rsyslog__example: true
-rsyslog__capabilities: [ 'network', 'remote-files', 'tls' ]
-rsyslog__group: root
-rsyslog__user: root
+rsyslog_viaq: true
+rsyslog_capabilities: [ 'viaq', 'viaq-k8s' ]
```
sradco: I think you have an error here @nhosoi

roles/rsyslog/README.md
```diff
@@ -21,53 +46,96 @@ localhost ansible_user=YOUR_ANSIBLE_USER openshift_logging_use_ops=True
...
-rsyslog__capabilities: [ 'viaq', 'viaq-k8s' ]
-rsyslog__group: root
-rsyslog__user: root
-# If 'viaq-k8s' is in rsyslog__capabilities, logging_mmk8s_* need to be specified.
-logging_mmk8s_token: "{{rsyslog__viaq_config_dir}}/mmk8s.token"
-logging_mmk8s_ca_cert: "{{rsyslog__viaq_config_dir}}/mmk8s.ca.crt"
-# If use_omelasticsearch_cert is True, logging_elasticsearch_* need to be specified.
+rsyslog_viaq: true
+rsyslog_capabilities: [ 'viaq', 'viaq-k8s' ]
+rsyslog_group: root
```
sradco: Why not set rsyslog_unprivileged=false instead of setting the group and user?
@nhosoi @richm

roles/rsyslog/README.md
sradco: I think this should be rsyslog_capabilities: [ 'example' ]

@nhosoi

sradco:
/lgtm
Fixes: #12 , #13 , #15.

PR #27, opened 2018-11-11
[WIP] Update the logging role structure

sradco:
Removed ovirt specific fluentd role.

Updated logging README.

Added tasks to set facts to pass to the rsyslog role.

Please review: @richm @pcahyna @nhosoi @tabowling


sradco:
Fixes issue #37

roles/rsyslog/roles/output_roles/viaq/tasks/main.yaml
```diff
@@ -0,0 +1,52 @@
+---
+- name: Collect local Elasticsearch cert files
+  set_fact:
+    elasticsearch_local_cert_files: "{{ elasticsearch_local_cert_files |d([]) }} + {{ [item.ca_cert_src|d([])] }} + {{ [item.cert_src|d([])] }} + {{ [item.key_src|d([])] }}"
+  with_items: "{{ rsyslog_elasticsearch }}"
+  when: item.use_local_omelasticsearch_cert|default(Flase)|bool
```
richm: `default(False)`

roles/rsyslog/README.md
```diff
@@ -269,17 +280,22 @@ Viaq sub-variables
 ------------------
 - `rsyslog_viaq_config_dir`: Directory to store viaq configuration files.  Default to '/etc/rsyslog.d/viaq'.
 - `rsyslog_viaq_log_dir`: Viaq log directory.  Default to '/var/log/containers'.
-- `openshift_logging_use_ops`: Set to 'True', if you have a second ES cluster for infrastructure logs. Default to 'False'.
 - `logging_mmk8s_token`: Path to token for kubernetes.  Default to "/etc/rsyslog.d/viaq/mmk8s.token"
 - `logging_mmk8s_ca_cert`: Path to CA cert for kubernetes.  Default to "/etc/rsyslog.d/viaq/mmk8s.ca.crt"
-- `rsyslog_elasticsearch_viaq`: A set of following variables to specify output elasticsearch configurations.  It could be an array if multiple elasticsearch clusters to be configured. 
+- `rsyslog_elasticsearch`: A set of following variables to specify output elasticsearch configurations. It could be an array if multiple elasticsearch clusters to be configured.
+- `use_omelasticsearch_cert` : If set to `True`, omelasticsearch is configured to use the certificates specified in rsyslog_elasticsearch.  Default to `False`.
```
richm: The CA cert is required if using `usehttps`: https://www.rsyslog.com/doc/v8-stable/configuration/modules/omelasticsearch.html#usehttps - so I suppose we need to provide an ansible parameter for `usehttps`.

In addition - if using client cert auth, you must have `usehttps="on"`, and you must specify a client cert and key

sradco: It is part of the action parameters. For viaq the default id "on".
I was thinking how to generelize this for any omelasticsearch output where user can just pass the plugin parameters to the "action". I still dont have a solution that works.
@richm @nhosoi @pcahyna

nhosoi: Indeed, this is a bug.  Currently, 'usehttps="on"' is hardcoded in the defaults/main.yaml.  (there are 4 of them.)  Can we move usehttps="on" in the if use_omelasticsearch_cert clause (exchanging line 992 and 993)?  And the client auth configuration needs to be added...
```
 977               action(
 978                   type="omelasticsearch"
 979                   name="{{ res.name | default('viaq-elasticsearch') }}"
                          ..........................................
 992                   usehttps="on"
 993           {% if use_omelasticsearch_cert | default(False) %}
 994                   tls.cacert="{{ res.ca_cert }}"
 995                   tls.mycert="{{ res.cert }}"
 996                   tls.myprivkey="{{ res.key }}"
 997           {% endif %}
 998               )
```

richm: > Can we move usehttps="on" in the if use_omelasticsearch_cert clause (exchanging line 992 and 993)?

For a general purpose elasticsearch configuration api, we need a separate option for `usehttps="on"` - maybe `use_omelasticsearch_https` - then add to defaults.yaml - `use_omelasticsearch_https: {{ true if use_omelasticsearch_cert else false }}` - not sure about the ansible syntax here , but the gist is that you want `use_omelasticsearch_https` to default to `true` if the user has set `use_omelasticsearch_cert: true`, otherwise, `false`.  Then something like this:
```
   usehttps="{{ 'on' if use_omelasticsearch_https else 'off' }}"
```
again - not sure of the exact ansible syntax here

sradco: so, if I understand correctly, if "use_omelasticsearch_cert" is "true" then "usehttps" is "on".
Is there an option to use "usehttps" is "on" without the tls certs?
@richm

nhosoi: Probably, we should distinguish encrypting the communication (usehttps on) from setting up certs?  We cannot enable usehttps without certs, but we can disable it even if the certs are configured.

richm: You can use omelasticsearch without certs and without https.

You can set `usehttps="on"` if
- you set `allowunsignedcerts="on"` (default `"off"` - insecure to use `"on"` - not recommended)
OR
- you specify a ca cert with `tls.cacert`

You can use client cert auth if
- you set `usehttps="on"`
AND
- you provide a CA cert with `tls.cacert`
AND
- you provide both `tls.mycert` and `tls.myprivkey`

> Probably, we should distinguish encrypting the communication (usehttps on) from setting up certs?

Yes, but you cannot use the client cert/key if `usehttps="off"`

> We cannot enable usehttps without certs,

You can - see above - but having a ca cert is strongly encouraged and secure.

> but we can disable it even if the certs are configured.

I suppose there are cases where the user may want to provide certs, but not use `https`?  But then the certs are useless.

nhosoi: Is allowunsignedcerts a parameter in the omelasticsearch?  I cannot find it in this doc...
https://www.rsyslog.com/doc/v8-stable/configuration/modules/omelasticsearch.html

The mmkubernetes module has it, though...
https://www.rsyslog.com/doc/v8-stable/configuration/module/mmkubernetes.html#allowunsignedcerts

sradco: Lets open a separate issue on this and track it there. ok?
@nhosoi @richm

richm: > Is allowunsignedcerts a parameter in the omelasticsearch? I cannot find it in this doc...
> https://www.rsyslog.com/doc/v8-stable/configuration/modules/omelasticsearch.html

It is not documented: https://github.com/rsyslog/rsyslog/blob/master/plugins/omelasticsearch/omelasticsearch.c#L203

> 
> The mmkubernetes module has it, though...
> https://www.rsyslog.com/doc/v8-stable/configuration/module/mmkubernetes.html#allowunsignedcerts

nhosoi: Thanks, @richm.  Indeed, it is in the source code.

sradco:
@nhosoi @richm @pcahyna @tabowling 

This PR is ready for review. Please review it.

Noriko, please test to see that the configuration files for viaq use case a set correctly.
Please note, I have separated the output to elasticsearch from the viaq input modules.
The configuration files will be placed under elasticsearch sub directory.

nhosoi:
@sradco, I tested this pr/27 in the logging test env using this private branch.
https://github.com/nhosoi/origin-aggregated-logging/tree/release-3.11-system-roles/hack/testing/rsyslog/roles/rsyslog
and got this successful result!!
```
[INFO] Logging test suite test-zzz-rsyslog started at Tue Dec 25 00:30:55 UTC 2018
[INFO] Logging test suite test-zzz-rsyslog succeeded at Tue Dec 25 00:33:22 UTC 2018
```

nhosoi:
@sradco, sorry for my slow response...
> Can you please explain how you test? What vars.yaml file did you use for the testing ?

Using this private branch, I replaced roles/rsyslog with the rsyslog roles in this PR, then ran the rsyslog test suite. 

https://github.com/nhosoi/origin-aggregated-logging/tree/release-3.11-system-roles/hack/testing/rsyslog/roles/rsyslog

> Did you try using the local elasticsearch certs?

No, local elasticsearch certs are not used in this test.  Note: this is the vars.yaml used by the test.

https://github.com/nhosoi/origin-aggregated-logging/blob/release-3.11-system-roles/test/zzz-rsyslog.sh#L92-L123

The elasticsearch certs are installed by this openshift_logging_rsyslog task.

https://github.com/nhosoi/origin-aggregated-logging/blob/release-3.11-system-roles/hack/testing/rsyslog/roles/openshift_logging_rsyslog/tasks/setup_certs.yaml

PR #30, opened 2018-11-12
rsyslog_enable controls the config files as well as the service.

nhosoi:
If true, it deploys the specified configuration and starts the service.
If false, it cleans up the configuration and stops the service.

Fixing "Remove task".


sradco:
Updated your changes to patch #27 . This commit can be abandoned. @nhosoi

PR #32, opened 2018-11-16
[WIP] Support elasticsearch certs deployment

nhosoi:
rsyslog: Supporting local paths of ca_cert, cert and key for the elasticsearch to be deployed to the remote host.
Adding use_local_omelasticsearch_cert.
Adding the suggestion from @sradco to "Collect local viaq cert files" task.

### DIFF
--- a/roles/rsyslog/README.md
+++ b/roles/rsyslog/README.md
@@ -61,8 +61,10 @@ rsyslog_enabled: true
 rsyslog_viaq: true
 rsyslog_capabilities: [ 'viaq' ]
 rsyslog_unprivileged: False
-elasticsearch_server_host: es_hostname
-elasticsearch_server_port: 9200
+rsyslog_elasticsearch_viaq:
+  - name: viaq-elasticsearch
+    server_host: es-hostname
+    server_port: 9200
 ```
 
 2. vars.yaml to configure to handle the inputs from openshift containers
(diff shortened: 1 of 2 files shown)


sradco:
Should it be either take files from local, which means user states ca_src, cert_src and key_src OR take them from the /etc/rsyslog.d/viaq/ directory and then user states ca, cert and key?

I see that in this patch we take both and documentation is not clear..

@richm @nhosoi

richm:
> I think this should be per elasticsearch. So potentially, you can send to 1 elasticsearch with cert auth and to another without. Does that make sense?

yes
or to have a different cert/key for each elasticsearch

richm:
> I updated your patch in my PR
> https://github.com/linux-system-roles/logging/pull/27/files#diff-659e5a90bded4fec488b5097947f94e5
> I believe it works now as expected. Please review.
> 
> But I have a question, Shouldn't we take the local files from ca_src, cert_src and key_src OR take them from form /etc/rsyslog.d/viaq/ directory + ca, cert and key?
> 
> I see that in this patch we take both and documentation is not clear..

We should take both but there should be a clear precedence - take take the local files from ca_src, cert_src and key_src if specified, or take them from form /etc/rsyslog.d/viaq/ directory + ca, cert and key

sradco:
@nhosoi , I believe this can be abandoned since PR #27 added support for it.
Please test current master handles this currectly.

PR #50, opened 2019-01-06
 Update ovirt role timestamp parsing and sub roles deploy config to /etc/rsyslog.d directory

sradco:
Updated required oVirt timestamp parsing.

Updated sub-roles default config directory to deploy all configuration files to /etc/rsyslog.d/ directory.
When having the pre sub-role directory, there was a problem with the ordering of the files containing the "$IncludeConfig" statement and caused the files to be run in a different order the expected.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -10,7 +10,3 @@ logging_custom_config_files: []
 
 logging_purge_confs: true
 
-logging_elasticsearch_default_ca_cert_path: /path/to/ca-cert.pem
-logging_elasticsearch_default_client_cert_path: /path/to/cert.pem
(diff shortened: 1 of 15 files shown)


sradco:
I kept the optional dedicated sub-directory, but set it to default to /etc/rsyslog.d.
Do you thing we should remove this option completely? @richm @nhosoi @pcahyna

nhosoi:
> Do you think we should remove this option completely?

+1
It makes the installation/configuration simpler.
Plus, we can just close issue/44 and pr/48 once the option is removed. :)

sradco:
Updated all config files deployment paths to /etc/rsyslog.d and verified.

PR #52, opened 2019-01-08
Fixing 2 issues in viaq-k8s/defaults/main.yaml

nhosoi:
1) Separating input and the rest from the module type mmk8s.conf.
   mmk8s.conf is set to the module type, which filename is 10-mmk8s.conf.
   On the other hand, the "imfile" module is loaded in viaq_main.  The
   filename is 10-viaq_main.conf which is evaluated after 10-mmk8s.conf.
   To ease the confusion, separating the imfile input and the rest and
   set it to the "input" type.
2) Replacing rsyslog_log_dir with rsyslog_viaq_log_dir since it fails with
   unknown variable.

### DIFF
--- a/roles/rsyslog/roles/input_roles/viaq-k8s/defaults/main.yaml
+++ b/roles/rsyslog/roles/input_roles/viaq-k8s/defaults/main.yaml
@@ -24,7 +24,8 @@ rsyslog_viaq_k8s_packages: [ 'rsyslog-mmjsonparse', 'rsyslog-mmkubernetes' ]
 
 rsyslog_viaq_k8s_rules:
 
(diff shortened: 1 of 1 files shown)


sradco:
I'm sorry for introducing this regression. Please let me know how it goes.

nhosoi:
@sradco, No, not a problem.  It's my fault not to test properly.  Now, I'm more confident. :)

The CI test issue was the type of the separated mmk8s config...  The order to include rsyslog config files is quite delicate. :)
```
   To ease the confusion, separating the imfile input and the rest and
   set it to the "template" type.  (Note: Since this separated config
   contains action which should be called before output. Thus 'template'
   has to be selected.)
```
Now I'm removing WIP.

PR #53, opened 2019-01-24
Porting openshift/cluster-logging-operator rsyslog configuration into…

nhosoi:
… linux-system-roles.

Along with this effort, cluster-logging-operator is being updated to
generate rsyslog configuration files from linux-system-roles/logging.
The generated files are placed in the working directory once and will
be deployed in the containers.  To support the ability to place files
in a directory which is different from the final location, additional
config_dir rsyslog_file_config_dir has been introduced.
1) rsyslog_config_dir for the path in the config files, and
2) rsyslog_file_config_dir for the path to place the config files.

New variables:
- `rsyslog_parent_config_dir`: Directory to store rsyslog.conf.
  Default to '/etc'.
- `rsyslog_files_config_dir`: Directory to locate rsyslog.conf.
  Used in deploying the config files.  Default to '/etc/rsyslog.d'.
- `rsyslog_mode`:  Mode for the config files.  Default to '0400'
- `rsyslog_max_message_size`: Maximum supported message size.
  Default to 8k.
- `use_rsyslog_image`: If set to 'True', the rsyslog is run in the
  OpenShift container.

Rsyslog allows to use environment variables as string constants as follows.
  key=`echo $ENVVAR`
It is used in the script to start the rsyslog container.  They are
activated only when use_rsyslog_image is true.
  {% if use_rsyslog_image|bool %}
    key=`echo $ENVVAR`
  {% endif %}

Another requirement is the include order. It turned out it is crucial
to keep the include order of the configuration files.  To support it,
'format' type is introduced, which sets the prefix 60- to the filename.
To maintain the order more precisely, some files have the secondary
prefix, e.g., 00-elasticsearch.  With the type 'format', it generates
60-00-elasticsearch.conf which is included prior to 60-10-mmk8s.conf.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -10,7 +10,8 @@ logging_custom_config_files: []
 
 logging_purge_confs: false
 
-# Configuration directory
-# Unless this is set, "Set rsyslog_elasticsearch" fails.
-rsyslog_config_dir: '/etc/rsyslog.d'
-
+# .. envvar:: can_mod_system
+#
+# Indicates if the ansible executer can do chmod, add user & group, etc.
+# It's checked by {{ lookup('env', 'USER') }} and set to True if the executer is root.
+can_mod_system: False
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -9,3 +9,8 @@ logging_outputs: []
 logging_custom_config_files: []
 
 logging_purge_confs: false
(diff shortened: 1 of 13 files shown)


nhosoi:
Since it is obsolete and no requirements for the enhancements, closing this pr.

PR #54, opened 2019-01-28
Add in-tree tests

nkinder:
This adds in-tree tests that utilize molecule to test the logging
role.  We use the molecule's docker driver, which will handle the
provisioning and cleanup of containers to use as the target hosts
when testing the roles.  The tests will use a CentOS 7 container
by default, but other targets can easily be specified when the
tests are launched.

The test playbook that we use for now just tests the logging role
defaults.  We can add additional test scenarios in future patches.

Note that there are a number of lint issues with the current role.
For this initial commit, I have disabled the lint tests that fail
so the tests will progress to actually executing the test playbook.
These lint issues should be cleaned up in future commits, and we
should re-enable those lint tests to catch future issues.


sradco:
LGTM

PR #60, opened 2019-02-20
Add meta information for Galaxy and fix bug in Elasticsearch output

### LINKED ISSUE #58: Include meta information
As of now, pulling this role using `ansible-galaxy` fails with:

```
 [WARNING]: - linux-system-roles-logging was NOT installed successfully: this
role does not appear to have a meta/main.yml file.
```

Please include a `meta/main.yml` file to prevent this.

sradco:
Add meta information for Galaxy
Fix bug in Elasticsearch output when using allowUnsignedCerts variable

Fixes #58 

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -1,0 +1,15 @@
+---
+galaxy_info:
+  author: Shirly Radco <sradco@redhat.com>
+  description: Configure logging collection
+  galaxy_tags: ['system', 'logging', 'redhat', 'rhel', 'fedora', 'centos']
+  company: Red Hat, Inc.
+  license: GPL-3.0+
+  min_ansible_version: 2.7
+  platforms:
+  - name: Fedora
+    versions: [27, 28]
+  - name: EL
+    versions: [7]
+
+allow_duplicates: true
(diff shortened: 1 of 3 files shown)


dawud:
Hello @sradco, it's probably my fault for not being more specific, but, as it stands, this repository cannot be considered a role by itself, rather an entrypoint playbook that uses a role (`rsyslog` in this case), which is the one lacking meta information (should the `rsyslog` role have a separate linux-system-role repository and `logging` declare a dependency on it?)
It is, however, the case that the `rsyslog` role also has nested roles in it. I assume it's a design decision, but it's a bit confusing.
It is my understanding that roles should be independent entities (and consumable using `ansible-galaxy`) and pull dependencies as needed (using the meta file).
Apologies if I wasn't clear in my initial ask (#58).

sradco:
@dawud 
I'm not sure what you mean.
Rsyslog sub role is **not** meant to be run independently and also its sub roles.

This role is intended to configure logging on RHEL and other projects machines.
It uses Rsyslog as the default collector, but can also be enriched with additional log collector.
The role's variables are general and not Rsyslog specific.
The Rsyslog sub roles define the packages required and the configuration files for each desperately for each project.

@pcahyna Do you agree?

PR #65, opened 2019-06-12
Add ovirt.entity to ovirt collectd records

sradco:
When moving to Rsyslog we removed the ovirt.entity
field that caused the VMS dashboard and VMs related
visualizations to break.
Adding the field back for backwards compatibility.

Bug-Url: https://bugzilla.redhat.com/1717974

### DIFF
--- a/roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
+++ b/roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
@@ -90,6 +90,11 @@ rsyslog_conf_ovirt_formatting:
               action(type="mmjsonparse" cookie="") # parse entire message as json
               set $!@timestamp = exec_template('cnvt_to_viaq_timestamp');
               unset $!time;
+              if $!collectd!plugin == 'virt' then {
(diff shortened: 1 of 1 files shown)


roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
```diff
@@ -90,6 +90,11 @@ rsyslog_conf_ovirt_formatting:
               action(type="mmjsonparse" cookie="") # parse entire message as json
               set $!@timestamp = exec_template('cnvt_to_viaq_timestamp');
               unset $!time;
+              if $!collectd!plugin == 'virt' then {
```
richm: @nhosoi is this safe to do?  Or should there first be a check for `strlen($!collectd!plugin) > 0`

sradco: @nhosoi @richm  In Collectd the  $!collectd!plugin always exists.

richm: ok

PR #70, opened 2019-06-30
Update oVirt config with default logs handling

sradco:
Added to oVirt formatting config the handling of
general logs, other the Collectd and oVirt vdsm and engine
logs.

This patch also fixes an issue in the default rsyslog.conf.

Please review my patch: @richm @pcahyna @nhosoi @sandrobonazzola @didib

### DIFF
--- a/roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
+++ b/roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
@@ -55,6 +55,14 @@ rsyslog_conf_ovirt_main_modules:
     sections:
 
       - options: |-
+          # Read from systemd journal
+          module(load="imjournal" StateFile="imjournal.state")
+
+      - options: |-
+          # Use default timestamp format
+          module(load="builtin:omfile" Template="RSYSLOG_TraditionalFileFormat")
+
+      - options: |-
           # Read from file with inotify enabled
           module(load="imfile" mode="inotify")
 
@@ -152,6 +160,28 @@ rsyslog_conf_ovirt_formatting:
               }
           }
           {% endif %}
+
+          if not ($syslogtag startswith "ovirt" or $syslogtag startswith "collectd") then {
+              # Log all kernel messages to the console.
+              # Logging much else clutters up the screen.
+              #kern.*                                                 /dev/console
+              # Log anything (except mail) of level info or higher.
+              # Don't log private authentication messages!
+              *.info;mail.none;authpriv.none;cron.none                /var/log/messages
+              # The authpriv file has restricted access.
+              authpriv.*                                              /var/log/secure
+              # Log all the mail messages in one place.
+              mail.*                                                  -/var/log/maillog
+
+              # Log cron stuff
+              cron.*                                                  /var/log/cron
+              # Everybody gets emergency messages
+              *.emerg                                                 :omusrmsg:*
+              # Save news errors of level crit and higher in a special file.
+              uucp,news.crit                                          /var/log/spooler
+              # Save boot messages also to boot.log
+              local7.*                                                /var/log/boot.log
+        }
 
 rsyslog_rulebase_ovirt_engine:
 
--- a/roles/rsyslog/roles/output_roles/elasticsearch/defaults/main.yaml
+++ b/roles/rsyslog/roles/output_roles/elasticsearch/defaults/main.yaml
@@ -77,7 +77,7 @@ rsyslog_conf_es_elasticsearch:
           }
 
           ruleset(name="try_es") {
-              if strlen($.omes!status) > 0 then {
+              if (strlen($.omes) > 0) and (strlen($.omes!status) > 0) then {
                   # retry case
                   if ($.omes!status == 200) or ($.omes!status == 201) or (($.omes!status == 409) and ($.omes!writeoperation == "create")) then {
                       stop # successful
@@ -94,14 +94,14 @@ rsyslog_conf_es_elasticsearch:
               }
               if strlen($!es_msg_id) > 0 then {
                   set $.es_msg_id = $!es_msg_id;
-              } else if strlen($.omes!_id) > 0 then {
+              } else if (strlen($.omes) > 0) and (strlen($.omes!_id) > 0) then {
                   # retry
                   set $.es_msg_id = $.omes!_id;
               } else {
                   # NOTE: depends on rsyslog being compiled with --enable-uuid
                   set $.es_msg_id = $uuid;
               }
-              if strlen($.omes!_index) > 0 then {
+              if (strlen($.omes) > 0) and ( strlen($.omes!_index) > 0) then {
                   # retry
                   set $.index_name = $.omes!_index;
               } else {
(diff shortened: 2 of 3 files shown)


roles/rsyslog/roles/output_roles/elasticsearch/defaults/main.yaml
```diff
@@ -77,7 +77,7 @@ rsyslog_conf_es_elasticsearch:
           }
 
           ruleset(name="try_es") {
-              if strlen($.omes!status) > 0 then {
+              if (strlen($.omes) > 0) and (strlen($.omes!status) > 0) then {
```
richm: Also these two cases:
```
              } else if strlen($.omes!_id) > 0 then {
```
and
```
              if strlen($.omes!_index) > 0 then {
```

sradco: This is already in the code ...
https://github.com/linux-system-roles/logging/blob/059d70fe43ca78b497c87f606e1ed133cf373867/roles/rsyslog/roles/output_roles/elasticsearch/defaults/main.yaml#L97
https://github.com/linux-system-roles/logging/blob/059d70fe43ca78b497c87f606e1ed133cf373867/roles/rsyslog/roles/output_roles/elasticsearch/defaults/main.yaml#L104
@richm

roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
```diff
@@ -153,6 +161,28 @@ rsyslog_conf_ovirt_formatting:
...
+              # Log cron stuff
+              cron.*                                                  /var/log/cron
+              # Everybody gets emergency messages
+              *.emerg                                                 :omusrmsg:*
+              # Save news errors of level crit and higher in a special file.
+              uucp,news.crit                                          /var/log/spooler
+              # Save boot messages also to boot.log
+              local7.*                                                /var/log/boot.log
+        }
+
```
sradco: @richm @didib @pcahyna Please review this patch.
I have a problem in the design with this. If I use the default rsyslog.conf the collectd logs go to /var/log/messages.
I needed to move the config to this section to avoid getting the collectd logs getting to /var/log/messages.
But this is really not a generic fix. 
I can add it here for now but would prefer to replace it to a more generic solution.

roles/rsyslog/roles/output_roles/elasticsearch/defaults/main.yaml
richm: > This is already in the code ...

Not sure what you mean

You cannot check for `strlen($.omes!_id)` or `strlen($.omes!_index)` without first checking `strlen($.omes)`

sradco: Sorry. I fixed it now @richm

roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
nhosoi: @sradco, I'd think what is implemented in https://github.com/sradco/logging/blob/updatedefaultconf/roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml#L164-L184 is not input, but action/output.

(Sorry, I haven't tested) but the code written in the old format could be interpreted to the current syntax as follows.  They are either omfile or omusrmsg action.  So, I guess you could place it in the output roles?  And if its input is not from viaq, they don't have $syslogtag (and $syslogfacility or $syslogseverity) at all?  (@richm, please correct me if I'm wrong).    Then, it'd be safe to have a default output role in output_roles and include it?

```
if strlen($syslogtag) == 0 then {
    if $facility == "kern" then {
        action(type="omfile" template="RSYSLOG_FileFormat" file="/dev/console")
    } else if $facility == "mail" then {
        action(type="omfile" template="RSYSLOG_FileFormat" file="/var/log/maillog")
    } else if $facility == "authpriv" then {
        action(type="omfile" template="RSYSLOG_FileFormat" file="/var/log/secure")
    } else if $facility == "cron" then {
        action(type="omfile" template="RSYSLOG_FileFormat" file="/var/log/cron")
    } else if $facility == "local7" then {
        action(type="omfile" template="RSYSLOG_FileFormat" file="/var/log/boot.log")
    } else {
        action(type="omfile" template="RSYSLOG_FileFormat" file="/var/log/messages")
    }
 
    if $severity == "emerg" then {
        action(type="omusrmsg" users="*")
    }
    if (($facility == "uucp") or ($facility == "news")) and ($severity == "crit") then {
        action(type="omfile" template="RSYSLOG_FileFormat" file="/var/log/spooler")
    }
}
```

sradco: @nhosoi Yes. you are correct. I did not add this to the correct location. It should defiantly be in output roles.
logs coming from journal don't have  $syslogtag ?

richm: > they don't have $syslogtag (and $syslogfacility or $syslogseverity) at all?

I don't know, but you could easily check.  But if there is an easy way to check system messages like @nhosoi proposes, that would be best.

roles/rsyslog/roles/output_roles/elasticsearch/defaults/main.yaml
richm: looks good

roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
sradco: I see that debops do have this rules already here:
https://github.com/linux-system-roles/logging/blob/059d70fe43ca78b497c87f606e1ed133cf373867/roles/rsyslog/roles/input_roles/debops/defaults/main.yaml#L166

How would you differentiate between a case we wants default rsyslog.conf that includes these rules and a case that we other input roles like viaq/ovirt and also need these rules?
We currently have only 1 variable "rsyslog_default" and it is true by default.

I think we can filter by $.logs_collection. If it is no set then go to these rules.

@nhosoi

richm: I think that some users of ovirt will want ovirt and collectd logs to go to `/var/log/messages`, and some will not.   It is probably a good idea for ovirt to configure rsyslog by default to not send these logs to `/var/log/messages`.

sradco:
Can we move 
# Include all config files in /etc/rsyslog.d/
$IncludeConfig /etc/rsyslog.d/*.conf
To the bottom of the rsysloc.conf file when using default rsysloc.conf?
@richm @nhosoi @jvymazal

jvymazal:
Well, it depends on use case, and what exactly are you including. Whether there are some global statements, module loading or anything else which could influence configuration in the main file. If you have some way to flag/recognize them (I expect you do as you have to route them somewhere) you can just put `& stop` behind the processing of those messages to prevent them to go to other filters.

sradco:
@nhosoi @richm  Can you recommend where to put the "stop"  in the elasticsearc output role?

richm:
I don't think we want to put a `stop` in the elasticsearch output role.  What if the user wants to send records to both elasticsearch and some other output?

PR #77, opened 2019-07-31
Add send_targets_only variable

sradco:
Added send_targets_only variable and
40-send-targets-only.conf config file.

The config file contains a filter that will filter
logs with 'send_targets_only' set to 'true'
and discard them after all output targets were processed
and before the default rsyslog.conf rules.

Please review: @pcahyna @nhosoi

### DIFF
--- a/roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
+++ b/roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
@@ -42,6 +42,9 @@ rsyslog_conf_local_ovirt_modules:
 
   - name: 'local-ovirt-modules'
     type: 'modules'
+    state: '{{ "present"
+              if not (rsyslog_default|default(true)|bool)
+              else "absent" }}'
     sections:
 
       - comment: 'Log messages sent to local UNIX socket with use=off'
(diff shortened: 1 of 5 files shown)


roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
```diff
@@ -46,7 +46,7 @@ rsyslog_conf_local_ovirt_modules:
 
       - comment: 'Log messages sent to local UNIX socket with use=off'
         options: |-
-          module(load="imuxsock" SysSock.use="off")
+          #module(load="imuxsock" SysSock.use="off")
```
nhosoi: an accident?

nhosoi:
Other than the only one question above, looks good to me.

roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
sradco: This setting does not work with default rsyslog conf. We are getting an error message. @nhosoi

nhosoi: I see since it's already there...

https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/templates/etc/rsyslog.conf.j2#L19-L20

I wonder could you generate the file only when rsyslog_default is false?

sradco: Thank you. I updated the state of the file instead with the condition as you suggested
Need to consider if we should move this to the rsyslog_common_rules and remove from viaq, debops and ovirt sub-roles. @nhosoi

PR #79, opened 2019-08-04
Add journald input role

sradco:
dded journald input role.
Now user can configure sending journald
logs to the supported outputs.

Please review: @richm  @nhosoi @pcahyna 
    
### DIFF
--- a/roles/rsyslog/roles/input_roles/journal/defaults/main.yaml
+++ b/roles/rsyslog/roles/input_roles/journal/defaults/main.yaml
@@ -1,0 +1,116 @@
+---
+# journal configuration
+# ---------------------
+
+# journal rpm packages
+# adding rsyslog_logging_packages
+# ------------------
+
+# .. envvar:: rsyslog_logging_packages
+#
+# List of rpm packages for Common Logging.
+rsyslog_journal_packages: ['rsyslog-mmnormalize', 'rsyslog-mmjsonparse', 'libfastjson', 'liblognorm', 'libestr']
+rsyslog_journal_prereq_packages: []
+
+# Available configuration set
+# ----------------------------
+
+rsyslog_unprivileged: 'False'
+
+# journal Rsyslog configuration rules
+# ---------------------------
+
+rsyslog_journal_rules:
+  - '{{ rsyslog_conf_journal_main_modules }}'
+  - '{{ rsyslog_conf_journal_formatting }}'
+  - '{{ rsyslog_rulebase_viaq_parse_json }}'
+
+# Default journal configuration options
+# -----------------------------------
+
+rsyslog_conf_journal_main_modules:
+
+  - name: 'journal_main'
+    type: 'modules'
+    sections:
+
+      - options: |-
+          # Read from journal
+          module(load="imjournal" StateFile="{{ rsyslog_work_dir }}/imjournal.state" RateLimit.Burst="{{ imjournal_ratelimit_burst|default(1000000) }}" RateLimit.Interval="{{ imjournal_ratelimit_interval|default(10) }}" PersistStateInterval="{{ imjournal_persist_state_interval|default(1000) }}" )
+
+      - options: |-
+          # Normalize logs
+          module(load="mmnormalize")
+
+rsyslog_conf_journal_formatting:
+
+  - name: 'journal_formatting'
+    type: 'template'
+    path: '{{rsyslog_config_dir }}'
+    sections:
+
+      - options: |-
+          # formatting
+          template(name="cnvt_to_viaq_timestamp" type="list") {
+              property(name="TIMESTAMP" dateFormat="rfc3339")
+          }
+
+          # rsyslog 8.30.0 and later does case insensitive variable name comparison
+          # which means $!MESSAGE is the same as $!message - HOWEVER - the case
+          # of the variable name is preserved when outputting, so we need to "normalize"
+          # to all lower case so that the internal JSON to string conversion will output
+          # "message" in the outgoing record
+          if $inputname == 'imjournal' then {
+              if strlen($!MESSAGE) > 0 then {
+                  action(type="mmnormalize" ruleBase="{{ rsyslog_config_dir }}/parse_json.rulebase" variable="$!MESSAGE")
+                  # ensure that $!message is set and $!MESSAGE is unset
+                  # if rsyslog is case sensitive, then $!MESSAGE == $!message
+                  if $!MESSAGE == $!message then {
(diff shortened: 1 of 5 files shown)


roles/rsyslog/roles/input_roles/journald/defaults/main.yaml
```diff
@@ -0,0 +1,367 @@
...
+
+rsyslog_conf_journald_main_modules:
+
+  - name: 'journald_main'
+    type: 'modules'
+    sections:
+
+      - options: |-
+          # Read from journal
+          module(load="imjournal" StateFile="{{ rsyslog_work_dir }}/imjournal.state" RateLimit.Burst="{{ imjournal_ratelimit_burst|default(1000000) }}" RateLimit.Interval="{{ imjournal_ratelimit_interval|default(10) }}" PersistStateInterval="1000" )
```
richm: You will want to parameterize the `PersistStateInterval` too - you can use `"1000"` as the default value.
You will want to add support for the other parameters too.
The RHEL7 parameters are `IgnorePreviousMessages` `DefaultSeverity` `DefaultFacility` `usepidfromsystem` `WorkAroundJournalBug`
RHEL8 adds support for `UsePid` `IgnoreNonValidStatefile`
The parameters are documented here - https://www.rsyslog.com/doc/v8-stable/configuration/modules/imjournal.html

roles/rsyslog/roles/input_roles/journald/defaults/main.yaml
```diff
@@ -0,0 +1,367 @@
...
+
+      - options: |-
+          # Normalize logs
+          module(load="mmnormalize")
```
richm: This and the following aren't really generic journald formatting/normalizing - this is very specific to the viaq common logging/openshift use case (which is the ovirt journald -> common logging use case as well).  It is perfectly acceptable to pass the journald logs as-is to the destination, with no further processing.  The `prio_to_level` lookup table is useful to have here, but again, not everyone will want to use it - they may want to pass the raw `PRIORITY` numerical value to the ultimate log destination.

roles/rsyslog/roles/input_roles/journal/defaults/main.yaml
```diff
@@ -0,0 +1,116 @@
...
+                  if $!MESSAGE == $!message then {
+                      set $.save_message = $!message;
+                      unset $!MESSAGE; # which also unsets $!message
+                      set $!message = $.save_message;
+                      unset $.save_message;
+                  } else {
+                      if strlen($!message) == 0 then {
+                          set $!message = $!MESSAGE;
+                      }
+                      unset $!MESSAGE;
```
sradco: @richm Is setting "MESSAGE" and "log"  to "message" required?

roles/rsyslog/roles/input_roles/journal/defaults/main.yaml
```diff
@@ -0,0 +1,116 @@
...
+# ---------------------
+
+# journal rpm packages
+# adding rsyslog_logging_packages
+# ------------------
+
+# .. envvar:: rsyslog_logging_packages
+#
+# List of rpm packages for Common Logging.
+rsyslog_journal_packages: ['rsyslog-mmnormalize', 'rsyslog-mmjsonparse', 'libfastjson', 'liblognorm', 'libestr']
```
sradco: @richm Do we need all these packages?

roles/rsyslog/roles/input_roles/journal/defaults/main.yaml
```diff
@@ -0,0 +1,116 @@
...
+              }
+
+              if strlen($!@timestamp) == 0 then {
+                  if strlen($!time) > 0 then {
+                      set $!@timestamp = $!time; # assumes json-file log timestamp in rfc3339 format already
+                      unset $!time;
+                  } else {
+                      set $!@timestamp = exec_template('cnvt_to_viaq_timestamp');
+                  }
+              }
```
sradco: @richm  Do you think this is the correct way to set the "@timestamp" ?

roles/rsyslog/roles/input_roles/journal/defaults/main.yaml
richm: It is for output to viaq/openshift common logging.  It is not required if you are not using viaq/openshift common logging.

roles/rsyslog/roles/input_roles/journal/defaults/main.yaml
richm: Yes, if you are doing processing for output to viaq/openshift common logging.  They are not necessarily needed for other processing, although anything except trivial parsing will require rsyslog-mmnormalize, and the `lib` packages are dependencies of rsyslog-mmnormalize

richm: rsyslog-mmjsonparse is needed primarily for docker json-file logs, or if you have nested JSON in your log/message/MESSAGE field you want to parse (although rsyslog-mmnormalize can also parse JSON)

roles/rsyslog/roles/input_roles/journal/defaults/main.yaml
richm: It is for the viaq/openshift common logging case.  In other cases, you might not need to output a `@timestamp` field (although that is the default time field for Elasticsearch), and you might not have a `time` field (which is specific to docker json-file logs).

roles/rsyslog/roles/input_roles/journal/defaults/main.yaml
sradco: @pcahyna Hi Pavel, Do you think that we should send the Journal logs as is or format the message at least to have "message" instead of "Message" or "log"(in case its docker logs) ?

roles/rsyslog/roles/input_roles/journal/defaults/main.yaml
sradco: @pcahyna @richm In that case, I think we should keep all of them. Do you agree?
Are they available on the RHEL channels? Do we need to mention anything in the documentation in order for the users to be able to consume them?

roles/rsyslog/roles/input_roles/journal/defaults/main.yaml
sradco: @pcahyna @richm Do we want to support the docker json-file logs? Are they logged to journal?

nhosoi:
Journald input was implemented as part of the basic input. Closing this pr.

PR #80, opened 2020-01-03
RHELPLAN-31390: Refactoring the basic input and output configuration …

nhosoi:
…in Linux-System-Roles/Logging

    RHELPLAN-31390: Refactoring the basic input and output configuration in Linux-System-Roles/Logging
    - renaming input_roles/debops with input_roles/basics.
    - adding output_roles/{files,forwards}.
    - moving system and remote templates and possibly rulesets to
      output_roles/{files,forwards}/defaults/main.yaml.
    - adding a unique name to every input and action.
    - updating the README file to describe the changes.
    
    Supporting both python2 and 3.
    - molecule Dockerfile.j2 - replacing with the latest upstream Dockerfile.j2
      https://github.com/ansible/molecule/blob/master/molecule/data/Dockerfile.j2
    - adding 'MOLECULE_DISTRO="centos:8"' to .travis.yml
    - adding "libselinux-python" or "python3-libselinux" based on the python version.
      and put it into the rsyslog_debug true condition since the libselinux is needed
      just by lineinfile.


molecule/default/Dockerfile.j2
```diff
@@ -7,7 +7,8 @@ FROM {{ item.image }}
 {% endif %}
 
 RUN if [ $(command -v apt-get) ]; then apt-get update && apt-get install -y python sudo bash ca-certificates && apt-get clean; \
-    elif [ $(command -v dnf) ]; then dnf makecache && dnf --assumeyes install python sudo python-devel python2-dnf bash && dnf clean all; \
+    elif [ $(command -v dnf) ] && [ $(dnf search python2-dnf) ]; then dnf makecache && dnf --assumeyes install python sudo python-devel python2-dnf bash && dnf clean all; \
+    elif [ $(command -v dnf) ] ; then dnf makecache && dnf --assumeyes install python3 sudo python3-devel python3-dnf bash && dnf clean all; \
```
tyll: It might make sense to synchronize the python3 case with the upstream file: https://github.com/ansible/molecule/blob/master/molecule/data/Dockerfile.j2

roles/rsyslog/roles/input_roles/basics/defaults/main.yaml
```diff
@@ -0,0 +1,182 @@
+---
+# Debops logs configuration set
```
richm: remove `Debops` here

roles/rsyslog/roles/input_roles/basics/defaults/main.yaml
```diff
@@ -0,0 +1,182 @@
...
+
+  - name: 'local-basics-modules'
+    type: 'modules'
+    sections:
+
+      - comment: 'Log messages sent to local UNIX socket'
+        options: |-
+          module(load="imuxsock" SysSock.use="on" )
+
+      - comment: 'Send messages over UDP or plain TCP without TLS'
```
richm: s/Send messages/Read messages sent/

roles/rsyslog/templates/etc/rsyslog.conf.j2
```diff
@@ -1,49 +1,56 @@
 # {{ ansible_managed }}
 
-{% if not rsyslog_default|default(true)|bool %}
+{% if rsyslog_use_roles|default(false)|bool %}
```
richm: So this is the default rhel7 rsyslog config file, just like the one provided by the rsyslog rpm package?  How different is it from the default on rhel8?

tasks/main.yaml
```diff
@@ -7,6 +7,20 @@
...
 
+    - name: Set rsyslog_use_roles
+      set_fact:
+        rsyslog_use_roles: "{{ logging_use_roles|d(false) }}"
+
+    - name: Set rsyslog_purge_original_conf
+      set_fact:
+        rsyslog_purge_original_conf: "{{ logging_purge_confs|d(true) }}"
+
+    - name: Set ansible_python_interpreter if python3 is used
```
richm: Is it necessary to do this here?  Seems like something that should be done outside of this role.

richm: Is `ansible_python_interpreter` used only by this role?

richm:
All `input` and `action` should have a unique name.  This is very useful when impstats is enabled, and for debugging.

tasks/main.yaml
pcahyna: Thanks @richm for spotting this. `ansible_python_interpreter` must be set by the environment, (outside roles/playbooks/tasks) otherwise nothing works, sio it really does not belong there. Ansible 2.8 and later autodetects it on RHEL8.

molecule/default/Dockerfile.j2
nhosoi: Thanks, @tyll.  Indeed, we should have the upstream file.  In my local testing, I replaced  `logging/molecule/default/Dockerfile.j2` with the upstream file, then there was no problem to run `molecule converge` against centos7 and 8.  Now, does that mean since python2 EOL has passed, we can only support python3, i.e., run the tests with python3 from now on?

Having said that, there's one thing left that confuses me...  I put this debug print in the LSR/Logging tasks/main.yaml.
```
+    - name: Check python version
+      debug:
+        msg: "Python version is {{ ansible_python_version }}"
```
Then it printed the version was 2.7.5 for centos7, while 3.6.8 for centos8.  I thought molecule/default/Dockerfile.j2 installed python3 on the centos7 target host, but it was not???

roles/rsyslog/templates/etc/rsyslog.conf.j2
nhosoi: Thanks, @richm .  Per discussion on IRC, eliminating rsyslog_use_roles and changing the behavior to NOOP if no LSR/Logging variables are given to the inventory file.

roles/rsyslog/templates/etc/rsyslog.conf.j2
```diff
@@ -1,49 +1,56 @@
...
 #### MODULES ####
 
-module(load="imuxsock"    # provides support for local system logging (e.g. via logger command)
-       SysSock.Use="off") # Turn off message reception via local log socket;
-       # local messages are retrieved through imjournal now.
-module(load="imjournal"   # provides access to the systemd journal
-       StateFile="imjournal.state") # File to store the position in the journal
-#module(load="imklog") # reads kernel messages (the same are read from journald)
-#module(load="immark") # provides --MARK-- message capability
+# The imjournal module bellow is now used as a message source instead of imuxsock.
```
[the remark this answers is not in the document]
nhosoi: This part is going to be eliminated...

tasks/main.yaml
```diff
@@ -11,6 +11,15 @@
...
 
+    - name: Set use_python3 to false if python3 is not installed.
+      set_fact:
+        use_python3: false
+      when: "ansible_python_version is version('3', '<')"
+
+    - name: Check python version
+      debug:
+        msg: "Python version is {{ ansible_python_version }}; use_python3 is {{ use_python3 }}"
+
```
pcahyna: I don't think that this is something specific to the logging role, so if it is really needed, it should be solved for all the roles in a common way.

nhosoi: +1
But do you think my expectation is valid?  That is, we will see ansible_python_version 3.X here if we install python3 on centos 7 with molecule Dockerfile?
If it is, shall I open an issue on JIRA?

defaults/main.yml
```diff
@@ -9,3 +9,5 @@ logging_outputs: []
 logging_custom_config_files: []
 
 logging_purge_confs: false
+
+use_python3: true
```
richm: not sure why this is needed

roles/rsyslog/roles/input_roles/viaq/defaults/main.yaml
```diff
@@ -87,7 +87,7 @@ rsyslog_conf_viaq_formatting:
           # "message" in the outgoing record
           if strlen($!MESSAGE) > 0 then {
               set $!pipeline_metadata!collector!original_raw_message = $!MESSAGE;
-              action(type="mmnormalize" ruleBase="{{ rsyslog_config_dir }}/parse_json.rulebase" variable="$!MESSAGE")
+              action(name="MESSAGE_mmnormalize" type="mmnormalize" ruleBase="{{ rsyslog_config_dir }}/parse_json.rulebase" variable="$!MESSAGE")
```
richm: `viaq_MESSAGE_mmnormalize`

roles/rsyslog/roles/input_roles/viaq/defaults/main.yaml
```diff
@@ -106,7 +106,7 @@ rsyslog_conf_viaq_formatting:
           } else {
               if strlen($!log) > 0 then {
                   set $!pipeline_metadata!collector!original_raw_message = $!log;
-                  action(type="mmnormalize" ruleBase="{{ rsyslog_config_dir }}/parse_json.rulebase" variable="$!log")
+                  action(name="log_mmnormalize" type="mmnormalize" ruleBase="{{ rsyslog_config_dir }}/parse_json.rulebase" variable="$!log")
```
richm: `viaq_log_mmnormalize`

defaults/main.yml
nhosoi: This is a workaround for this issue.  (Please see the conversation attached to the area)
https://github.com/linux-system-roles/logging/pull/80/files#diff-985705db1379cc4bdac80eaa88789109R14-R22

Also please see my comment https://github.com/linux-system-roles/logging/pull/80#discussion_r364941142.  Not sure in which layer (ansible, molecule or the way how I use them...), but there's some confusion in switching the python version 2 to 3...

tasks/main.yaml
pcahyna: First of all, why do you worry about the python version? What is the problem you are trying to solve?

nhosoi: Ok.  What I need to make sure is the target host has the python libselinux package.  When I was testing the first cut of LSR/Logging, the package was not in the target host and my attempt to deploy config failed due to the error [2].  So, I included the package name in rsyslog_base_packages (see [1]).  Unfortunately, the package name was changed from libselinux-python to python3-libselinux starting from RHEL8 [3].  So, I had to add this trick [1] to get the right libselinux package name.  That's the reason why I was worried about the python version.

Regarding the python3 on centos 7, I'm updating molecule Dockerfile.j2 like this [4].  I thought python3 is used on all testing platforms including centos 7, but the ansible log says python2 is being used, which made me confused and asked the question.

[1] https://github.com/linux-system-roles/logging/pull/80/files#diff-def9bf9161e11520d46eea7aa85093e1R59
[2] "msg": "Aborting, target uses selinux but python bindings (libselinux-python) aren't installed!"
[3] https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/considerations_in_adopting_rhel_8/changes-to-packages_considerations-in-adopting-rhel-8
[4] https://github.com/linux-system-roles/logging/pull/80/files#diff-a8de8e86be756e09de886c06dbf13c06

roles/rsyslog/roles/README.md
```diff
@@ -16,8 +16,6 @@ The `defaults` directory includes:
   - It is required to set for all logs the project identfier for pipelining:
     set $.logs_collection = "project name";
     For example: In `ovirt` input role, in the default/main.yaml, for every log `$.logs_collection` is set to `ovirt`.
-  - If rsyslog_default equals to`"true, It is required to set for all logs you don't want to be processed by the default rules:
-    set $.send_targets_only = "true";
```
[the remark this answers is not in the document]
sradco: @nhosoi Why do you remove rsyslog_default?

roles/rsyslog/roles/output_roles/files_forwards/defaults/main.yaml
```diff
@@ -0,0 +1,193 @@
...
+          {% endif %}
+
+# rsyslog_conf_files_output_rules:
+#
+# A set of default ``rsyslog`` options which store local system logs in files
+# located in `{{rsyslog_system_log_dir}}/` directory. This is mostly the
+# same as the default ``rsyslogd`` configuration provided in the installations.
+rsyslog_conf_files_output_rules:
+
+  - name: 'output-files'
```
sradco: @nhosoi What is the benefit of having this in the rsyslog.d/ instead of the rsyslog.conf.
If we want to keep this here we need a way to filter messages that should not go through these filters.

tasks/main.yaml
```diff
@@ -17,14 +30,23 @@
...
         - item.1.state|d('present') == 'present'
 
+    - name: Set rsyslog_files_and_forwards
+      set_fact:
+        rsyslog_files_and_forwards: "{{ rsyslog_files_and_forwards|d([]) }} + {{ [  item.0|combine( {'logs_collections_name': item.1.name} )] }}"
+      with_subelements:
+        - "{{ logging_outputs }}"
+        - logs_collections
+      when:
+        - item.1.state|d('present') == 'present'
```
sradco: @nhosoi I believe you are missing a when condition here.
- item.0.type|d('') == 'files_forwards'

sradco:
@noriko There are a few issues here that affect oVirt.
Please don't merge before I test and see this does not create a regression for us or before I make the required changes for it to work for us.

roles/rsyslog/roles/output_roles/files_forwards/defaults/main.yaml
```diff
@@ -0,0 +1,193 @@
...
+          kern.*                                                 /dev/console
+        state: '{{ "present"
+                  if ("kernel-message" in rsyslog_capabilities)
+                  else "absent" }}'
+
+      - comment: |-
+          Log anything (except mail) of level info or higher.
+          Don't log private authentication messages!
+        options: |-
+          *.info;mail.none;authpriv.none;cron.none                {{rsyslog_system_log_dir}}/messages
```
sradco: @nhosoi Why is each filter in a separate comment?

sradco: Also, you will have here duplicates if:
"network" in rsyslog_capabilities and "remote-files" in rsyslog_capabilities.

README.md
```diff
@@ -73,22 +67,55 @@ vars.yaml
...
+logging_purge_confs: True
+```
 
+2) Deploying basic LSR/Logging config files in /etc/rsyslog.d.
+```
+logging_enabled: True
+logging_purge_confs: True
+logging_outputs:
+  - name: files-and-forwards
+    type: files_forwards
```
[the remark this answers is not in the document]
nhosoi: > @nhosoi I'm not sure what this output type does. Do you have documentation for it?

No specific doc yet.  Sorry.  Basically, files_forwards implements the rsyslog configuration which is identical to the one in rpm in the LSR/logging manner.  Plus, forwarding is appended.

As I put in the title, it's for refactoring the existing basic role, and no plan to add new functionalities.

README.md
```diff
@@ -101,40 +128,38 @@ logging_outputs:
...
       - name: 'ovirt'
     server_host: logging-es-ovirt
     server_port: 9200
     index_prefix: project.ovirt-logs
-    ca_cert:
-    cert:
-    key:
-  - name: custom_files-test
-    type: custom_files
-    custom_config_files: [ '/path/to/custom_A.conf', '/path/to/custom_B.conf' ]
```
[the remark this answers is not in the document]
nhosoi: > @nhosoi Hi, You remove the custom_files. Was that on purpose? BTW, We need to better document each input/output ..

Resurrected.  +1 to the documentation.  Maybe, we could add README to https://github.com/nhosoi/logging/tree/27211/roles/rsyslog/roles/input_roles/ and https://github.com/nhosoi/logging/tree/27211/roles/rsyslog/roles/output_roles?  Do you have any good idea about the format and contents to be placed?

README.md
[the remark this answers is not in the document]
nhosoi: > @nhosoi I'm not sure what this output type does. Do you have documentation for it?

Well, according to your input_roles/output_roles framework, the directory name under output_roles is treated as a type, isn't it?  I created `files_forwards` to handle outputs into files and/or forwarding to other hosts in [1], so the name `files_forwards` appears there.
[1] - https://github.com/nhosoi/logging/tree/27211/roles/rsyslog/roles/output_roles

roles/rsyslog/roles/README.md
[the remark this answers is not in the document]
nhosoi: @sradco:
> @nhosoi Why do you remove rsyslog_default?

Please see this JIRA card.
https://projects.engineering.redhat.com/browse/RHELPLAN-31390?focusedCommentId=2055981&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2055981

It may be a bit confusing to follow, but first I renamed rsyslog_default with rsyslog_use_roles (having the opposite boolean), then I got a question from @pcahyna why this is needed.  In the discussion, it's concluded since by rerunning yum reinstall, we could achieve the goal, LSR/Logging does not have to do it.  Also, any other LSR subroles do not do that.  It seems it somehow "violating" the LSR policy.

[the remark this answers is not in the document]
nhosoi: @sradco:
> We needed a way to keep the default filters from rsyslog.conf for the system and env logs,

You mean you use the original rsyslog.conf from the rsyslog rpm with the config files in /etc/rsyslog.d/ deployed by the LSR/Logging?  That is, ovirt is deploying config files with the original rsyslog.conf.

> but we needed to stop the other logs, that we sent to elasticsearch for example, from being logged also to journald.

I think the original rsyslog.conf could be replaced with files_forwards output.  Currently, the type is `output`, but it actually contains rules, so it's more appropriate to have `rules` type, which is 50.
https://github.com/linux-system-roles/logging/pull/80/files#diff-c30f197e75c01b82b3a0dda13e1ef238R36

roles/rsyslog/roles/output_roles/files_forwards/defaults/main.yaml
nhosoi: @sradco:
> @nhosoi What is the benefit of having this in the rsyslog.d/ instead of the rsyslog.conf.
> If we want to keep this here we need a way to filter messages that should not go through these filters.

Our final goal is to manage the rsyslog configuration from LSR/Logging.  That includes logging the system logs into the local files, forwarding them to the rsyslog server on the remote host or other servers or message queue, etc.  I thought all the parts are preferable to handle in the same way instead of making the output to the local file specially.  

Also, we'd like to make the upgrade the configuration as easy as possible.  For instance, if a customer modifies rsyslog.conf as they wish, it might be hard for us to keep track of the changes and apply them to the new version of rsyslog.conf.  The update issue has not been fully solved yet, though.  I'm afraid all in one rsyslog.conf could make the task harder.

roles/rsyslog/roles/output_roles/files_forwards/defaults/main.yaml
nhosoi: @sradco:
> @nhosoi Why is each filter in a separate comment?

That's only because I wanted to have the same comments in the original rsyslog.conf in the files_forwards output role.
```
#### RULES ####

# Log all kernel messages to the console.
# Logging much else clutters up the screen.
#kern.*                                                 /dev/console

# Log anything (except mail) of level info or higher.
# Don't log private authentication messages!
*.info;mail.none;authpriv.none;cron.none                /var/log/messages

....
```
> Also, you will have here duplicates for all the filters if:
> "network" in rsyslog_capabilities and "remote-files" in rsyslog_capabilities.

Could you elaborate?  I ran ansible-playbook with the following vars, but I only see one 30-output-files.system and there's no duplicates [1].
```
logging_enabled: true
rsyslog_capabilities: [ 'network', 'remote-files' ]
rsyslog_unprivileged: False
logging_purge_confs: True
rsyslog_backup_dir: /home/nhosoi/tmp/backup
logging_debug: True
logging_outputs:
  - name: files-and-forwards
    type: files_forwards
    logs_collections:
      - name: basics
        state: 'present'
```
[1]
```
# egrep '*.info;mail.none;authpriv.none;cron.none' *
30-output-files.system:*.info;mail.none;authpriv.none;cron.none                /var/log/messages
```

roles/rsyslog/templates/etc/rsyslog.conf.j2
```diff
@@ -1,91 +1,6 @@
 # {{ ansible_managed }}
 
-{% if not rsyslog_default|default(true)|bool %}
 #
 # Include all config files in {{rsyslog_config_dir}}
 #
 $IncludeConfig {{rsyslog_config_dir}}/*.conf
-
-{% else %}
-
```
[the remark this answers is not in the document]
nhosoi: @richm, @pcahyna, the use case of ovirt is described by @sradco here.  They deploy the original rsyslog.conf with the ovirt specific configurations in rsyslog.d.  And it relies on rsyslog.conf for logging the system logs into the local files.

@sradco, I still think we should improve files_forwards to make it more generic and useful to satisfy your goals.  For instance, you may want to add more ability to configure imjournal for tuning up.  Having rsyslog.conf makes it harder.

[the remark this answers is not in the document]
richm: >  They deploy the original rsyslog.conf with the ovirt specific configurations in rsyslog.d.

So ovirt does not touch at all the original /etc/rsyslog.conf?  Then reverting to the original is just removing the contents of /etc/rsyslog.d, then restoring the original /etc/rsyslog.d/21-cloudinit.conf and /etc/rsyslog.d/listen.conf?

nhosoi:
@sradco:
> @noriko There are a few issues here that affect oVirt.
> Please don't merge before I test and see this does not create a regression for us or before I make the required changes for it to work for us.

Sure, I won't merge this pr without your approval.  Don't worry.  I think it is a good opportunity to improve the design of LSR/Logging.

BTW, do we want to create a branch like logging-0.1 from the current master and use it as a stable branch?

nhosoi:
Another /hold.
Considering this task, we'd better separate forwards from files since we should be able to send message type A to files and message type B to forwards...
https://projects.engineering.redhat.com/browse/RHELPLAN-32351
Add an ability to distribute messages to different outputs based on severity and facility

tasks/main.yaml
nhosoi: Investigated some more on this python selinux library issue and figured out the logging_debug added in this commit required libselinux-python for rhel7 and python3-libselinux for rhel8.
https://github.com/linux-system-roles/logging/commit/3bc23d034a8deee0ab34eeaae64a6f58cb62e44e

We may be able to avoid the selinux issue if we choose the path carefully...

roles/rsyslog/templates/etc/rsyslog.conf.j2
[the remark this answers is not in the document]
sradco: @richm  Currently the logging role does not touch the original /etc/rsyslog.conf except maybe for adding the 
# Include all config files in {{rsyslog_config_dir}}
$IncludeConfig {{rsyslog_config_dir}}/*.conf

We do not restore /etc/rsyslog.d/21-cloudinit.conf and /etc/rsyslog.d/listen.conf since we don't use them.

First we had a bug that we where missing the default output filters that saved the system and env logs to the local files, since we enheritted the role from OpenShift that did not require them.
When I tried to fix the issue I also tried to generalize the default output filters, but I hit the issues I mentioned , that all logs where also saved to journald and that was not the behavior we wanted.
So at this point "fixed" it by using the /etc/rsyslog.conf that also handles the cleanup issue and added a filter before it that stops the specific logs that has the the flag "send_targets_only" as "true".

@nhosoi If the "rsyslog_default" is set to "false" then the /etc/rsyslog.conf will only include the 
#
# Include all config files in {{rsyslog_config_dir}}
#
$IncludeConfig {{rsyslog_config_dir}}/*.conf

roles/rsyslog/roles/output_roles/files_forwards/defaults/main.yaml
sradco: @nhosoi I agree with you, but we need a way to only collect to the files the relevant logs. 
Currently viaq/ovirt and any other logs will also be saved to the files (usually /var/log/messages).
This is not the desired state, unless for debugging purposes for example.

roles/rsyslog/roles/output_roles/files_forwards/defaults/main.yaml
sradco: > @sradco:
> 
> > @nhosoi Why is each filter in a separate comment?
> 
> That's only because I wanted to have the same comments in the original rsyslog.conf in the files_forwards output role.

@nhosoi I see. I thought you can have them all together as a single paragraph with the comments included.

> 
> ```
> #### RULES ####
> 
> # Log all kernel messages to the console.
> # Logging much else clutters up the screen.
> #kern.*                                                 /dev/console
> 
> # Log anything (except mail) of level info or higher.
> # Don't log private authentication messages!
> *.info;mail.none;authpriv.none;cron.none                /var/log/messages
> 
> ....
> ```
> 
> > Also, you will have here duplicates for all the filters if:
> > "network" in rsyslog_capabilities and "remote-files" in rsyslog_capabilities.
> 
> Could you elaborate? I ran ansible-playbook with the following vars, but I only see one 30-output-files.system and there's no duplicates [1].
> 
> ```
> logging_enabled: true
> rsyslog_capabilities: [ 'network', 'remote-files' ]
> rsyslog_unprivileged: False
> logging_purge_confs: True
> rsyslog_backup_dir: /home/nhosoi/tmp/backup
> logging_debug: True
> logging_outputs:
>   - name: files-and-forwards
>     type: files_forwards
>     logs_collections:
>       - name: basics
>         state: 'present'
> ```
> 
> [1]
> 
> ```
> # egrep '*.info;mail.none;authpriv.none;cron.none' *
> 30-output-files.system:*.info;mail.none;authpriv.none;cron.none                /var/log/messages
> ```
@nhosoi I thought you will have in both 30-remote-dynamic-logs.remote file and in 30-output-files.system.
In the remote use case when rsyslog_capabilities: [ 'network', 'remote-files' ], do you also need the 30-output-files.system?
How do you differentiate between the logs collected on the rsyslog server and the logs that are collected from the remote hosts in the rsyslog server?

roles/rsyslog/templates/etc/rsyslog.conf.j2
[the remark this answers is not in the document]
nhosoi: @sradco:

> @nhosoi If the "rsyslog_default" is set to "false" then the /etc/rsyslog.conf will only include the
```
# Include all config files in {{rsyslog_config_dir}}
$IncludeConfig {{rsyslog_config_dir}}/*.conf
```
Yep, I know...  IMHO, I don't mind resurrecting to install the original rsyslog.conf from the LSR/Logging template.  But, I don't feel like advertizing it loudly.  So, if you don't mind, may we change the variable name rsyslog_default to rsyslog_keep_original_rsyslog_conf or something clearer?  The reason is, again, we don't want the users to modify the rsyslog.conf which would be harder for us to track the changes to apply them in the future upgrade.  Also, if the file is there as is it'd break our (future :) supported use cases such as distributing the messages to the different targets based on the priority (facility or severity), etc...

roles/rsyslog/roles/output_roles/files_forwards/defaults/main.yaml
nhosoi: > @nhosoi I agree with you, but we need a way to only collect to the files the relevant logs.
> Currently viaq/ovirt and any other logs will also be saved to the files (usually /var/log/messages).
> This is not the desired state, unless for debugging purposes for example.

+1
This task is one of such requirements.
https://projects.engineering.redhat.com/browse/RHELPLAN-32351
In the ovirt case, what could be the source of the distribution process?  Tag?  Or some string embedded in the message?  Could you please open a task with the specific requirement in https://projects.engineering.redhat.com/browse/RHELPLAN-27211?

nhosoi:
Hi @sradco,
I've resurrected rsyslog_default with the same default value - 'True' to default.  Could you please review this PR to check this change does not break ovirt deployment?  Thanks, in advance.

roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
```diff
@@ -87,10 +87,10 @@ rsyslog_conf_ovirt_formatting:
               property(name="TIMESTAMP" dateFormat="rfc3339")
           }
 
-          input (type="imtcp" port="{{ rsyslog_read_collectd_port|d(44514) }}")
+          input (name="ovirt_imtcp" type="imtcp" port="{{ rsyslog_read_collectd_port|d(44514) }}")
```
sradco: @nhosoi 
This is failing oVirt configuration
error during parsing file /etc/rsyslog.d/20-ovirt_formatting.conf, on or before line 30: parameter 'name' not known -- typo in config file? [v8.24.0-41.el7_7.2 try http://www.rsyslog.com/e/2207 ]

I believe you can only add name to "action".

roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
```diff
@@ -112,20 +112,20 @@ rsyslog_conf_ovirt_formatting:
           }
 
           {% if collect_ovirt_engine_log %}
-          input(type="imfile" file="{{ rsyslog_system_log_dir }}/ovirt-engine/engine.log" tag="ovirt.engine" reopenOnTruncate="on"
+          input(name="ovirt_engine_imfile" type="imfile" file="{{ rsyslog_system_log_dir }}/ovirt-engine/engine.log" tag="ovirt.engine" reopenOnTruncate="on"
```
sradco: @nhosoi same as above

roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
```diff
@@ -112,20 +112,20 @@ rsyslog_conf_ovirt_formatting:
...
 
           if $syslogtag == 'ovirt.engine' then {
-            action(type="mmnormalize" rulebase="{{ rsyslog_config_dir }}/ovirt_engine.rulebase" path="$.ovirt")
+            action(name="ovirt-engine_mmnormalize" type="mmnormalize" rulebase="{{ rsyslog_config_dir }}/ovirt_engine.rulebase" path="$.ovirt")
           }
           {% endif %}
 
           {% if collect_ovirt_vdsm_log %}
-          input(type="imfile" file="{{ rsyslog_system_log_dir }}/vdsm/vdsm.log" tag="ovirt.vdsm" reopenOnTruncate="on"
+          input(name="ovirt_vdsm_imfile" type="imfile" file="{{ rsyslog_system_log_dir }}/vdsm/vdsm.log" tag="ovirt.vdsm" reopenOnTruncate="on"
```
sradco: @nhosoi same as above

roles/rsyslog/roles/input_roles/viaq-k8s/defaults/main.yaml
```diff
@@ -58,18 +58,18 @@ rsyslog_conf_viaq_mmk8s_input:
     sections:
 
       - options: |-
-          input(type="imfile" file="{{ rsyslog_viaq_log_dir }}/*.log" tag="kubernetes" addmetadata="on" reopenOnTruncate="on")
+          input(name="viaq-k8s_imfile" type="imfile" file="{{ rsyslog_viaq_log_dir }}/*.log" tag="kubernetes" addmetadata="on" reopenOnTruncate="on")
```
sradco: @nhosoi same as above

roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
nhosoi: Ahhh, got it.  After writing the following note, I noticed the failed line is 30, which is not for imtcp...  Could you please find out what the line 30 has?   The parameter `name` was added for the stats.  So, we'd like to keep them as much as possible. :)

==========================================================================================
Very strange...  I have parameter `name` in basics role, as well.  And I have no problem to start rsyslog [1]. 
```
module(load="imtcp")
........
input(
  name="basics_imtcp"
  type="imtcp"
  port="514"
)
```
Here's the doc of imtcp input parameters, where `name` is supported.
https://www.rsyslog.com/doc/v8-stable/configuration/modules/imtcp.html#input-parameters
Please note that I'm running the same version of rsyslog...
```
# rpm -q rsyslog
rsyslog-8.24.0-41.el7_7.2.x86_64
```
[1]
```
# systemctl status rsyslog -l
● rsyslog.service - System Logging Service
   Loaded: loaded (/usr/lib/systemd/system/rsyslog.service; enabled; vendor preset: enabled)
   Active: active (running) since Sun 2020-01-26 10:06:00 PST; 2min 27s ago
     Docs: man:rsyslogd(8)
           http://www.rsyslog.com/doc/
 Main PID: 19367 (rsyslogd)
    Tasks: 9
   CGroup: /system.slice/rsyslog.service
           └─19367 /usr/sbin/rsyslogd -n

Jan 26 10:06:00 momo7.localdomain systemd[1]: Stopped System Logging Service.
Jan 26 10:06:00 momo7.localdomain systemd[1]: Starting System Logging Service...
Jan 26 10:06:00 momo7.localdomain rsyslogd[19367]:  [origin software="rsyslogd" swVersion="8.24.0-41.el7_7.2" x-pid="19367" x-info="http://www.rsyslog.com"] start
Jan 26 10:06:00 momo7.localdomain systemd[1]: Started System Logging Service.
```

roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
nhosoi: Removed `name`.

roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml
nhosoi: Removed `name`.

roles/rsyslog/roles/input_roles/viaq-k8s/defaults/main.yaml
nhosoi: Removed `name`.

sradco:
@nhosoi I tested the oVirt use case and it is verified.

defaults/main.yml
pcahyna: @nhosoi where is the issue please? If I remove `use_python3` and the associated istallation of libselinux, the Travis CI build passes. To me this indicates that there has been no issue with Molecule in the first place...  #82

tasks/main.yaml
pcahyna: @nhosoi I am curious how to reproduce the problem, preferably with a test.

nhosoi: @pcahyna 
> @nhosoi I am curious how to reproduce the problem, preferably with a test.

A good question. :)  On my laptop, I happen to have these images [1] and each has libselinux-python-2.5-14.1.el7.x86_64 and python3-libselinux-2.9-2.1.el8.x86_64 preinstalled, respectively.  That means we don't need to install the python libselinux by ourselves.  When we wrote the Logging role, our target host was RHCOS, which is lean enough not to include the python selinux library.  I think the issue is how much we could rely on the preinstalled packages on the target host.  @sradco, in your use case, do you have a chance to run Logging role against a target host which does not have python selinux library?
```
[1]
CentOS-7-x86_64-GenericCloud.qcow2c
CentOS-Stream-GenericCloud-8-20200113.0.x86_64.qcow2
```

pcahyna: Is RHCOS still a target that we need to support? We never had such a request for the other system roles, is logging special in this respect? Does RHCOS even have Python installed?

richm: > Is RHCOS still a target that we need to support? We never had such a request for the other system roles, is logging special in this respect? Does RHCOS even have Python installed?

I think we should not support RHCOS, but let's wait and see if others want to comment.

pcahyna: @sradco do you recall why there was a need to support RHCOS?

tasks/main.yaml
```diff
@@ -56,6 +85,14 @@
         - item.type|d('') == 'custom'
         - logging_enabled|d(true)
 
+    - name: Install python selinux library for logging_debug
+      yum:
+        name: '{{ "python3-libselinux" if use_python3 else "libselinux-python" }}'
+        state: 'latest'
```
pcahyna: package module should have been used here, not yum. https://github.com/oasis-roles/meta_standards/blob/master/README.md#ansible-best-practices

tasks/main.yaml
```diff
@@ -71,7 +108,7 @@
     - name: Delete debug file
       file:
         path: "{{ logging_role_path|d('logging') }}/debug/main.yaml"
-        state: absent
+        state: 'absent'
```
pcahyna: Why this change? Examples of Ansible module usage seem to never enclose such short strings in quotes (applies to many other places like `'latest'` above). See https://docs.ansible.com/ansible/latest/modules/file_module.html . Please do not deviate from examples in docs or existing code unnecessarily.

PR #82, opened 2020-01-30
Remove the use_python3 variable and fact

pcahyna:
- it is redundant: there is ansible_python_version for this
- no other role needs it, so it is doubful that it is actually needed
  (the role does not even contain any Python code)
- it is not even properly namespaced
- should not be in defaults: it is not supposed to be set by user
  and a fact is not supposed to shadow a default
- a fact is not scoped and will persist in the namespace after the role finishes
- libselinux-python is nothing specific to this role and no other role (except
  selinux) installs it - if it is really needed, it should be solved in a common
  way

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -9,5 +9,3 @@ logging_outputs: []
 logging_custom_config_files: []
 
 logging_purge_confs: false
-
-use_python3: true
--- a/tasks/main.yaml
+++ b/tasks/main.yaml
@@ -7,15 +7,6 @@
       set_fact:
         rsyslog_enabled: "{{ logging_enabled|d(true) }}"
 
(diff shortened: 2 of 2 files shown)


richm:
lgtm - @nhosoi - do your tests pass with this PR?

nhosoi:
> libselinux-python is nothing specific to this role and no other role (except selinux) installs it - if it is really needed, it should be solved in a common way

The lineinfile module fails unless the selinux library is installed [1].  If it could be solved "in a common way", can we make it happen for the module?  Please note that Logging role uses the module when logging_debug is true, so it may not have that high priority for the logging roles.  But in general if it prevents to use a particular module like lineinfile, it's not nice, I'd think.
[1] https://github.com/linux-system-roles/logging/blob/master/tasks/main.yaml#L115-L130

PR #83, opened 2020-02-04
RHELPLAN-33869 - Changing the default value of rsyslog_default from t…

nhosoi:
…rue to false

If rsyslog_default is set to true, it deploys the default rsyslog.conf to /etc.
The rsyslog.conf contains the initial input and output plugins for the system logs.
The logging role adds the functionality to process logs including the system logs. 
For instance, which input roles to be used, which output roles to be used, how the
log messages are normalized and distributed.  The default rsyslog.conf could conflict
with the functionality. There is one particular use case which needs rsyslog_default
set to true.  Ovirt already relies on the default rsyslog.conf for the system log
handling. To guarantee it does not break the ovirt logging deployment, we chose to
keep the rsyslog_default var and ability to deploy the default rsyslog.conf for now.

Additionally, Changing logs_collection parameter(s) as follows:
[current]
```
logging_outputs:
  -name: <output_role_name>
   type: <output_role_type>  # type of the output_role
   logs_collections:
     - name: <input_role_name>  # type of the input_role
```
[new]
```
logging_outputs:
  -name: <output_role_name>  # unique name
   type: <output_role_type>  # type of the output_role
   logs_collections:
     - name: <input_role_name>  # unique name
       type: <input_role_type>  # type of the input_role
```

where <input_role_name> in the "current" format specifies the type of
the input_role in the sense of the output_role convention.  To reduce
the confusion, it'd be preferable that "name" and "type" have the same
meaning between output_role (logging_outputs) and input_role (logs_collections).

### DIFF
--- a/molecule/default/playbook.yml
+++ b/molecule/default/playbook.yml
@@ -1,5 +1,7 @@
 ---
 - name: Converge
   hosts: all
+  vars:
+    rsyslog_default: true
   roles:
     - role: logging
--- a/roles/rsyslog/defaults/main.yaml
+++ b/roles/rsyslog/defaults/main.yaml
@@ -9,8 +9,9 @@ rsyslog_enabled: true
 
 # rsyslog_default
 #
-# Diploy default ``rsyslog`` configuration file.
-rsyslog_default: true
+# Diploy default ``rsyslog`` configuration file /etc/rsyslog.conf if it is set to true.
+# Otherwise, rsyslog.conf just includes configuration files in /etc/rsyslog.d.
+rsyslog_default: false
 
 # rsyslog_system_log_dir
 #
@@ -360,7 +361,7 @@ rsyslog_conf_global_options:
     options: |-
       global(
         defaultNetstreamDriver="{{ rsyslog_default_netstream_driver }}"
-      {% if not rsyslog_default | default(true) | bool %}
+      {% if not rsyslog_default | bool %}
         workDirectory="{{ rsyslog_work_dir }}"
       {% endif %}
       {% if rsyslog_pki | bool and "tls" in rsyslog_capabilities %}
--- a/roles/rsyslog/tasks/main.yaml
+++ b/roles/rsyslog/tasks/main.yaml
@@ -6,6 +6,7 @@
   when:
     - rsyslog_enabled | bool
     - not use_rsyslog_image | default(false) | bool
+  ignore_errors: yes
   tags:
     - skip_ansible_lint
 
@@ -159,7 +160,7 @@
 
     - name: Run deploy input sub-roles configs
       include_role:
-        name: "{{ role_path }}/roles/input_roles/{{ input_item.name }}"
+        name: "{{ role_path }}/roles/input_roles/{{ input_item.type }}"
       loop: "{{ rsyslog_logs_collections | flatten(1) }}"
       loop_control:
         loop_var: input_item
@@ -170,7 +171,7 @@
 
     - name: Run remove input sub-roles configs
       include_role:
-        name: "{{ role_path }}/roles/input_roles/{{ input_item.name }}"
+        name: "{{ role_path }}/roles/input_roles/{{ input_item.type }}"
         tasks_from: cleanup.yaml
       loop: "{{ rsyslog_logs_collections | flatten(1) }}"
       loop_control:
--- a/tests/tests_enabled.yml
+++ b/tests/tests_enabled.yml
@@ -2,6 +2,8 @@
 - name: Ensure that the role runs with rsyslog_enabled=true
   hosts: all
   become: true
+  vars:
+    rsyslog_default: True
 
   tasks:
     - name: default run (deploy rsyslog.conf)
--- a/README.md
+++ b/README.md
@@ -105,40 +105,42 @@ logging_purge_confs: true
 2) Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the local system and outputs into the local files.
 ```
 logging_enabled: true
-rsyslog_default: false
 logging_purge_confs: true
 logging_outputs:
   - name: local-files
     type: files
     logs_collections:
-      - name: basics
+      - name: system-input
+        type: basics
 ```
 
 3) Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the local system and remote rsyslog and outputs into the local files.
 ```
 logging_enabled: true
-rsyslog_default: false
 rsyslog_capabilities: [ 'network', 'remote-files' ]
 logging_purge_confs: true
 logging_outputs:
   - name: local-files
     type: files
     logs_collections:
-      - name: basics
+      - name: system-and-remote-input
+        type: basics
 ```
 
 4) Deploying config files for collecting logs from OpenShift pods as well as RHV and forwarding them to elasticsearch.
 ```
 logging_enabled: true
-rsyslog_default: false
 logging_outputs:
   - name: viaq-elasticsearch
     type: elasticsearch
     logs_collections:
-      - name: viaq
-    # 'state' is not a mandatory field. Defaults to `present`.
-      - name: viaq-k8s
+      - name: viaq-input
+        type: viaq
+    # 'state' is not a mandatory field. Defaults to 'present'.
+      - name: viaq-k8s-input
+        type: viaq-k8s
       - name: ovirt
+        type: ovirt-input
         state: absent
     server_host: logging-es
     server_port: 9200
@@ -149,7 +151,8 @@ logging_outputs:
   - name: ovirt-elasticsearch
     type: elasticsearch
     logs_collections:
-      - name: ovirt
+      - name: ovirt-input
+        type: ovirt
     server_host: logging-es-ovirt
     server_port: 9200
     index_prefix: project.ovirt-logs
@@ -180,8 +183,9 @@ Variables in vars.yaml
    -  **If `type: elasticsearch`**, send logs to one or more remote elasticsearch or Viaq installations.
       - `name`: Name of the elasticsearch element.
       - `type`: Type of the output element. Optional values: `elasticsearch`, `local`, `custom_files`.
-      - `logs_collections` : List of optional logs collections, dictionaries with `name` and `state` attributes, that were pre-configured.
-        - `name`: The name of the pre-configured logs to collect. **Note:** Currently only ['viaq', 'viaq-k8s', 'ovirt'] are supported for the elasticsearch output.
+      - `logs_collections` : List of optional logs collections, dictionaries with `name`, `type` and `state` attributes, that were pre-configured.
+        - `name`: Unique name of the input.
+          `type`: The type of the pre-configured logs to collect. **Note:** Currently ['viaq', 'viaq-k8s', 'ovirt'] are supported for the elasticsearch output.
           
           `state`: The state of the configuration files states if they should be `present` or `absent`. Default to `present`.
       - `server_host`: Hostname elasticsearch is running on.
--- a/roles/rsyslog/README.md
+++ b/roles/rsyslog/README.md
@@ -57,11 +57,14 @@ logging_outputs:
    type: <output_type0>
    logs_collections:
      - name: <input_role_nameA>
+       type: <input_role_typeA>
      - name: <input_role_nameB>
+       type: <input_role_typeB>
   -name: <output_role_name1>
    type: <output_type1>
    logs_collections:
      - name: <input_role_nameC>
+       type: <input_role_typeC>
 ```
 
 See the [variables section](#variables) for each variable.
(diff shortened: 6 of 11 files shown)


nhosoi:
Hi @sradco,
Could you please take a look at this PR?  In this PR, basically 2 changes are being made.
1) changing the default value of rsyslog_default from `true` to `false`.
2) adjusting the sub-variables for logs_collections to the ones for logging_outputs.

I might have surprised you by (2).  I'm revisiting the doc/readme now and have noticed the way how to specify the input type is different from the output type.  I'm afraid it may confuse the people who write logging system role playbook.  I think both had better use "type" instead of "name" to specify the type.  What do you think?

sradco:
@nhosoi @pcahyna 

> Hi @sradco,
> Could you please take a look at this PR? In this PR, basically 2 changes are being made.
> 
> 1. changing the default value of rsyslog_default from `true` to `false`.

OK. I'll submit a patch on the oVirt side for setting it to 'true'.
> 2. adjusting the sub-variables for logs_collections to the on
es for logging_outputs.
> 
> I might have surprised you by (2). I'm revisiting the doc/readme now and have noticed the way how to specify the input type is different from the output type. I'm afraid it may confuse the people who write logging system role playbook. I think both had better use "type" instead of "name" to specify the type. What do you think?

We currently the "name" in the "logs_collections" as the actual role name. We use it in order to loop over the input sub-roles.
If you want to add  "type" I'm not sure it will have the correct meaning. Its a bit confusing.
"Type" is usually unique and "name" is usually free text and not unique.

nhosoi:
> @nhosoi @pcahyna
> 
> > Hi @sradco,
> > Could you please take a look at this PR? In this PR, basically 2 changes are being made.
> > 
> > 1. changing the default value of rsyslog_default from `true` to `false`.
> 
> OK. I'll submit a patch on the oVirt side for setting it to 'true'.

Thanks!

> > 1. adjusting the sub-variables for logs_collections to the on
> >    es for logging_outputs.
> > 
> > I might have surprised you by (2). I'm revisiting the doc/readme now and have noticed the way how to specify the input type is different from the output type. I'm afraid it may confuse the people who write logging system role playbook. I think both had better use "type" instead of "name" to specify the type. What do you think?
> 
> We currently the "name" in the "logs_collections" as the actual role name. We use it in order to loop over the input sub-roles.
> If you want to add "type" I'm not sure it will have the correct meaning. Its a bit confusing.
> "Type" is usually unique and "name" is usually free text and not unique.
> 
> The "name" in "logging_outputs" is not unique and the "type" is unique.

Hmmm, then why the "name" in "logging_outputs" is used for the actual role name as in input sub-roles?

I think I understand the current implementation.  And I'm proposing to change it as I described.  The mismatch the usage of the parameter name ("type" vs. "name") between the output and input loop makes me worry a bit.  People may think why not "type" but "name" for logs_collections or vice versa...

roles/rsyslog/handlers/main.yaml
```diff
@@ -1,5 +1,5 @@
 ---
 - name: restart rsyslogd
-  service:
-    name: 'rsyslog'
-    state: 'restarted'
+  systemd:
```
pcahyna: why this change?

nhosoi: Well, at that time, I was just afraid that `service` could be deprecated some day. :p   Most likely, `service` would cover both non-systemd and systemd, wouldn't it?  If so, I'm going to revert this change.  Thanks!

pcahyna: https://docs.ansible.com/ansible/latest/modules/service_module.html#status
"This module is guaranteed to have backward compatible interface changes going forward. [stableinterface]
This module is maintained by the Ansible Core Team. [core]"
So I don't think there is more risk of deprecation of `service` than for any other core Ansible feature. (In principle, any feature can become deprecated, what makes `service` special?)
Also, please follow examples of other roles. I believe we are using`service` at many other places. If it becomes deprecated, it will be a problem that we will solve globally for all roles. logging is not special and I don't see a reason to make it different.
Also please read the best practices here: https://github.com/oasis-roles/meta_standards#ansible-best-practices . Relevant part: "Use meta modules when possible. Instead of using the upstart and systemd modules, use the service module when at all possible. Similarly for package management, use package instead of yum or dnf or similar. This will allow our playbooks to run on the widest selection of operating systems possible without having to modify any more tasks than is necessary."
Also please try to avoid mixing unrelated changes in one PR - there already two unrelated changes in this PR and this change is not related to any of them.

nhosoi: Thanks a lot for your comments, @pcahyna.

Indeed, it was wrong to include the unrelated changes.  I'm going to open separated pull requests instead of merging them into one.

pcahyna:
> > I might have surprised you by (2). I'm revisiting the doc/readme now and have noticed the way how to specify the input type is different from the output type. I'm afraid it may confuse the people who write logging system role playbook. I think both had better use "type" instead of "name" to specify the type. What do you think?
> 
> We currently the "name" in the "logs_collections" as the actual role name. We use it in order to loop over the input sub-roles.
> If you want to add "type" I'm not sure it will have the correct meaning. Its a bit confusing.
> "Type" is usually unique and "name" is usually free text and not unique.
> 
> The "name" in "logging_outputs" is not unique and the "type" is unique.

@sradco I am curious about this. Usually we use `name` at an attribute that uniquely identifies an item in the list. Like name of the network connection for the network role. If you say that "name" in "logging_outputs" is not unique and the "type" is unique, do you mean that there can be several items in the list with the same `name`, but each of the items must have a unique `type`, i.e there can't be several items with the same `type`?
@nhosoi do your changes mean that it will be possible to have several items in the `logs_collections` list with the same `type` and not possible to have several items with the same `name`?

sradco:
On Mon, Feb 10, 2020, 21:34 pcahyna <notifications@github.com> wrote:

> I might have surprised you by (2). I'm revisiting the doc/readme now and
> have noticed the way how to specify the input type is different from the
> output type. I'm afraid it may confuse the people who write logging system
> role playbook. I think both had better use "type" instead of "name" to
> specify the type. What do you think?
>
> We currently the "name" in the "logs_collections" as the actual role name.
> We use it in order to loop over the input sub-roles.
> If you want to add "type" I'm not sure it will have the correct meaning.
> Its a bit confusing.
> "Type" is usually unique and "name" is usually free text and not unique.
>
> The "name" in "logging_outputs" is not unique and the "type" is unique.
>
> @sradco <https://github.com/sradco> I am curious about this. Usually we
> use name at an attribute that uniquely identifies an item in the list.
> Like name of the network connection for the network role. If you say that
> "name" in "logging_outputs" is not unique and the "type" is unique, do you
> mean that there can be several items in the list with the same name, but
> each of the items must have a unique type, i.e there can't be several
> items with the same type?
>
logging_outputs is a list of the outputs.
It means we can have several elasticsearch setups and we will have same
type but different names. The names are free text just for clarity.

"name" under logs_collections was the name of the input sub-role like
oVirt/viaq and so on.

If we want to set things similer to the logging_outputs I think we need to
replace "name" with "type". "name" will be more semantic and  free text.

If this is what we need to do, then we need to replace the name and type in
the code so that the sub-roles will loop over the type.

Best
Shirly

nhosoi:
Thank you for your comments, @sradco.
> If we want to set things similer to the logging_outputs I think we need to
> replace "name" with "type". "name" will be more semantic and  free text.
>
> If this is what we need to do, then we need to replace the name and type in
> the code so that the sub-roles will loop over the type.

These are the places I'm replacing the name with the type in the loop.
https://github.com/linux-system-roles/logging/pull/83/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32L162-R163
https://github.com/linux-system-roles/logging/pull/83/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32L173-R174

And this is the doc in readme which describes the change.
https://github.com/linux-system-roles/logging/pull/83/files#diff-1c7bee374b17a63912b35d50adb51f95L89-R96

Do they fit into your descriptions?

nhosoi:
> @nhosoi do your changes mean that it will be possible to have several items in the `logs_collections` list with the same `type` and not possible to have several items with the same `name`?

Yes, for instance, we need to support multiple imfile inputs as follows.  (please note that imfile input is not yet pushed.)
```
        logging_outputs:
          - name: output0
            type: forwards
            [........]
            logs_collections:
              - name: input_from_dir0
                type: files
                rsyslog_input_log_path: /path/to/dir0/*.log
                rsyslog_input_log_tag: "tag0"
              - name: input_from_dir1
                type: files
                rsyslog_input_log_path: /path/to/dir1/*.log
                rsyslog_input_log_tag: "tag1"
```

nhosoi:
Note: this PR depends on https://github.com/linux-system-roles/logging/pull/84, which depends on https://github.com/linux-system-roles/logging/pull/85.

sradco:
> Thank you for your comments, @sradco.
> 
> > If we want to set things similer to the logging_outputs I think we need to
> > replace "name" with "type". "name" will be more semantic and  free text.
> > If this is what we need to do, then we need to replace the name and type in
> > the code so that the sub-roles will loop over the type.
> 
> These are the places I'm replacing the name with the type in the loop.
> https://github.com/linux-system-roles/logging/pull/83/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32L162-R163
> https://github.com/linux-system-roles/logging/pull/83/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32L173-R174
> 
> And this is the doc in readme which describes the change.
> https://github.com/linux-system-roles/logging/pull/83/files#diff-1c7bee374b17a63912b35d50adb51f95L89-R96
> 
> Do they fit into your descriptions?

Yes. They look good to me.

pcahyna:
Can you please edit the PR description to explain why the change of default is desirable? I think I understand it now, but it would be good to have it recorded.

pcahyna:
> The "name" in "logging_outputs" is not unique and the "type" is unique.
> 
> >  @sradco I am curious about this. Usually we
> >  use name at an attribute that uniquely identifies an item in the list.
> >  Like name of the network connection for the network role. If you say that
> >  "name" in "logging_outputs" is not unique and the "type" is unique, do you
> >  mean that there can be several items in the list with the same name, but
> >  each of the items must have a unique type, i.e there can't be several
> >  items with the same type?
> > 
> 
> logging_outputs is a list of the outputs.
> It means we can have several elasticsearch setups and we will have same
> type but different names. The names are free text just for clarity.

@sradco I don't understand - if you can have several outputs with same `type`, then `type` is not unique, `name` is.

.travis.yml
```diff
@@ -7,8 +7,7 @@ env:
   matrix:
     - MOLECULE_DISTRO="centos:7"
     - MOLECULE_DISTRO="centos:8"
-    - MOLECULE_DISTRO="fedora:28"
-    - MOLECULE_DISTRO="fedora:29"
+    - MOLECULE_DISTRO="fedora:31"
```
pcahyna: Do you need this to test the changes? if not, I would prefer to leave it for later PR, as it is not related to the subject of this PR.

molecule/default/playbook.yml
```diff
@@ -1,5 +1,7 @@
 ---
 - name: Converge
   hosts: all
+  vars:
+    rsyslog_default: True
```
pcahyna: lowercase

roles/rsyslog/defaults/main.yaml
```diff
@@ -362,7 +362,7 @@ rsyslog_conf_global_options:
     options: |-
       global(
         defaultNetstreamDriver="{{ rsyslog_default_netstream_driver }}"
-      {% if not rsyslog_default|default(true)|bool %}
+      {% if not rsyslog_default|default(false)|bool %}
```
pcahyna: perhaps the use of the default filter could be eliminated completely, given that we have the variable in defaults. This way you have to maintain the default value at two places.

roles/rsyslog/tasks/main.yaml
```diff
@@ -192,10 +193,11 @@
         - skip_ansible_lint
 
     - name: Enable rsyslog service
-      service:
-        name: 'rsyslog'
+      systemd:
+        name: rsyslog.service
+        daemon_reload: yes
```
pcahyna: `service:` was better, wasn't it?

tests/tests_enabled.yml
```diff
@@ -2,11 +2,11 @@
 - name: Ensure that the role runs with rsyslog_enabled=True
   hosts: all
   become: true
+  vars:
+    rsyslog_default: True
```
pcahyna: lowercase

nhosoi:
> Can you please edit the PR description to explain why the change of default is desirable? I think I understand it now, but it would be good to have it recorded.

I've updated the PR description.

nhosoi:
Thanks for reviewing the PR, @pcahyna!  Merging...

nhosoi:
@sradco, if you could test the master branch with your updated var/inventory file, I'd greatly appreciate it.

tests/tests_enabled.yml
pcahyna: the change is still there.

nhosoi: (;_;) Thanks for finding it out, @pcahyna.  It seems I might have made some mistakes in the conflict resolution...  I'm planning to open another bit PR for changing the internal variables to start with '_'.  Is it ok to add the fix to the PR?

pcahyna: @nhosoi I suggest that you just force-push to master with this part corrected, I think you will not create problems for anyone by force pushing so soon after you merged the PR @richm what do you think?

roles/rsyslog/handlers/main.yaml
pcahyna: > Well, at that time, I was just afraid that `service` could be deprecated some day.

FYI, `service` will survive even the upcoming project restructuring: https://github.com/ansible-community/collection_migration/blob/6a42daee3d1a7443f64ebccff0994dcc29ecd439/scenarios/nwo/ansible.yml#L286

PR #84, opened 2020-02-05
RHELPLAN-32351 - Add an ability to distribute messages to different o…

nhosoi:
…utputs based on severity and facility

- Adding rsyslog_filter_actions, which is an array of dictionary to specify
  files to store logs for files outputs or forward target host to redirect
  the logs to the remote rsyslog servers for forwards output.
- Adding test cases:
  tests_default_files.yml - outputs to default local files
  tests_files.yml - outputs to configured local files
  tests_forwards.yml - sends to remote hosts

### DIFF
--- a/roles/rsyslog/README.md
+++ b/roles/rsyslog/README.md
@@ -52,6 +52,7 @@ Currently, the role supports 3 types of logs collections ([input_roles](https://
 To make an effect with the following setting, vars.yaml has to have `rsyslog_enabled: true`.  Unless `rsyslog_enabled` is set to true, LSR/Logging does not deploy rsyslog config files.
 ```
 rsyslog_enabled: true
+rsyslog_default: false
 logging_outputs:
   -name: <output_role_name0>
    type: <output_type0>
@@ -69,21 +70,22 @@ logging_outputs:
 
 See the [variables section](#variables) for each variable.
 
-**vars.yaml examples:**
+vars.yaml examples
+------------------
 
-0. Deploying default /etc/rsyslog.conf.
+**0. Deploying default /etc/rsyslog.conf.**
 ```
 rsyslog_enabled: true
 ```
 
-0. Overriding existing /etc/rsyslog.conf and files in /etc/rsyslog.d with default rsyslog.conf.  Pre-existing config files are copied to the directory specified by `rsyslog_backup_dir`.
+**1. Overriding existing /etc/rsyslog.conf and files in /etc/rsyslog.d with default rsyslog.conf.  Pre-existing config files are copied to the directory specified by `rsyslog_backup_dir`.**
 ```
 rsyslog_enabled: true
 rsyslog_purge_original_conf: true
 rsyslog_backup_dir: /tmp/rsyslog_backup
 ```
 
-1. Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the local system and outputs into the local files.  Pre-existing config files are copied to the directory specified by `rsyslog_backup_dir`.
+**2. Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the local system and outputs into the local files.  Pre-existing config files are copied to the directory specified by `rsyslog_backup_dir`.**
 ```
 rsyslog_enabled: true
 rsyslog_purge_original_conf: true
@@ -96,7 +98,7 @@ logging_outputs:
         type: basics
 ```
 
-2. Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the local system and remote rsyslog and outputs into the local files.
+**3. Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the local system and remote rsyslog and outputs into the local files.**
 ```
 rsyslog_enabled: true
 rsyslog_capabilities: [ 'network', 'remote-files' ]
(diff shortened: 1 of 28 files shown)


nhosoi:
/hold
Sorry, found a bug...

roles/rsyslog/README.md
```diff
@@ -51,7 +51,8 @@ Currently, the role supports 3 types of logs collections ([input_roles](https://
 
 To make an effect with the following setting, vars.yaml has to have `rsyslog_enabled: true`.  Unless `rsyslog_enabled` is set to true, LSR/Logging does not deploy rsyslog config files.
 ```
-rsyslog_enabled: true
+rsyslog_enabled: True
```
pcahyna: just a nit, but why are you changing true to True? Ansible documentation almost universally uses yes/no or true/false (lowercase).

nhosoi: Oops.  Sorry...  Do you recommend to consolidate all the boolean setting to yes/no?  (I think I read it in some doc...)

pcahyna: Maybe, but it wouldn't belong to this PR anyway (such stylistic changes have no relation with the subject of the PR, which is my other objection to this particular change).

nhosoi: @pcahyna, I've opened another PR [1] for these issues.  
- Replacing 'yum' with 'package'.
- Removing quotes from single values in the tasks.
- Lowering the letter cases of True/False.

Could you please take a look?  Thanks!
[1] https://github.com/linux-system-roles/logging/pull/85

nhosoi: > just a nit, but why are you changing true to True? Ansible documentation almost universally uses yes/no or true/false (lowercase).

@pcahyna, undid this change.  (sorry for my slow understanding...)  Please let me know if there are any other changes I should eliminate.  Thanks!

roles/rsyslog/defaults/main.yaml
```diff
@@ -301,6 +292,22 @@ rsyslog_weight_map:
   'input': '90'
   'inputs': '90'
 
+rsyslog_severity_to_level:
```
pcahyna: You don't need to include dictionary keys in quotes/apostrophes.

Concerning the values. Are they really supposed to be strings (enclosed in apostrophes) rather than integers? Wouldn't integers be more natural? (Similar question applies to the already existing `rsyslog_weight_map` above.)

Also, is this variable supposed to be user-settable? It does not seem so, it seems rather internal. If so, it should not be in defaults but in vars and prefixed with underscores (as an internal variable).

nhosoi: Ahhh, thank you for finding it out, @pcahyna!  I introduced it at one time, but it turned out not needed.  I'll eliminate rsyslog_severity_to_level.  (and when I find it's necessary again, I'll  use integer... and starting with '_')  Thanks!

roles/rsyslog/roles/input_roles/basics/defaults/main.yaml
```diff
@@ -14,6 +14,8 @@ rsyslog_unprivileged: False
 rsyslog_capabilities: []
 
 rsyslog_use_imuxsock: False
+rsyslog_input_log_path: "/var/log/containers/*.log"
+rsyslog_input_log_tag: "container"
```
pcahyna: Are those internal variables, or variables expected to be configured by users?

nhosoi: They need to be configured by users.

roles/rsyslog/defaults/main.yaml
nhosoi: @pcahyna, BTW, this is also internal: rsyslog_weight_map.  And I believe there are more of them.  Do you think I should change the internal name as you suggested?  Is it ok to update https://github.com/linux-system-roles/logging/pull/85 to take care of them?  Or do you prefer to have it in another new PR?

roles/rsyslog/roles/input_roles/basics/defaults/main.yaml
pcahyna: Shouldn't it be added to the documentation then?

roles/rsyslog/defaults/main.yaml
pcahyna: You can do this cleanup in #85 or another PR. Please don't try to fix such stylistic issues here to keep the PR focused, just avoid introducing new ones.

roles/rsyslog/roles/input_roles/basics/defaults/main.yaml
nhosoi: > Shouldn't it be added to the documentation then?

Done.
https://github.com/linux-system-roles/logging/pull/84/files#diff-1c7bee374b17a63912b35d50adb51f95R298-R301

roles/rsyslog/defaults/main.yaml
pcahyna: To answer the original question, another new PR would be preferred, so you can merge #85 faster and avoid conflicts.

nhosoi: > To answer the original question, another new PR would be preferred, so you can merge #85 faster and avoid conflicts.

Eliminated `rsyslog_severity_to_level`.  Thanks!

nhosoi:
@sradco, 
Once this pr/84 is merged, could you please test your updated var/inventory file on ovirt?  Thanks!!

nhosoi:
@pcahyna, @richm. @sradco, is it ok to merge this PR?  Thanks...

nhosoi:
Thank you soooooo much for reviewing the PR and sharing the fixing and improving ideas, @pcahyna.

PR #85, opened 2020-02-11
Following Ansible Best Practices

nhosoi:
Replacing 'yum' with 'package'.
Removing quotes from single values in the tasks.
Lowering the letter cases of True/False.
White spaces in jinja2: {{ variable_name_here }} and {{ variable_name | default(...) }}

### DIFF
--- a/roles/rsyslog/defaults/main.yaml
+++ b/roles/rsyslog/defaults/main.yaml
@@ -5,12 +5,12 @@
 # rsyslog_enabled
 #
 # Enable or disable ``rsyslog`` management.
-rsyslog_enabled: True
+rsyslog_enabled: true
 
 # rsyslog_default
 #
 # Diploy default ``rsyslog`` configuration file.
-rsyslog_default: True
+rsyslog_default: true
 
 # rsyslog_system_log_dir
 #
@@ -38,21 +38,19 @@ rsyslog_capabilities: []
 # Enable or disable unprivileged ``rsyslogd`` operation. Warning, enabling this
 # option requires additional configuration outside of the ``rsyslog``
 # role.
-rsyslog_unprivileged: '{{ "True"
-                           if (ansible_distribution in [ "CentOS", "RedHat", "Fedora" ])
-                           else "False" }}'
+rsyslog_unprivileged: '{{ ansible_distribution in [ "CentOS", "RedHat", "Fedora" ] }}'
 
 # rsyslog_message_reduction
 #
 # Enable or disable message reduction. This is disabled by default so that log
 # parsers like :program:`fail2ban` can work correctly.
-rsyslog_message_reduction: False
+rsyslog_message_reduction: false
 
 # rsyslog_purge_original_conf
 #
 # The pre-existing config files are archived in the tempdir and purged if the
-# rsyslog_purge_original_conf is set to `True`.  Default to 'False`.
-rsyslog_purge_original_conf: False
+# rsyslog_purge_original_conf is set to `true`.  Default to 'false`.
+rsyslog_purge_original_conf: false
 
 # rpm packages
 # adding rsyslog_logging_packages
@@ -82,8 +80,8 @@ rsyslog_custom_config_files: []
 # rsyslog_basics_use_traditional_timestamp_format
 #
 # Traditional timestamp format looks like 'Mar 27 14:16:47'
-# By setting False, it'd change 2020-03-27T14:16:47.139796+00:00)
-rsyslog_basics_use_traditional_timestamp_format: True
+# By setting false, it'd change 2020-03-27T14:16:47.139796+00:00)
+rsyslog_basics_use_traditional_timestamp_format: true
 
 
 # User, group, home, file ownership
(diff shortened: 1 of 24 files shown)


roles/rsyslog/defaults/main.yaml
```diff
@@ -38,21 +38,21 @@ rsyslog_capabilities: []
 # Enable or disable unprivileged ``rsyslogd`` operation. Warning, enabling this
 # option requires additional configuration outside of the ``rsyslog``
 # role.
-rsyslog_unprivileged: '{{ "True"
+rsyslog_unprivileged: '{{ "true"
                            if (ansible_distribution in [ "CentOS", "RedHat", "Fedora" ])
-                           else "False" }}'
+                           else "false" }}'
```
pcahyna: can't you simplify to just `'{{ ansible_distribution in [ "CentOS", "RedHat", "Fedora" ] }}'`, given that this should be a boolean expression already?
Also, explicit tests for distribution should be discouraged, it would be better to use a distribution-specific var file, but let's keep it for later, this is a bigger change.

pcahyna:
Looks good in general, except for the comment above. There are certainly more style issues in the role and others, but we don't need to fix them all in one PR, I would prefer to have this one merged and open separate PRs for more cleanups.

nhosoi:
> Looks good in general, except for the comment above. There are certainly more style issues in the role and others, but we don't need to fix them all in one PR, I would prefer to have this one merged and open separate PRs for more cleanups.

Thanks for your reviewing, @pcahyna.  I updated ansible_distribution following your comment.

roles/rsyslog/defaults/main.yaml
pcahyna: oh and I think this has another problem: you probably want boolean values true/false and not strings "true"/"false", so the values should not be quoted.

roles/rsyslog/tasks/main.yaml
```diff
@@ -1,16 +1,16 @@
...
-    - not use_rsyslog_image|default(False)|bool
+    - rsyslog_enabled | bool
+    - not use_rsyslog_image | default(false) | bool
   tags:
     - skip_ansible_lint
 
 - name: Get rsyslog version
-  yum:
+  package:
     list: rsyslog
```
pcahyna: does the package module actually support this parameter? https://docs.ansible.com/ansible/latest/modules/package_module.html.

I suggest to use https://docs.ansible.com/ansible/latest/modules/package_facts_module.html , we have been using it elsewhere like here: https://github.com/linux-system-roles/timesync/commit/e145b8cda93dbbd64aba2d049ed562b29b974ae5

roles/rsyslog/defaults/main.yaml
nhosoi: Opening another PR for "true"/"false" --> true/false...

roles/rsyslog/tasks/main.yaml
nhosoi: Thanks!  I'm fixing it in a new PR...

roles/rsyslog/defaults/main.yaml
pcahyna: um, why? The code is not there anymore.

nhosoi: @pcahyna 
> um, why? The code is not there anymore.

Because I fixed it as you suggested and merged it with --force...

pcahyna: @nhosoi I don't understand. Here: https://github.com/linux-system-roles/logging/blob/f63b8439e2faa4586d3855c95446def98e39f586/roles/rsyslog/defaults/main.yaml#L42 
there is no trace of "true" ad "false" anymore because you fixed it to not need it at all. (My comment was about the original code and did not apply to the fixed version.) Therefore, you don't need to open a new PR for "true"/"false" --> true/false, unless they appear in some other places in the code.

nhosoi: Ahhh, I see...  I somehow misunderstood you were talking about "true" and "false" in general, as I thought I made the mistake many places, but it looks I didn't [1].  *phew*  Sorry for my confusion and noise.
[1]
```
$ find logging -type f | xargs egrep "\"true\"|\"false\""
logging/roles/rsyslog/roles/README.md:  - If `rsyslog_default` equals to "true", It is required to set for all logs you don't want to be processed by the default rules:
logging/roles/rsyslog/roles/README.md:    set $.send_targets_only = "true";
logging/roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml:              set $.send_targets_only = "true";
logging/roles/rsyslog/roles/input_roles/ovirt/defaults/main.yaml:              set $.send_targets_only = "true";
```

roles/rsyslog/tasks/main.yaml
nhosoi: @pcahyna, I created a patch following your advice and submitted a pr: 
https://github.com/linux-system-roles/logging/pull/87 "Fixing package madule related issues."
Could you please take a look?  Thanks, in advance!!

PR #86, opened 2020-02-13
.travis.yml - Upgrading fedora versions to test from {28,29} to {30,31}.

nhosoi:


pcahyna:
I think this would be better done in the common template - @i386x do we have anything analogous?

pcahyna:
our Molecule and Travis configurations: https://github.com/linux-system-roles/template/blob/master/.travis.yml https://github.com/linux-system-roles/template/blob/master/molecule/default/molecule.yml
@i386x @tyll @vcrhonek wouldn't the approach in https://github.com/linux-system-roles/logging/blob/00595f00c56be3b4f80a89da1bdd6d474c93f853/molecule/default/molecule.yml#L12 be useful for other roles as well? By using a parameter we could possibly avoid some of the repeated boilerplate that our attempts at using Molecule suffered from.

tyll:
> our Molecule and Travis configurations: https://github.com/linux-system-roles/template/blob/master/.travis.yml https://github.com/linux-system-roles/template/blob/master/molecule/default/molecule.yml
> @i386x @tyll @vcrhonek wouldn't the approach in
> 
> https://github.com/linux-system-roles/logging/blob/00595f00c56be3b4f80a89da1bdd6d474c93f853/molecule/default/molecule.yml#L12
> 
> be useful for other roles as well? By using a parameter we could possibly avoid some of the repeated boilerplate that our attempts at using Molecule suffered from.

Currently we only have centos 6 and 7 and they differ regarding the volumes. If the volume does not hurt for CentOS 6, it would be possible.

pcahyna:
@nhosoi I think you can merge the PR for now, but we will look at the possibility of doing the whole Travis/Molecule CI setup differently and unifying with our template.

i386x:
As @tyll said, if the `volumes` does not break the `centos-6` case, this approach can be used also in our common template and thus in all roles. I see one disadvantage here and that is increasing a number of Molecule runs, in this case two times.

However, there can be a workaround which can work in next Molecule stable release: generating platforms list automatically from `galaxy_info`, putting them to `base.yml` and passing `base.yml` via `-c` option to Molecule (this requires no `platforms` in `molecule.yml`, see [[1](https://github.com/ansible-community/molecule/issues/1423), [2](https://github.com/ansible-community/molecule/issues/1328)]).

Another solution can be having a customized copy of Docker's `create.yml` playbook, but this will lead to code divergence, so I propose to wait to the next stable Molecule release.

pcahyna:
@robled @seandst @greg-hellings any advice here please, is it advisable to use an environment variable in Molecule configuration and put the list of images into Travis, instead of defining it inside Molecule, or not?

greg-hellings:
We have been using different scenarios for different systems, rather than trying to unify that. We do all of our unification at the level of pushing provisioning and teardown code out into our [roles](https://github.com/oasis-roles/molecule_docker_ci) and including shared playbooks under `molecule/shared/{playbook,prepare,teardown}.yml`.

Each scenario we still give its own managed molecule.yml file. But we also don't test a very large matrix of systems. Typically just CentOS 7 on Travis and RHEL 7 on Jenkins.

Personally I see no issues with pulling a variable in from the environment to expand a test matrix like this. The only downside is that it makes localized testing more tedious, but the advantage is that CI is more streamlined.

pcahyna:
@greg-hellings thanks for your reply - could you please point to a good example of an OASIS role using Molecule tests (a working example, not just meta_skeleton)?

PR #87, opened 2020-02-14
Fixing package module related issues.

nhosoi:
- Gathering package facts with package_facts first.
- Using ansible_facts.packages['rsyslog'][0].version for the rsyslog version.
- Renaming a variable rsyslog_packages with rsyslog_extra_packages to make it clearer. With this variable, customers are allowed to install extra rsyslog packages required for their custom configurations if any.
- Renaming a variable backupdir with rsyslog_backup_dir. Pre-existing config files are copied to the directory.  Default to /tmp/rsyslog.d-DATETIME.
- Renaming a pair of variables rsyslog_role_packages and rsyslog_role_rules with __rsyslog_packages and __rsyslog_rules, respectively.  Note: the vars start with `__` since they are internal.  By the same token, the underneath sub rule names and their items are also changed to start with __, e.g.,
      rsyslog_common_rules --> __rsyslog_common_rules
      rsyslog_conf_global_options --> __rsyslog_conf_global_options
- The internal variables are moved from defaults/main.yaml to vars/main.yaml.
- Fixing the relative path from role_path in roles/rsyslog/roles/*_roles/*/tasks/main.yaml.
- Removing set_test from CI tests.
- To call include_role, instead of setting variables to pass with set_fact, set them with "vars" to make them local in the scope.
- If rsyslog is not found, e.g., in case of the check mode, set 0.0.0 to __rsyslog_version.

Fixing travis-ci
- forcing to install molecule < 3.
- fixing yamllint and ansible-lint errors excepot 'state: latest' for the package module, lines longer than 160 chars, and commands should not change things if nothing needs doing.
- replacing var use_rsyslog_image with rsylog_in_image and set it to false by default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ env:
     - MOLECULE_DISTRO="fedora:30"
     - MOLECULE_DISTRO="fedora:31"
 install:
-  - pip install molecule docker
+  - pip install 'molecule<3' docker
   # workaround until https://github.com/ansible/molecule/issues/1727 is fixed
   - pip uninstall -y testinfra
   - pip install testinfra
--- a/roles/rsyslog/README.md
+++ b/roles/rsyslog/README.md
@@ -264,6 +264,7 @@ Common sub-variables
 - `rsyslog_config_dir`: Directory to store configuration files.  Default to '/etc/rsyslog.d'.
 - `rsyslog_custom_config_files`: List of custom configuration files are deployed to /etc/rsyslog.d.  The format is an array which element is the full path to each custom configuration file.  Default to none.
 - `rsyslog_group`: Owner group of rsyslogd.  Default to 'syslog' if `rsyslog_unprivileged` is true.  Otherwise, 'root'.
+- `rsyslog_in_image`: Specifies if the target host is a container and use rsyslog in the image. Default to false.
 - `rsyslog_purge_original_conf`: By default, the Rsyslog configuration files are applied on top of pre-existing configuration files. To purge local files prior to setting new ones, set `rsyslog_purge_original_conf` variable to 'true', it will move all Rsyslog configuration files to a backup directory before deploying the new configuration files. Defaults to 'false'.
 - `rsyslog_system_log_dir`: System log directory.  Default to '/var/log'.
 - `rsyslog_unprivileged`: If set to true, you could specify non-root `rsyslog_user` and `rsyslog_group`.  If ansible_distribution is one of "CentOS", "RedHat", and "Fedora", default to true.  Otherwise, 'false'.
@@ -340,9 +341,9 @@ Describing how the configuration files are defined to be deployed using the viaq
 
 Viaq configuration files are defined in {{ __rsyslog_viaq_rules }} in /roles/input_roles/viaq/defaults/main.yaml.  The set is made from the generic modules{__rsyslog_conf_global_options, __rsyslog_conf_local_modules, __rsyslog_conf_common_defaults} and viaq specific configurations.
 
-To make a new configuration file installed in addition to the current {{ rsyslog_viaq_rules }}, create an rsyslog config item based on the following skelton and add the title {{ rsyslog_conf_yourname }} to {{ rsyslog_viaq_rules }}.
+If you are a role developer and planning to add a new default configuration to the role depo, create an rsyslog config item based on the following skelton and add the title {{ __rsyslog_conf_yourname }} to {{ __rsyslog_viaq_rules }}.
 ```
-rsyslog_conf_yourname:
+__rsyslog_conf_yourname:
 
   - name: some_name
     type: choose one of 'global' 'module' 'modules' 'template' 'templates' 'output' 'service' 'rule' 'rules' 'ruleset' 'input'
--- a/roles/rsyslog/README.md
+++ b/roles/rsyslog/README.md
@@ -338,7 +338,7 @@ WARNING: Pre-existing rsyslog.conf and configuration files in /etc/rsyslog.d are
 
 Describing how the configuration files are defined to be deployed using the viaq case.
 
-Viaq configuration files are defined in {{ rsyslog_viaq_rules }} in /roles/input_rols/viaq/defaults/main.yaml.  The set is made from the generic modules{rsyslog_conf_global_options, rsyslog_conf_local_modules, rsyslog_conf_network_modules, rsyslog_conf_common} and viaq specific configurations.
+Viaq configuration files are defined in {{ __rsyslog_viaq_rules }} in /roles/input_roles/viaq/defaults/main.yaml.  The set is made from the generic modules{__rsyslog_conf_global_options, __rsyslog_conf_local_modules, __rsyslog_conf_common_defaults} and viaq specific configurations.
 
 To make a new configuration file installed in addition to the current {{ rsyslog_viaq_rules }}, create an rsyslog config item based on the following skelton and add the title {{ rsyslog_conf_yourname }} to {{ rsyslog_viaq_rules }}.
 ```
--- a/roles/rsyslog/defaults/main.yaml
+++ b/roles/rsyslog/defaults/main.yaml
@@ -7,6 +7,11 @@
 # Enable or disable ``rsyslog`` management.
 rsyslog_enabled: true
 
+# rsyslog_in_image
+#
+# true if the target host is a container and use rsyslog in the image.
+rsyslog_in_image: false
+
 # rsyslog_default
 #
 # Diploy default ``rsyslog`` configuration file /etc/rsyslog.conf if it is set to true.
@@ -54,23 +59,15 @@ rsyslog_message_reduction: false
 rsyslog_purge_original_conf: false
 
 # rpm packages
-# adding rsyslog_logging_packages
+# adding rsyslog packages
 # ------------------
 
-# rsyslog_base_packages
+# rsyslog_extra_packages
 #
-# List of default rpm packages to install.
-rsyslog_base_packages: ['rsyslog']
-
-# rsyslog_tls_packages
-#
-# List of rpm packages required for TLS support.
-rsyslog_tls_packages: ['rsyslog-gnutls', 'ca-certificates']
-
-# rsyslog_packages
-#
-# List of additional rpm packages to install.
-rsyslog_packages: []
+# List of extra rpm packages to install.
+# If custom configurations are to be installed, which requires additional rsyslog
+# packages, this variable can be used to install them.
+rsyslog_extra_packages: []
 
 # rsyslog_custom_config_files
 #
@@ -271,8 +268,6 @@ rsyslog_forwards_actions: []
 #
 # If you change the default weight map values, you will most likely need to
 # remove all files from `{{ rsyslog_config_dir }}` to reset the configuration.
-#
-# See :ref:`rsyslog_rules` for more details.
 rsyslog_weight_map:
   'global': '05'
   'globals': '05'
@@ -322,29 +317,28 @@ rsyslog_host_rules: []
 # details. This list should be used for configuration by other Ansible roles.
 rsyslog_dependent_rules: []
 
-# rsyslog_.*_rules
+# __rsyslog_.*_rules
 #
 # List of YAML dictionaries, each dictionary should contain ``rsyslogd``
-# configuration in a special format. See :ref:`rsyslog_rules` for more
-# details. This lis specifies default ``rsyslogd`` configuration enabled in the
-# role.
-rsyslog_common_rules:
+# configuration in a special format. This lis specifies default ``rsyslogd``
+# configuration enabled in the role.
+__rsyslog_common_rules:
 
-  - '{{ rsyslog_conf_global_options }}'
-  - '{{ rsyslog_conf_local_modules }}'
-  - '{{ rsyslog_conf_common_defaults }}'
-  - '{{ rsyslog_conf_send_targets_only }}'
+  - '{{ __rsyslog_conf_global_options }}'
+  - '{{ __rsyslog_conf_local_modules }}'
+  - '{{ __rsyslog_conf_common_defaults }}'
+  - '{{ __rsyslog_conf_send_targets_only }}'
 
 
 # Default configuration options
 # -----------------------------
 
-# rsyslog_conf_global_options
+# __rsyslog_conf_global_options
 #
 # Some of the global ``rsyslogd`` configuration options. See
 # http://www.rsyslog.com/doc/v8-stable/rainerscript/global.html for more
 # details.
-rsyslog_conf_global_options:
+__rsyslog_conf_global_options:
 
   - filename: '00-global.conf'
     comment: 'Global options'
@@ -366,11 +360,11 @@ rsyslog_conf_global_options:
       {% endif %}
       )
 
-# rsyslog_conf_local_modules
+# __rsyslog_conf_local_modules
 #
 # List of ``rsyslogd`` modules that enable logs from the local system to be
 # received and parsed by the ``rsyslogd`` daemon.
-rsyslog_conf_local_modules:
+__rsyslog_conf_local_modules:
 
   - name: 'local-modules'
     type: 'modules'
@@ -383,11 +377,11 @@ rsyslog_conf_local_modules:
                   if ("mark" in rsyslog_capabilities)
                   else "absent" }}'
 
-# rsyslog_conf_common_defaults
+# __rsyslog_conf_common_defaults
 #
 # List of common ``rsyslogd`` configuration, like new file owner/group and
 # permissions, work directory, configuration of message reduction.
-rsyslog_conf_common_defaults:
+__rsyslog_conf_common_defaults:
 
   - name: 'common-defaults'
     type: 'global'
@@ -409,7 +403,7 @@ rsyslog_conf_common_defaults:
         options: |-
           $RepeatedMsgReduction {{ "on" if rsyslog_message_reduction | bool else "off" }}
 
-rsyslog_conf_send_targets_only:
+__rsyslog_conf_send_targets_only:
 
   - name: 'send-targets-only'
     type: 'output'
--- a/roles/rsyslog/defaults/main.yaml
+++ b/roles/rsyslog/defaults/main.yaml
@@ -320,7 +320,7 @@ rsyslog_dependent_rules: []
 # __rsyslog_.*_rules
 #
 # List of YAML dictionaries, each dictionary should contain ``rsyslogd``
-# configuration in a special format. This lis specifies default ``rsyslogd``
+# configuration in a special format. This list specifies default ``rsyslogd``
 # configuration enabled in the role.
 __rsyslog_common_rules:
 
--- a/roles/rsyslog/roles/README.md
+++ b/roles/rsyslog/roles/README.md
@@ -21,7 +21,7 @@ The `defaults` directory includes:
 
 The `tasks` directory includes 2 tasks file:
   - `main.yaml` - tasks for deploying the config files
-    This file is sets `rsyslog_role_packages` and `rsyslog_role_rules` and includes the task that deploys the files.
+    This file is sets `__rsyslog_packages` and `__rsyslog_rules` and includes the task that deploys the files.
   - `cleanup.yml` - tasks that cleanup the files deployed for this project.
 
 Examples can be found in the existing projects.
--- a/roles/rsyslog/roles/README.md
+++ b/roles/rsyslog/roles/README.md
@@ -10,7 +10,7 @@ Each project adds a sub-role to ./logging/roles/rsyslog/roles/input_roles/.
 
 The sub-role usually includes `tasks` and `defaults` directories.
 The `defaults` directory includes:
-  - List of required packages that are **not** the base rsyslog_base_packages: ['rsyslog', 'libselinux-python']
+  - List of required packages that are **not** the base rsyslog_base_packages: ['rsyslog']
   - List of modules to load  like `imfile`, `imtcp`, etc.
   - Defines the formatting and the rulebases for parsing the logs.
   - It is required to set for all logs the project identfier for pipelining:
@@ -19,9 +19,9 @@ The `defaults` directory includes:
   - If `rsyslog_default` equals to "true", It is required to set for all logs you don't want to be processed by the default rules:
     set $.send_targets_only = "true";
 
-The `tasks` directory includes 2 tasks file:
+The `tasks` directory includes 2 task files:
   - `main.yaml` - tasks for deploying the config files
-    This file is sets `__rsyslog_packages` and `__rsyslog_rules` and includes the task that deploys the files.
+    This file sets `__rsyslog_packages` and `__rsyslog_rules` and includes the task that deploys the files.
   - `cleanup.yml` - tasks that cleanup the files deployed for this project.
 
 Examples can be found in the existing projects.
--- a/roles/rsyslog/roles/input_roles/basics/defaults/main.yaml
+++ b/roles/rsyslog/roles/input_roles/basics/defaults/main.yaml
@@ -12,7 +12,6 @@ rsyslog_unprivileged: false
 #       Note: unles rsyslog_send_over_tls_only is set to true, insecure network connection
 #       is also allowed.
 rsyslog_capabilities: []
-rsyslog_basics_packages: ['rsyslog-gnutls']
 
 rsyslog_use_imuxsock: false
 
@@ -22,25 +21,25 @@ rsyslog_use_imuxsock: false
 rsyslog_input_log_path: "/var/log/containers/*.log"
 rsyslog_input_log_tag: "container"
 
-# rsyslog_basics_rules
+# __rsyslog_basics_rules
 #
 # List of YAML dictionaries, each dictionary should contain ``rsyslogd``
 # configuration in a special format. See :ref:`rsyslog_rules` for more
 # details. This list specifies basics ``rsyslogd`` example configuration.
-rsyslog_basics_rules:
-  - '{{ rsyslog_conf_local_basics_modules }}'
-  - '{{ rsyslog_conf_default_rulesets }}'
-  - '{{ rsyslog_conf_system_input }}'
-  - '{{ rsyslog_conf_network_input }}'
+__rsyslog_basics_rules:
+  - '{{ __rsyslog_conf_local_basics_modules }}'
+  - '{{ __rsyslog_conf_default_rulesets }}'
+  - '{{ __rsyslog_conf_system_input }}'
+  - '{{ __rsyslog_conf_network_input }}'
 
 # Debops example configuration options
 # --------------------------------------------
 
-# rsyslog_conf_local_modules
+# __rsyslog_conf_local_modules
 #
 # List of ``rsyslogd`` modules that enable logs from the local system to be
 # received and parsed by the ``rsyslogd`` daemon.
-rsyslog_conf_local_basics_modules:
+__rsyslog_conf_local_basics_modules:
 
   - name: 'local-basics-modules'
     type: 'modules'
@@ -92,7 +91,7 @@ rsyslog_conf_local_basics_modules:
                   if ("network" in rsyslog_capabilities) and ("tls" in rsyslog_capabilities)
                   else "absent" }}'
 
-# rsyslog_conf_default_rulesets
+# __rsyslog_conf_default_rulesets
 #
 # The ``rsyslogd`` configuration can contain multiple rulesets, each one
 # connected to an input channel (see
@@ -104,7 +103,7 @@ rsyslog_conf_local_basics_modules:
 # network from remote hosts. The local log rules are stored in
 # `{{ rsyslog_config_dir }}/*.system` configuration files, and remote log
 # rules are stored in `{{ rsyslog_config_dir }}/*.remote` configuration files.
-rsyslog_conf_default_rulesets:
+__rsyslog_conf_default_rulesets:
 
   - name: 'default-rulesets'
     type: 'rules'
@@ -123,9 +122,9 @@ rsyslog_conf_default_rulesets:
                   if ("network" in rsyslog_capabilities)
                   else "absent" }}'
 
-# rsyslog_conf_system_input
+# __rsyslog_conf_system_input
 #
-rsyslog_conf_system_input:
+__rsyslog_conf_system_input:
 
   - name: 'system-input'
     type: 'input'
@@ -138,11 +137,11 @@ rsyslog_conf_system_input:
                   if rsyslog_use_imuxsock | d(false) | bool
                   else "absent" }}'
 
-# rsyslog_conf_network_input
+# __rsyslog_conf_network_input
 #
 # Configuration of UDP, TCP and TCP over TLS inputs to receive logs from remote
 # hosts, enabled by the ``network`` capability.
-rsyslog_conf_network_input:
+__rsyslog_conf_network_input:
 
   - name: 'network-input'
     type: 'input'
--- a/roles/rsyslog/roles/input_roles/basics/tasks/main.yaml
+++ b/roles/rsyslog/roles/input_roles/basics/tasks/main.yaml
@@ -1,10 +1,9 @@
 ---
-- name: Set basics facts
-  set_fact:
-    rsyslog_role_packages: "{{rsyslog_basics_packages | flatten | d([])}}"
-    rsyslog_role_rules: "{{ rsyslog_basics_rules }}"
-
+# Deploy configuration files
 - name: Install/Update basics role packages and generate example configuration files to role subdir
+  vars:
+    __rsyslog_packages: "{{ __rsyslog_basics_packages }}"
+    __rsyslog_rules: "{{ __rsyslog_basics_rules }}"
   include_role:
     name: "{{ role_path }}/../../../"
     tasks_from: deploy.yaml
--- a/roles/rsyslog/roles/input_roles/files/defaults/main.yaml
+++ b/roles/rsyslog/roles/input_roles/files/defaults/main.yaml
@@ -2,7 +2,7 @@
 # Imfile configuration setting
 # ----------------------------
 
-rsyslog_unprivileged: False
+rsyslog_unprivileged: false
 
 # rsyslog_capabilities is an array which takes 'network', 'remote-files', 'tls'
 # 'network' enables input and output over network.
@@ -11,7 +11,7 @@ rsyslog_unprivileged: False
 #       Note: unles rsyslog_send_over_tls_only is set to true, insecure network connection
 #       is also allowed.
 rsyslog_capabilities: []
-rsyslog_files_input_packages: []
+__rsyslog_files_input_packages: []
 
 rsyslog_input_log_path: "/var/log/containers/*.log"
 rsyslog_input_log_tag: "container"
@@ -21,19 +21,19 @@ rsyslog_input_log_tag: "container"
 # List of YAML dictionaries, each dictionary should contain ``rsyslogd``
 # configuration in a special format. See :ref:`rsyslog_rules` for more
 # details. This list specifies imfile ``rsyslogd`` example configuration.
-rsyslog_files_input_rules:
-  - '{{ rsyslog_conf_imfile_modules }}'
-  - '{{ rsyslog_conf_imfile_rulesets }}'
-  - '{{ rsyslog_conf_imfile_inputs }}'
+__rsyslog_files_input_rules:
+  - '{{ __rsyslog_conf_imfile_modules }}'
+  - '{{ __rsyslog_conf_imfile_rulesets }}'
+  - '{{ __rsyslog_conf_imfile_inputs }}'
 
 # Debops example configuration options
 # --------------------------------------------
 
-# rsyslog_conf_local_modules
+# __rsyslog_conf_local_modules
 #
 # List of ``rsyslogd`` modules that enable logs from the local system to be
 # received and parsed by the ``rsyslogd`` daemon.
-rsyslog_conf_imfile_modules:
+__rsyslog_conf_imfile_modules:
 
   - name: 'imfile-modules'
     type: 'modules'
@@ -43,9 +43,9 @@ rsyslog_conf_imfile_modules:
         options: |-
           module(load="imfile" mode="inotify")
 
-# rsyslog_conf_imfile_rulesets
+# __rsyslog_conf_imfile_rulesets
 #
-rsyslog_conf_imfile_rulesets:
+__rsyslog_conf_imfile_rulesets:
 
   - name: 'imfile-rulesets'
     type: 'rules'
@@ -57,9 +57,9 @@ rsyslog_conf_imfile_rulesets:
             $IncludeConfig {{rsyslog_config_dir}}/*.files
           }
 
-# rsyslog_conf_system_inputs
+# __rsyslog_conf_system_inputs
 #
-rsyslog_conf_imfile_inputs:
+__rsyslog_conf_imfile_inputs:
 
   - name: 'imfile-input'
     type: 'input'
@@ -68,7 +68,7 @@ rsyslog_conf_imfile_inputs:
       - comment: 'Log messages from log files'
         options: |-
           # FIXME: Need to normalize a log before passing it to elasticsearch output
-          {% if rsyslog_elasticsearch is defined and rsyslog_elasticsearch|length > 0 %}
+          {% if rsyslog_elasticsearch is defined and rsyslog_elasticsearch | length > 0 %}
 
           {% for element in rsyslog_logs_collections %}
           {% if element.name is defined and element.type == "files" %}
--- a/roles/rsyslog/roles/input_roles/files/defaults/main.yaml
+++ b/roles/rsyslog/roles/input_roles/files/defaults/main.yaml
@@ -16,7 +16,7 @@ __rsyslog_files_input_packages: []
 rsyslog_input_log_path: "/var/log/containers/*.log"
 rsyslog_input_log_tag: "container"
 
-# rsyslog_files_input_rules
+# __rsyslog_files_input_rules
 #
 # List of YAML dictionaries, each dictionary should contain ``rsyslogd``
 # configuration in a special format. See :ref:`rsyslog_rules` for more
@@ -68,7 +68,7 @@ __rsyslog_conf_imfile_inputs:
       - comment: 'Log messages from log files'
         options: |-
           # FIXME: Need to normalize a log before passing it to elasticsearch output
-          {% if rsyslog_elasticsearch is defined and rsyslog_elasticsearch | length > 0 %}
+          {% if rsyslog_elasticsearch is defined and rsyslog_elasticsearch %}
 
           {% for element in rsyslog_logs_collections %}
           {% if element.name is defined and element.type == "files" %}
--- a/roles/rsyslog/roles/input_roles/files/tasks/cleanup.yaml
+++ b/roles/rsyslog/roles/input_roles/files/tasks/cleanup.yaml
@@ -1,10 +1,8 @@
 ---
-- name: Set input files facts
-  set_fact:
-    rsyslog_role_packages: "{{ rsyslog_files_input_packages|flatten }}"
-    rsyslog_role_rules: "{{ rsyslog_files_input_rules }}"
-
+# Deploy configuration files
 - name: Remove input files example config files from rsyslog.d
+  vars:
+    __rsyslog_rules: "{{ rsyslog_files_input_rules }}"
   include_role:
     name: "{{ role_path }}/../../../"
     tasks_from: cleanup.yaml
--- a/roles/rsyslog/roles/input_roles/files/tasks/cleanup.yaml
+++ b/roles/rsyslog/roles/input_roles/files/tasks/cleanup.yaml
@@ -2,7 +2,7 @@
 # Deploy configuration files
 - name: Remove input files example config files from rsyslog.d
   vars:
-    __rsyslog_rules: "{{ rsyslog_files_input_rules }}"
+    __rsyslog_rules: "{{ __rsyslog_files_input_rules }}"
   include_role:
     name: "{{ role_path }}/../../../"
     tasks_from: cleanup.yaml
--- a/roles/rsyslog/roles/input_roles/ovirt/tasks/main.yaml
+++ b/roles/rsyslog/roles/input_roles/ovirt/tasks/main.yaml
@@ -1,17 +1,16 @@
 ---
-- name: Set oVirt facts
-  set_fact:
-    rsyslog_role_packages: "{{ rsyslog_ovirt_prereq_packages|flatten|d([]) }}+{{ rsyslog_ovirt_packages|flatten|d([]) }}"
-    rsyslog_role_rules: "{{ rsyslog_ovirt_rules }}"
-
+# Deploy configuration files
 - name: Install/Update role packages and generate role configuration files to role subdir
+  vars:
+    __rsyslog_packages: "{{ __rsyslog_ovirt_prereq_packages | d([]) + __rsyslog_ovirt_packages | d([]) }}"
+    __rsyslog_rules: "{{ __rsyslog_ovirt_rules }}"
   include_role:
     name: "{{ role_path }}/../../../"
     tasks_from: deploy.yaml
 
 - name: Allow rsyslog to listen on collectd port
   seport:
-    ports: "{{ rsyslog_read_collectd_port|d(44514) }}"
+    ports: "{{ rsyslog_read_collectd_port | d(44514) }}"
     proto: tcp
     setype: syslogd_port_t
     state: present
--- a/roles/rsyslog/roles/input_roles/viaq-k8s/tasks/main.yaml
+++ b/roles/rsyslog/roles/input_roles/viaq-k8s/tasks/main.yaml
@@ -1,10 +1,9 @@
 ---
-- name: Set Viaq-k8s facts
-  set_fact:
-    rsyslog_role_packages: "{{ rsyslog_viaq_k8s_packages|flatten }}"
-    rsyslog_role_rules: "{{ rsyslog_viaq_k8s_rules }}"
-
+# Deploy configuration files
 - name: Install/Update role packages and generate role configuration files to role subdir
+  vars:
+    __rsyslog_packages: "{{ __rsyslog_viaq_k8s_packages }}"
+    __rsyslog_rules: "{{ __rsyslog_viaq_k8s_rules }}"
   include_role:
     name: "{{ role_path }}/../../../"
     tasks_from: deploy.yaml
--- a/roles/rsyslog/roles/input_roles/viaq/tasks/main.yaml
+++ b/roles/rsyslog/roles/input_roles/viaq/tasks/main.yaml
@@ -1,10 +1,9 @@
 ---
-- name: Set Viaq facts
-  set_fact:
-    rsyslog_role_packages: "{{ rsyslog_viaq_prereq_packages|flatten|d([]) }}+{{ rsyslog_viaq_packages|flatten|d([]) }}"
-    rsyslog_role_rules: "{{ rsyslog_viaq_rules }}"
-
+# Deploy configuration files
 - name: Install/Update role packages and generate role configuration files to role subdir
+  vars:
+    __rsyslog_packages: "{{ __rsyslog_viaq_prereq_packages + __rsyslog_viaq_packages }}"
+    __rsyslog_rules: "{{ __rsyslog_viaq_rules }}"
   include_role:
     name: "{{ role_path }}/../../../"
     tasks_from: deploy.yaml
--- a/roles/rsyslog/roles/output_roles/elasticsearch/tasks/main.yaml
+++ b/roles/rsyslog/roles/output_roles/elasticsearch/tasks/main.yaml
@@ -9,7 +9,7 @@
         - '{{ rsyslog_elasticsearch }}'
       when:
         - item.ca_cert_src is defined
-        - item.ca_cert_src != ""
+        - item.ca_cert_src
 
     - name: Copy local Elasticsearch certs to /etc/rsyslog.d directory in Rsyslog
       copy:
@@ -19,7 +19,7 @@
         - '{{ rsyslog_elasticsearch }}'
       when:
         - item.cert_src is defined
-        - item.cert_src != ""
+        - item.cert_src
 
     - name: Copy local Elasticsearch keys to /etc/rsyslog.d directory in Rsyslog
       copy:
@@ -29,21 +29,24 @@
         - '{{ rsyslog_elasticsearch }}'
       when:
         - item.key_src is defined
-        - item.key_src != ""
+        - item.key_src
 
     - name: Update ca_cert path to /etc/rsyslog.d directory in Rsyslog
       set_fact:
-        rsyslog_elasticsearch_temp: "{{ rsyslog_elasticsearch_temp | d([]) }} + {{ [ item | combine( { 'ca_cert': item.ca_cert_src | basename , 'cert': item.cert_src | basename, 'key': item.key_src | basename } ) ] }}"
+        __rsyslog_elasticsearch_temp: >-
+          {{ __rsyslog_elasticsearch_temp | d([]) }} +
+          {{ [ item | combine( { 'ca_cert': item.ca_cert_src | basename ,
+          'cert': item.cert_src | basename, 'key': item.key_src | basename } ) ] }}
       with_items:
         - '{{ rsyslog_elasticsearch }}'
       when:
         - item.ca_cert_src is defined
-        - item.ca_cert_src != ""
+        - item.ca_cert_src
       register: result
 
     - name: Set updated rsyslog_elasticsearch
       set_fact:
-        rsyslog_elasticsearch: "{{ rsyslog_elasticsearch_temp }}"
+        rsyslog_elasticsearch: "{{ __rsyslog_elasticsearch_temp }}"
 
   when:
     - rsyslog_elasticsearch is defined
@@ -51,12 +54,10 @@
     - use_local_omelasticsearch_cert | default(false) | bool
 
 # Deploy configuration files
-- name: Set Elasticsearch facts
-  set_fact:
-    rsyslog_role_packages: "{{ rsyslog_elasticsearch_packages | flatten([]) }}"
-    rsyslog_role_rules: "{{ rsyslog_elasticsearch_rules }}"
-
 - name: Install/Update role packages and generate role configuration files (elasticsearch)
+  vars:
+    __rsyslog_packages: "{{ __rsyslog_elasticsearch_packages }}"
+    __rsyslog_rules: "{{ __rsyslog_elasticsearch_rules }}"
   include_role:
     name: "{{ role_path }}/../../../"
     tasks_from: deploy.yaml
--- a/roles/rsyslog/roles/output_roles/elasticsearch/tasks/main.yaml
+++ b/roles/rsyslog/roles/output_roles/elasticsearch/tasks/main.yaml
@@ -9,7 +9,7 @@
         - '{{ rsyslog_elasticsearch }}'
       when:
         - item.ca_cert_src is defined
-        - item.ca_cert_src
+        - item.ca_cert_src | length > 0
 
     - name: Copy local Elasticsearch certs to /etc/rsyslog.d directory in Rsyslog
       copy:
@@ -19,7 +19,7 @@
         - '{{ rsyslog_elasticsearch }}'
       when:
         - item.cert_src is defined
-        - item.cert_src
+        - item.cert_src | length > 0
 
     - name: Copy local Elasticsearch keys to /etc/rsyslog.d directory in Rsyslog
       copy:
@@ -29,7 +29,7 @@
         - '{{ rsyslog_elasticsearch }}'
       when:
         - item.key_src is defined
-        - item.key_src
+        - item.key_src | length > 0
 
     - name: Update ca_cert path to /etc/rsyslog.d directory in Rsyslog
       set_fact:
@@ -41,7 +41,7 @@
         - '{{ rsyslog_elasticsearch }}'
       when:
         - item.ca_cert_src is defined
-        - item.ca_cert_src
+        - item.ca_cert_src | length > 0
       register: result
 
     - name: Set updated rsyslog_elasticsearch
--- a/roles/rsyslog/tasks/deploy.yaml
+++ b/roles/rsyslog/tasks/deploy.yaml
@@ -1,20 +1,23 @@
 ---
 - name: Install/Update required packages
   package:
-    name: "{{ rsyslog_role_packages }}"
+    name: "{{ __rsyslog_packages }}"
     state: latest
   when:
-    - not use_rsyslog_image | default(false) |bool
+    - not rsyslog_in_image | default(false) | bool
   notify: restart rsyslogd
 
 - name: Generate role configuration files in rsyslog.d and subdir
   template:
     src: 'etc/rsyslog.d/rules.conf.j2'
-    dest: '{{ item.path | d(rsyslog_config_dir) }}/{{ item.filename | d((item.weight if item.weight | d() else rsyslog_weight_map[item.type | d("rules")]) + "-" + (item.name | d("rules")) + "." + (item.suffix | d("conf"))) }}'
+    dest: >-
+      {{ item.path | d(rsyslog_config_dir) }}/{{ item.filename | d((item.weight
+      if item.weight | d() else rsyslog_weight_map[item.type | d("rules")]) +
+      "-" + (item.name | d("rules")) + "." + (item.suffix | d("conf"))) }}
     owner: '{{ item.owner | d("root") }}'
     group: '{{ item.group | d("root") }}'
     mode: '{{ item.mode  | d("0400") }}'
-  loop: '{{ rsyslog_role_rules | flatten }}'
+  loop: '{{ __rsyslog_rules | flatten }}'
   when:
     - item.filename | d() or item.name | d()
     - item.options | d() or item.sections | d()
--- a/roles/rsyslog/tasks/deploy.yaml
+++ b/roles/rsyslog/tasks/deploy.yaml
@@ -4,7 +4,7 @@
     name: "{{ __rsyslog_packages }}"
     state: latest
   when:
-    - not rsyslog_in_image | default(false) | bool
+    - not rsyslog_in_image | bool
   notify: restart rsyslogd
 
 - name: Generate role configuration files in rsyslog.d and subdir
--- a/roles/rsyslog/tasks/main.yaml
+++ b/roles/rsyslog/tasks/main.yaml
@@ -1,6 +1,6 @@
 - name: Install/Update required packages
   package:
-    name: "{{ (__rsyslog_base_packages) + (__rsyslog_tls_packages if rsyslog_pki else []) + (rsyslog_extra_packages) }}"
+    name: "{{ (__rsyslog_base_packages) + (__rsyslog_tls_packages if rsyslog_pki else []) + (rsyslog_extra_packages | flatten) }}"
     state: latest
   when:
     - rsyslog_enabled | bool
@@ -29,7 +29,7 @@
       set_fact:
         __rsyslog_failed_validation: true
       when: __rsyslog_version is version('8.37.0-7.2', '<')
-  when: use_rsyslog_image | default(false) | bool
+  when: rsyslog_in_image | bool
 
 - block:
     - name: Create required system group
@@ -87,15 +87,22 @@
         remove: '{{ true if rsyslog_purge_original_conf | bool else false }}'
       changed_when: false
 
+    - name: Get directory stat
+      stat:
+        path: "{{ rsyslog_system_log_dir }}"
+      register: stat_rsyslog_system_log_dir
+
     - name: Update directory and file permissions
       shell: |
         [ ! -d {{ rsyslog_system_log_dir }} ] ||
         ( [ "$(stat -c '%G' {{ rsyslog_system_log_dir }})" = "{{ rsyslog_group }}" ] ||
         ( chown -v root:{{ rsyslog_group }} {{ rsyslog_system_log_dir }} ;
-        chmod -v 775 {{ rsyslog_system_log_dir }} ) )
+        chmod -v 0775 {{ rsyslog_system_log_dir }} ) )
       register: rsyslog_register_file_permissions
       when: rsyslog_unprivileged | bool
-      changed_when: rsyslog_register_file_permissions.stdout
+      changed_when:
+        - rsyslog_register_file_permissions.rc == 0
+        - stat_rsyslog_system_log_dir.stat.mode != "0755"
 
     - name: Generate main rsyslog configuration
       template:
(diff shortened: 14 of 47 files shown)


nhosoi:
Note to myself...
This PR is depend on https://github.com/linux-system-roles/logging/pull/84/.
Once pr/84 is merged, solve the conflicts and merge this PR.

roles/rsyslog/tasks/main.yaml
```diff
@@ -1,19 +1,19 @@
 ---
+- name: Get required packages
+  set_fact: _rsyslog_packages="{{ (rsyslog_base_packages) + (rsyslog_tls_packages if rsyslog_pki else []) + (rsyslog_additional_packages) }}"
+
 - name: Install/Update required packages
   package:
-    name: "{{ lookup('flattened', rsyslog_base_packages, rsyslog_tls_packages if rsyslog_pki else [], rsyslog_packages) }}"
+    name: "{{ item }}"
     state: latest
+  loop: "{{ _rsyslog_packages | flatten | d('') }}"
```
pcahyna: Why this change? https://github.com/oasis-roles/meta_standards#ansible-best-practices says: Avoid calling the package module iteratively with the {{ item }} argument, as this is impressively more slow than calling it with the line name: "{{ foo_packages }}". The same can go for many other modules that can be given an entire list of items all at once.

nhosoi: Understood...   I introduced the loop to check whether each package is already installed or not here [1].  (Otherwise, I got an error from package/yum, so I had to put [2] to ignore it.  I thought it'd be better getting rid of [2].)  What'd be the best way to handle that kind of contradictory issue?  Do you recommend just passing the list to the `package` module and setting the ignore error?
[1] https://github.com/linux-system-roles/logging/pull/87/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32R13
[2] https://github.com/linux-system-roles/logging/pull/87/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32L9

pcahyna: What error did you get?

roles/rsyslog/defaults/main.yaml
```diff
@@ -67,10 +67,10 @@ rsyslog_base_packages: ['rsyslog']
 # List of rpm packages required for TLS support.
 rsyslog_tls_packages: ['rsyslog-gnutls', 'ca-certificates']
 
-# rsyslog_packages
+# rsyslog_additional_packages
 #
 # List of additional rpm packages to install.
-rsyslog_packages: []
+rsyslog_additional_packages: []
```
pcahyna: Do you use this anywhere? IMO this only complicates the role and breaks abstraction. If installing extra packages is needed for some functionality, there should be a parameter for the functionality and the role should know what packages to install, instead of requiring the user to know this (it may be different for different supported distributions and versions).
If we decide to keep this functionality, I suggest to follow the example of OASIS, where it is suggested to be named extra_packages instead of additional_packages: https://github.com/oasis-roles/meta_standards#vars-vs-defaults
@robled @greg-hellings @seandst have you found such a variable useful in practice?

roles/rsyslog/roles/input_roles/basics/defaults/main.yaml
```diff
@@ -12,6 +12,7 @@ rsyslog_unprivileged: false
 #       Note: unles rsyslog_send_over_tls_only is set to true, insecure network connection
 #       is also allowed.
 rsyslog_capabilities: []
+rsyslog_basics_packages: ['rsyslog-gnutls']
```
pcahyna: is the user supposed to override this? If not, it should go to vars, not defaults, and be prefixed with double underscore.

roles/rsyslog/roles/input_roles/basics/tasks/main.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 - name: Set basics facts
   set_fact:
-    rsyslog_role_packages: rsyslog-gnutls
+    rsyslog_role_packages: "{{ rsyslog_basics_packages }}"
```
pcahyna: shouldn't it be called just `rsyslog_packages`? Why the `role` in the name? Also, why not drop the `set_fact` and rename `rsyslog_basics_packages` to `rsyslog_packages` (or `__rsyslog_packages` if the variable is supposed to be internal, and I suspect it is)?

roles/rsyslog/roles/input_roles/ovirt/tasks/main.yaml
```diff
@@ -1,12 +1,12 @@
 ---
 - name: Set oVirt facts
   set_fact:
-    rsyslog_role_packages: "{{ rsyslog_ovirt_prereq_packages|flatten|d([]) }}+{{ rsyslog_ovirt_packages|flatten|d([]) }}"
+    rsyslog_role_packages: "{{ rsyslog_ovirt_prereq_packages | d([]) + rsyslog_ovirt_packages | d([]) }}"
```
pcahyna: See above - can it be renamed to `__rsyslog_packages` and moved to vars, eliminating `rsyslog_ovirt_packages` ?
What will happen if one uses both the `basics` and `ovirt` subroles? Won't the second `set_fact` clobber the result of the former? is that expected?

roles/rsyslog/roles/input_roles/viaq-k8s/tasks/main.yaml
```diff
@@ -1,10 +1,10 @@
 ---
 - name: Set Viaq-k8s facts
   set_fact:
-    rsyslog_role_packages: "{{ rsyslog_viaq_k8s_packages|flatten }}"
+    rsyslog_role_packages: "{{ rsyslog_viaq_k8s_packages }}"
```
pcahyna: see above...

roles/rsyslog/roles/input_roles/viaq/tasks/main.yaml
```diff
@@ -1,10 +1,10 @@
 ---
 - name: Set Viaq facts
   set_fact:
-    rsyslog_role_packages: "{{ rsyslog_viaq_prereq_packages|flatten|d([]) }}+{{ rsyslog_viaq_packages|flatten|d([]) }}"
+    rsyslog_role_packages: "{{ rsyslog_viaq_prereq_packages + rsyslog_viaq_packages }}"
```
pcahyna: see above

roles/rsyslog/roles/output_roles/elasticsearch/tasks/main.yaml
```diff
@@ -53,7 +53,7 @@
 # Deploy configuration files
 - name: Set Elasticsearch facts
   set_fact:
-    rsyslog_role_packages: "{{ rsyslog_elasticsearch_packages | flatten([]) }}"
+    rsyslog_role_packages: "{{ rsyslog_elasticsearch_packages }}"
```
pcahyna: see above

tests/tests_listen.yml
```diff
@@ -30,6 +30,8 @@
       package:
         name: lsof
         state: present
+      when:
+        - not 'lsof' in ansible_facts.packages
```
pcahyna: I know this makes the test a bit faster, but is it worth complicating the code and making it more fragile because of this optimization?

roles/rsyslog/roles/input_roles/ovirt/tasks/main.yaml
pcahyna: maybe I misunderstood - I see there are `rsyslog_base_packages` and `rsyslog_role_packages` variables with different functions - why is it the case?

roles/rsyslog/tasks/main.yaml
pcahyna: (and yes, using `ignore_errors` like [2] is usually not a very good idea. But other roles and the best practices have been using package list without a loop and without `ignore_errors`. If there is a problem with applying the best practices, please investigate it and report it, instead of doing things differently.)

roles/rsyslog/tasks/main.yaml
```diff
@@ -1,19 +1,19 @@
...
-  ignore_errors: yes
-  tags:
-    - skip_ansible_lint
+    - not item in ansible_facts.packages
 
 - name: Get rsyslog version
-  package:
-    list: rsyslog
-  register: rsyslog_version
+  set_fact: rsyslog_version=ansible_facts.packages['rsyslog'][0].version | default('') | int
```
pcahyna: Doesn't the block below need adjusting? Also, does it need to be a fact instead of a variable local to the block?

roles/rsyslog/roles/input_roles/basics/defaults/main.yaml
nhosoi: Rsyslog supports openssl, as well.  But it looks the plugin is not included in the RHEL/CentOS distribution.  So, there's no chance to override for now.  I will follow your advice.  Thanks!

roles/rsyslog/roles/input_roles/basics/tasks/main.yaml
nhosoi: The var rsyslog_role_packages is used here [1].  They need to match.  The current rsyslog_role_packages could be set in each input and output sub-roles.  So, I think we should replace rsyslog_role_packages with __rsyslog_packages to make it internal, and replace each rsyslog_something_packages with __rsyslog_something_packages?
[1] https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/tasks/deploy.yaml#L2-L8

roles/rsyslog/roles/input_roles/ovirt/tasks/main.yaml
nhosoi: The rsyslog_role_packages are taken care in roles/rsyslog/tasks/deploy.yaml vir each deploy.yaml following tasks_from line (e.g., https://github.com/linux-system-roles/logging/pull/87/files#diff-32e44b317e7ab0e0016656a25254fa0bL10, here).  So, ovirt packages are installed immediately after this set_fact and the same happens for basics, too.

> Won't the second set_fact clobber the result of the former?

So, the answer is "no".  It does not clobber the result of the former, but the value of rsyslog_role_packages is used immediately before being clobbered.

roles/rsyslog/tasks/main.yaml
nhosoi: Hmmm, even though I reverted this change, I cannot reproduce the error...  Maybe, the cause was something else.  Sorry for the noise.  I keep reverting this change for now.

tests/tests_listen.yml
nhosoi: If it's not preferable, I'm removing the line 33 and 34.  I looked into other linux-system-roles and thought it's recommended. :)

roles/rsyslog/defaults/main.yaml
nhosoi: I understand the issue, but I reluctant to remove this variable...
The rsyslog configuration is quite flexible and it supports these plugins.  Some of them are not installed by the current logging role.  But customers may want to manually configure it in their environment.  To cover such a use case, this rsyslog_additional_packages might be useful.
```
rsyslog-crypto-8.1911.0-3.el8.x86_64.rpm
rsyslog-elasticsearch-8.1911.0-3.el8.x86_64.rpm
rsyslog-gnutls-8.1911.0-3.el8.x86_64.rpm
rsyslog-gssapi-8.1911.0-3.el8.x86_64.rpm
rsyslog-kafka-8.1911.0-3.el8.x86_64.rpm
rsyslog-mmaudit-8.1911.0-3.el8.x86_64.rpm
rsyslog-mmjsonparse-8.1911.0-3.el8.x86_64.rpm
rsyslog-mmkubernetes-8.1911.0-3.el8.x86_64.rpm
rsyslog-mmnormalize-8.1911.0-3.el8.x86_64.rpm
rsyslog-mmsnmptrapd-8.1911.0-3.el8.x86_64.rpm
rsyslog-mysql-8.1911.0-3.el8.x86_64.rpm
rsyslog-pgsql-8.1911.0-3.el8.x86_64.rpm
rsyslog-relp-8.1911.0-3.el8.x86_64.rpm
rsyslog-snmp-8.1911.0-3.el8.x86_64.rpm
```

roles/rsyslog/roles/input_roles/basics/tasks/main.yaml
nhosoi: I modified the code like this
https://github.com/linux-system-roles/logging/pull/87/files#diff-532de15639f14a0202b8cd2880a8bd4cR4-R10
with the comment:
- To call include_role, instead of setting variables to pass with set_fact, put them in the block with vars to make them local in the scope.

Please note that I made the same change to all the include_role.

roles/rsyslog/roles/input_roles/ovirt/tasks/main.yaml
nhosoi: https://github.com/linux-system-roles/logging/pull/87#discussion_r381552024 is applied here, as well.

roles/rsyslog/roles/input_roles/viaq-k8s/tasks/main.yaml
nhosoi: #87 (comment) is applied here, as well.

roles/rsyslog/roles/input_roles/viaq/tasks/main.yaml
nhosoi: #87 (comment) is applied here, as well.

roles/rsyslog/roles/output_roles/elasticsearch/tasks/main.yaml
nhosoi: #87 (comment) is applied here, as well.

nhosoi:
Hi, @pcahyna, @sradco, @richm, could you please review this pr one more time?  I've updated the patch to eliminate set_fact for include_role.

Actually, I have another patch to make the logging role more compliant to the best practices [1].  (I think this one is rather straightforward although it is huge... :)  It's still my private pr, but once this pr/87 is merged, I'm going to submit an official pull request from it.  Thanks!
[1] https://github.com/nhosoi/logging/pull/1

roles/rsyslog/roles/input_roles/basics/tasks/main.yaml
pcahyna: You don't need a block. Put vars: directly on the include_role statement.

nhosoi: Thanks, @pcahyna.  Fixed.

roles/rsyslog/README.md
```diff
@@ -338,7 +338,7 @@ WARNING: Pre-existing rsyslog.conf and configuration files in /etc/rsyslog.d are
 
 Describing how the configuration files are defined to be deployed using the viaq case.
 
-Viaq configuration files are defined in {{ rsyslog_viaq_rules }} in /roles/input_rols/viaq/defaults/main.yaml.  The set is made from the generic modules{rsyslog_conf_global_options, rsyslog_conf_local_modules, rsyslog_conf_network_modules, rsyslog_conf_common} and viaq specific configurations.
+Viaq configuration files are defined in {{ __rsyslog_viaq_rules }} in /roles/input_rols/viaq/defaults/main.yaml.  The set is made from the generic modules{rsyslog_conf_global_options, rsyslog_conf_local_modules, rsyslog_conf_network_modules, rsyslog_conf_common} and viaq specific configurations.
```
pcahyna: If you are prefixing the variable name with underscores, it is internal and should it be documented then at all? (maybe it should, but in some developer/internal focused documentation, not in user documentation.) also, there is a typo in `input_rols`

roles/rsyslog/defaults/main.yaml
pcahyna: I see. Still, can't you introduce more generic variables to control whether a specific functionality will be enabled and the plugin installed? I have tho concerns:
- the name of a plugin implementing a specific functionality can change, the role should isolate the user from such changes, but if the user has to know the exact names of packages, it does not fulfil this function,
- presumably to make use of a plugin, sometimes it could be necessary to configure it in addition to installing it (is that the case?), and this again calls for a more high-level approach, because when the role just installs given plugins, it will not help the user with configuring them.
If you decide to keep the variable, I would suggest to at least rename it to the OASIS convention.

tests/tests_files_forwards.yml
```diff
@@ -51,27 +53,23 @@
...
     - name: Get the files config stat
       stat:
-        path: "{{ test_files_conf }}"
+        path: "{{ __test_files_conf }}"
       register: stat_output_files_system
 
     - name: Check if the files config exists
       assert:
-        that: stat_output_files_system.stat.exists == True
+        that: stat_output_files_system.stat.exists == true
```
pcahyna: why compare a boolean to true? won't it work to just use `stat_output_files_system.stat.exists` ?

roles/rsyslog/README.md
nhosoi: :p  thanks!

roles/rsyslog/defaults/main.yaml
```diff
@@ -291,60 +281,29 @@ rsyslog_weight_map:
...
-# rsyslog_.*_rules
-#
-# List of YAML dictionaries, each dictionary should contain ``rsyslogd``
-# configuration in a special format. See :ref:`rsyslog_rules` for more
-# details. This lis specifies default ``rsyslogd`` configuration enabled in the
-# role.
-rsyslog_common_rules:
+# configuration in a special format. This lis specifies default ``rsyslogd``
+# configuration enabled in the role.
+__rsyslog_common_rules:
```
pcahyna: If you are prepending underscores to variable names, does it mean that the variables are internal and therefore don't need to be settable by users? - but then they belong to vars and not defaults, no?

Also I would say that this diverges from the topic of fixing package module related issues.

nhosoi: > If you are prepending underscores to variable names, does it mean that the variables are internal and therefore don't need to be settable by users? - but then they belong to vars and not defaults, no?

That's correct.  They should belong to vars.

> Also I would say that this diverges from the topic of fixing package module related issues.

+1.  May I stop this here and move to another PR for the internal var fixes (including docs)?

nhosoi:
@pcahyna, could you please help me?  With the changes I made today, the molecule (ansible lint) started failing like this [1].  But I could not find out where the failure "{'lint': ['must be of string type']}" comes from... :(

https://travis-ci.org/linux-system-roles/logging/jobs/653146080?utm_medium=notification&utm_source=github_status

pcahyna:
> @pcahyna, could you please help me? With the changes I made today, the molecule (ansible lint) started failing like this [1]. But I could not find out where the failure "{'lint': ['must be of string type']}" comes from... :(
> 
> https://travis-ci.org/linux-system-roles/logging/jobs/653146080?utm_medium=notification&utm_source=github_status

I am about to leave. Please ask @richm

roles/rsyslog/defaults/main.yaml
pcahyna: yes please, finish the package related changes in this PR and continue variable renaming (if not related to the package module changes) in another PR.

roles/rsyslog/tasks/main.yaml
```diff
@@ -1,43 +1,31 @@
...
+- name: Gather package facts
+  package_facts:
+    manager: auto
 
 - name: Get rsyslog version
-  package:
-    list: rsyslog
-  register: rsyslog_version
+  set_fact:
+    __rsyslog_version: "{{ ansible_facts.packages['rsyslog'][0].version }}"
```
pcahyna: Hmm I thought about a problem, what if rsyslog is not installed here? With @vcrhonek we have been discussing a similar problem in the timesync role.

tests/tests_files_forwards.yml
pcahyna: this applies to other places as well - but please test it first

richm:
Looks like we are now using molecule v3?
```
>pwd
/home/rmeggins/linux-system-roles/logging
>git pull
up-to-date
>python -mvenv .venv
>. .venv/bin/activate
>pip install molecule docker
>molecule --version
molecule 3.0.1
   ansible==2.9.5 python==3.7
```
and they changed the molecule config?  https://github.com/ansible-community/molecule/blob/master/docs/configuration.rst
```
Lint

Starting with v3, Molecule handles project linting by invoking and external lint commands as exemplified below.

The decision to remove the complex linting support was not easily taken as we do find it very useful. The issue was that molecule runs on scenarios and linting is usually performed at repository level.

It makes little sense to perform linting in more than one place per project. Molecule was able to use up to three linters and while it was aimed to flexible about them, it ended up creating more confusions to the users. We decided to maximize flexibility by just calling an external shell command.

lint: |
    yamllint .
    ansible-lint
    flake8

The older format is no longer supported and you have to update the molecule.yml when you upgrade. If you don't want to do any linting, it will be enough to remove all lint related sections from the file.

# old v2 format, no longer supported
lint:
  name: yamllint
  enabled: true
...
```

richm:
@nhosoi so what I would suggest is to update the molecule config for v3 - you can test it locally as I have above
If that seems too difficult, we can change the tests - travis - whatever to use molecule v2.  For example:
```
>pip install 'molecule<2'
...
>molecule --version
molecule, version 2.22
```

richm:
the molecule config file is `molecule/default/molecule.yml`

nhosoi:
> molecule/default/molecule.yml

Thanks a lot, @richm.  By upgrading the molecule, I see more errors from ansible-lint... :)  @richm, @pcahyna, is it ok to fix them in this pr/87?  Or shall I just merge this pr with the failed travis and submit a new pull request for the lint fixes?

richm:
I think the molecule config/lint fixes should be in a separate PR - that's a lot of stuff . . .

roles/rsyslog/tasks/main.yaml
nhosoi: > Hmm I thought about a problem, what if rsyslog is not installed here? 

Could it happen even if rsyslog is installed with the package module just above this task?

nhosoi:
Thanks, @richm!
> I think the molecule config/lint fixes should be in a separate PR - that's a lot of stuff . . .

@pcahyna, is it ok to merge this pr without the second commit "Fixing travis-ci"?
https://github.com/linux-system-roles/logging/pull/87/commits/c772da4a0045e0f45608a28406704db7a2c6d87d

pcahyna:
@i386x @tyll @vcrhonek @xjezda00 this is something to be taken account of in our template: https://github.com/linux-system-roles/logging/pull/87#issuecomment-589298865
FYI @robled @greg-hellings @seandst

roles/rsyslog/tasks/main.yaml
pcahyna: yes, think about check mode.

nhosoi: > yes, think about check mode.

I see.  Do you recommend to add `check_mode: no` to the install packages tasks?

pcahyna: No, this would defeat the purpose of using check mode, which is to not make changes to the system.

roles/rsyslog/defaults/main.yaml
pcahyna: and please rename `rsyslog_packages` to `rsyslog_extra_packages` instead of to `rsyslog_additional_packages`, the former seems more idiomatic to me.

roles/rsyslog/tasks/deploy.yaml
```diff
@@ -1,20 +1,23 @@
 ---
 - name: Install/Update required packages
   package:
-    name: "{{ rsyslog_role_packages }}"
+    name: "{{ __rsyslog_packages }}"
     state: latest
```
pcahyna: I propose to use `present`, not `latest`. Most of our roles use this, so let's be consistent. @robled @greg-hellings @seandst do you have an opinion here?
If we want to be able to use `latest`, we can introduce a _xxx_`_packages_state` variable, like here:

https://github.com/geerlingguy/ansible-role-apache/blob/77f39119bc12fd839d16a55a19354d5c159041fa/defaults/main.yml#L58

roles/rsyslog/tasks/main.yaml
```diff
@@ -1,44 +1,25 @@
----
 - name: Install/Update required packages
   package:
-    name: "{{ lookup('flattened', rsyslog_base_packages, rsyslog_tls_packages if rsyslog_pki else [], rsyslog_packages) }}"
+    name: "{{ (__rsyslog_base_packages) + (__rsyslog_tls_packages if rsyslog_pki else []) + (rsyslog_additional_packages) }}"
     state: latest
```
pcahyna: see the note above regarding `latest`

tests/tests_listen.yml
```diff
@@ -29,25 +28,36 @@
     - name: Install lsof
       package:
         name: lsof
-        state: present
+        state: latest
```
pcahyna: do you need it? I suggest to omit this change.

roles/rsyslog/tasks/main.yaml
nhosoi: Indeed...  We can do either 1) quit with no such package error here or 2) pretend having an rsyslog package and keep running since it is a dry-run.  Which would be better?

roles/rsyslog/tasks/main.yaml
nhosoi: I'm wondering when the system has the older rsyslog package is installed and there is a newer rsyslog in the repository, what happens if the state is present vs. latest.  Can we make sure we could install the latest package?

tests/tests_listen.yml
nhosoi: I don't care the version of lsof...  So, reverting this. :)

roles/rsyslog/tasks/main.yaml
pcahyna: It hasn't been a problem for other system roles apparently.

roles/rsyslog/tasks/main.yaml
pcahyna: I would say some version of 2) since I don't think we should fail in check mode if we would not fail in normal mode. We could even skip the rest of the role since we properly reported changed status.

pcahyna:
> > molecule/default/molecule.yml
> 
> Thanks a lot, @richm. By upgrading the molecule, I see more errors from ansible-lint... :) @richm, @pcahyna, is it ok to fix them in this pr/87? Or shall I just merge this pr with the failed travis and submit a new pull request for the lint fixes?

@i386x @tyll have you seen such problems in other roles? If not, why not? We are using molecule elsewhere.

roles/rsyslog/tasks/main.yaml
nhosoi: Ok, it's guaranteed to be upgraded with `present`.  Reading this doc, I was not sure it was. :)
https://docs.ansible.com/ansible/latest/modules/package_module.html#parameter-state

I'm going to revert "latest" changes...

richm: > what happens if the state is present vs. latest

If the rsyslog package does not exist, then both `state: present` and `state: latest` will install the latest package available.

If the rsyslog package does exist, then `state: present` is a no-op, and `state: latest` will do an upgrade to the latest package available, if there is a newer one available.

richm: Also - if we do not hardcode `state: latest` in the role, then we need some other mechanism like https://github.com/geerlingguy/ansible-role-apache/blob/77f39119bc12fd839d16a55a19354d5c159041fa/defaults/main.yml#L58 - also note that if we do this, we should be consistent among *all* of the system roles - that is, every system role will need to have a `ROLE_packages_state: present` in the `defaults/main.yml` - we'll also need to add this to the README.md, the template, and the best practices guide.

richm:
> > > molecule/default/molecule.yml
> > 
> > 
> > Thanks a lot, @richm. By upgrading the molecule, I see more errors from ansible-lint... :) @richm, @pcahyna, is it ok to fix them in this pr/87? Or shall I just merge this pr with the failed travis and submit a new pull request for the lint fixes?
> 
> @i386x @tyll have you seen such problems in other roles? If not, why not? We are using molecule elsewhere.

You would only see this if you are actively working on a role, or actively working on a PR, within the past few days, after the version of molecule on pypi was updated to v3.

roles/rsyslog/tasks/main.yaml
nhosoi: > > what happens if the state is present vs. latest
> If the rsyslog package does exist, then `state: present` is a no-op, and `state: latest` will do an upgrade to the latest package available, if there is a newer one available.

Thanks, @richm.  Hmmm, then I'd like to keep `latest` for rsyslog package install/upgrade.  The logic is also used for updating security related packages.  So, I think we need to make sure it updates to the latest.

pcahyna: Ok, keep it for now, but we should set a common standard for all the roles, because I don't see why logging should be different.

nhosoi: Thanks, @pcahyna.  Just out of curiosity, there's no requirement for other roles to _upgrade_ existing packages?

pcahyna: I am not aware of such a requirement. As you can see above, the default for many popular roles on Galaxy is to not update.

tyll:
> > > molecule/default/molecule.yml
> > 
> > 
> > Thanks a lot, @richm. By upgrading the molecule, I see more errors from ansible-lint... :) @richm, @pcahyna, is it ok to fix them in this pr/87? Or shall I just merge this pr with the failed travis and submit a new pull request for the lint fixes?
> 
> @i386x @tyll have you seen such problems in other roles? If not, why not? We are using molecule elsewhere.

What are the acutal errors, I saw this:
https://travis-ci.com/linux-system-roles/network/jobs/289283294#L460 in https://github.com/linux-system-roles/network/pull/153

```
verifier:
  lint:
    name: flake8
  name: testinfra

ERROR: Failed to pre-validate.

{'lint': ['must be of string type']}
ERROR: InvocationError for command /bin/bash .travis/runsyspycmd.sh .tox/env-3.5/bin/python /usr/bin/python3 molecule lint -s default (exited with code 1)
___________________________________ summary ____________________________________
ERROR:   molecule: commands failed
```

Sounds like it could be the issue.

richm:
> {'lint': ['must be of string type']}

Yes, this is the issue - this means the molecule.yml config needs to be upgraded to support molecule v3.

tyll:
> > {'lint': ['must be of string type']}
> 
> Yes, this is the issue - this means the molecule.yml config needs to be upgraded to support molecule v3.

What should it look like now?

```
 lint: |
    yamllint --config-file molecule/default/yamllint.yml .
```

checks also everything in the tox directory...

richm:
> > > {'lint': ['must be of string type']}
> > 
> > 
> > Yes, this is the issue - this means the molecule.yml config needs to be upgraded to support molecule v3.
> 
> What should it look like now?
> 
> ```
>  lint: |
>     yamllint --config-file molecule/default/yamllint.yml .
> ```
> 
> checks also everything in the tox directory...

I don't know - here is the document about the new config:  https://github.com/ansible-community/molecule/blob/master/docs/configuration.rst

@nhosoi - did you run into any problems with the checker descending into venv or tox directories, and if so, how did you disable that?

nhosoi:
> > > > {'lint': ['must be of string type']}
> > > 
> > > 
> > > Yes, this is the issue - this means the molecule.yml config needs to be upgraded to support molecule v3.
> > 
> > 
> > What should it look like now?
> > ```
> >  lint: |
> >     yamllint --config-file molecule/default/yamllint.yml .
> > ```
> > 
> > 
> > checks also everything in the tox directory...
> 
> I don't know - here is the document about the new config: https://github.com/ansible-community/molecule/blob/master/docs/configuration.rst
> 
> @nhosoi - did you run into any problems with the checker descending into venv or tox directories, and if so, how did you disable that?

Ahhh, I was not running yamllint correctly.  Fixed it.  I listed subdirectories to test instead of '.'.

pcahyna:
@nhosoi I think you don't need to worry about Travis tests, because we have the same problems in other roles and we need to solve them together.

nhosoi:
> @nhosoi I think you don't need to worry about Travis tests, because we have the same problems in other roles and we need to solve them together.

@pcahyna, ok.  Then, what shall I do?  Is it ok to merge this pr or further reviews/fixes are needed?  If it's ok to merge, is it with Fixing travis-ci https://github.com/linux-system-roles/logging/pull/87/commits/9d606287ec06e5a02e6a98c69d7664f768047fcc, or without it?  Please note that the patch contains the fixes reported by the travis tests, as well.  So, if it's allowed, I'd like to merge both of the patches https://github.com/linux-system-roles/logging/pull/87/commits/a27c2c0f8eb50edababb400cd8416f8cab1a97c7 and https://github.com/linux-system-roles/logging/pull/87/commits/9d606287ec06e5a02e6a98c69d7664f768047fcc together.  Thanks.

pcahyna:
Sure, if you have the issue already fixed, keep the fix, I meant that you don't need to work further on a fix sepcifically for logging now.

nhosoi:
> Sure, if you have the issue already fixed, keep the fix, I meant that you don't need to work further on a fix sepcifically for logging now.

@pcahyna, thank you!  Now, may I merge this patch?  Or further reviews are needed?  If it's ok to merge, could some of you please approve this pr?  Thanks! > @pcahyna, @richm, @sradco

roles/rsyslog/defaults/main.yaml
```diff
@@ -291,60 +286,29 @@ rsyslog_weight_map:
...
-rsyslog_host_rules: []
-
-# rsyslog_dependent_rules
-#
-# List of YAML dictionaries, each dictionary should contain ``rsyslogd``
-# configuration in a special format. See :ref:`rsyslog_rules` for more
-# details. This list should be used for configuration by other Ansible roles.
-rsyslog_dependent_rules: []
-
-# rsyslog_.*_rules
```
pcahyna: how is the deleted block above related to the subject of this PR?

roles/rsyslog/defaults/main.yaml
```diff
@@ -383,11 +347,57 @@ rsyslog_conf_local_modules:
...
+            {% else %}
+              permittedPeer=["{{ rsyslog_permitted_peers | join('","') }}"]
+            {% endif %}
+          {% endif %}
+          )
+        state: '{{ "present"
+                  if ("tls" in rsyslog_capabilities)
+                  else "absent" }}'
+
+# __rsyslog_conf_common_defaults
```
pcahyna: this is a big change, what is the relation to the rest of the PR?

pcahyna:
@nhosoi I looked at the PR again and found two big changes that seem to be unrelated to the PR description, are they intended?

roles/rsyslog/defaults/main.yaml
nhosoi: Since I found these "rules" are not used anywhere, I cleaned this up.  Do you want me to move this to a new pr?

roles/rsyslog/defaults/main.yaml
nhosoi: It's not...  I'm removing this section.  Sorry for the noise. :(

nhosoi:
@pcahyna, they were part of cleaning up, but they should not be in this pr.  Sorry, I'm removing them.

pcahyna:
@nhosoi can you please review the PR again for unintended changes yourself?

richm:
please revert the change to ` molecule/default/molecule.yml` - instead, change your `tox.ini` like this:
```
--- a/tox.ini
+++ b/tox.ini
@@ -176,7 +176,7 @@ commands =
 deps =
     {env:LSR_ANSIBLE_DEP:ansible}
     docker
-    molecule
+    molecule<3
     selinux
     -rmolecule_extra_requirements.txt
```
The plan is to use molecule v2 until we complete all of the work necessary to port all of the system roles to v3

pcahyna:
@richm do we want to save that change in a branch, since it seems to work and could be used as a basis for further changes?

richm:
> @richm do we want to save that change in a branch, since it seems to work and could be used as a basis for further changes?

Yes, good idea

nhosoi:
> please revert the change to ` molecule/default/molecule.yml` - instead, change your `tox.ini` like this:
> 
> ```
> --- a/tox.ini
> +++ b/tox.ini
> @@ -176,7 +176,7 @@ commands =
>  deps =
>      {env:LSR_ANSIBLE_DEP:ansible}
>      docker
> -    molecule
> +    molecule<3
>      selinux
>      -rmolecule_extra_requirements.txt
> ```
> 
> The plan is to use molecule v2 until we complete all of the work necessary to port all of the system roles to v3

It seems there's no tox.ini in the logging roles... :(

richm:
> It seems there's no tox.ini in the logging roles... :(

yet another enhancement/fixup to bring this role into sync with the other roles . . . 

for now - looks like molecule is installed in a couple of places:
```
.travis.yml:  - pip install molecule docker
README.md:$ pip install molecule docker
```
change both of these to `'molecule<3'`
also - if you want to test locally, and you already have molecule v3, do this:
```
pip uninstall molecule
pip install 'molecule<3'
```

roles/rsyslog/defaults/main.yaml
nhosoi: Done.

roles/rsyslog/tasks/main.yaml
nhosoi: Note: I reverted the code pointed out by @pcahyna.

roles/rsyslog/defaults/main.yaml
nhosoi: I renamed the misleading var name rsyslog_additional_packages with rsyslog_extra_packages as suggested.

roles/rsyslog/roles/input_roles/basics/defaults/main.yaml
nhosoi: Note: removed the var.

roles/rsyslog/roles/input_roles/ovirt/tasks/main.yaml
nhosoi: Note: var name `rsyslog_role_packages` was renamed to `__rsyslog_packages` and the var is set with instead of `set_fact` but in `var:`.

roles/rsyslog/roles/input_roles/viaq/tasks/main.yaml
nhosoi: Note: var name `rsyslog_role_packages` was renamed to `__rsyslog_packages` and the var is set with instead of `set_fact` but in `var:`.

roles/rsyslog/roles/output_roles/elasticsearch/tasks/main.yaml
nhosoi: Note: var name `rsyslog_role_packages` was renamed to `__rsyslog_packages` and the var is set with instead of `set_fact` but in `var:`.

roles/rsyslog/roles/input_roles/viaq-k8s/tasks/main.yaml
nhosoi: Note: var name `rsyslog_role_packages` was renamed to `__rsyslog_packages` and the var is set with instead of `set_fact` but in `var:`.

tests/tests_listen.yml
nhosoi: Note: `when: not 'lsof' in ansible_facts.packages` was removed;

roles/rsyslog/tasks/main.yaml
nhosoi: Fixed.

roles/rsyslog/README.md
nhosoi: The typo input_rols was fixed.
Documentation for the internal vars will be added in a separate pr.

roles/rsyslog/tasks/deploy.yaml
nhosoi: Leaving the state as latest to make sure the newer packages are installed/updated if available.
Leaving the conversation opened.

nhosoi:
@pcahyna,
> @nhosoi can you please review the PR again for unintended changes yourself?

I went through the pr and removed the 2 changes out of the scope of this pr, which you pointed out.  I also updated the commit comments to describe the patches.  Thanks.

nhosoi:
> > It seems there's no tox.ini in the logging roles... :(
> 
> yet another enhancement/fixup to bring this role into sync with the other roles . . .

@richm, I looked into the other roles.  (I don't plan to add it to this pr.  Please don't worry... :)  The network role has tox.ini, but not timesync and selinux.  I see tox.ini in template.  That means it's new and recommended to add?

richm:
> That means it's new and recommended to add?

It means we will fix it later, if necessary

roles/rsyslog/defaults/main.yaml
```diff
@@ -322,29 +317,28 @@ rsyslog_host_rules: []
...
 
-# rsyslog_.*_rules
+# __rsyslog_.*_rules
 #
 # List of YAML dictionaries, each dictionary should contain ``rsyslogd``
-# configuration in a special format. See :ref:`rsyslog_rules` for more
-# details. This lis specifies default ``rsyslogd`` configuration enabled in the
-# role.
-rsyslog_common_rules:
+# configuration in a special format. This lis specifies default ``rsyslogd``
```
sradco: @nhosoi list* instead of lis*

roles/rsyslog/roles/README.md
```diff
@@ -21,7 +21,7 @@ The `defaults` directory includes:
 
 The `tasks` directory includes 2 tasks file:
   - `main.yaml` - tasks for deploying the config files
-    This file is sets `rsyslog_role_packages` and `rsyslog_role_rules` and includes the task that deploys the files.
+    This file is sets `__rsyslog_packages` and `__rsyslog_rules` and includes the task that deploys the files.
```
sradco: @nhosoi This file sets. I think we can drop the "is".

roles/rsyslog/roles/input_roles/basics/tasks/main.yaml
```diff
@@ -1,10 +1,9 @@
 ---
-- name: Set basics facts
-  set_fact:
-    rsyslog_role_packages: "{{rsyslog_basics_packages | flatten | d([])}}"
-    rsyslog_role_rules: "{{ rsyslog_basics_rules }}"
-
+# Deploy configuration files
 - name: Install/Update basics role packages and generate example configuration files to role subdir
+  vars:
+    __rsyslog_packages: "{{ __rsyslog_basics_packages }}"
```
sradco: @nhosoi you don't need faltten?

roles/rsyslog/roles/input_roles/files/tasks/cleanup.yaml
```diff
@@ -1,10 +1,8 @@
 ---
-- name: Set input files facts
-  set_fact:
-    rsyslog_role_packages: "{{ rsyslog_files_input_packages|flatten }}"
-    rsyslog_role_rules: "{{ rsyslog_files_input_rules }}"
-
+# Deploy configuration files
 - name: Remove input files example config files from rsyslog.d
+  vars:
+    __rsyslog_rules: "{{ rsyslog_files_input_rules }}"
```
sradco: @nhosoi You don't need __rsyslog_packages here?

roles/rsyslog/roles/input_roles/ovirt/tasks/main.yaml
```diff
@@ -1,17 +1,16 @@
 ---
-- name: Set oVirt facts
-  set_fact:
-    rsyslog_role_packages: "{{ rsyslog_ovirt_prereq_packages|flatten|d([]) }}+{{ rsyslog_ovirt_packages|flatten|d([]) }}"
-    rsyslog_role_rules: "{{ rsyslog_ovirt_rules }}"
-
+# Deploy configuration files
 - name: Install/Update role packages and generate role configuration files to role subdir
+  vars:
+    __rsyslog_packages: "{{ __rsyslog_ovirt_prereq_packages | d([]) + __rsyslog_ovirt_packages | d([]) }}"
```
sradco: @nhosoi Why did you remove the "flatten"?

roles/rsyslog/roles/input_roles/viaq/tasks/main.yaml
```diff
@@ -1,10 +1,9 @@
 ---
-- name: Set Viaq facts
-  set_fact:
-    rsyslog_role_packages: "{{ rsyslog_viaq_prereq_packages|flatten|d([]) }}+{{ rsyslog_viaq_packages|flatten|d([]) }}"
-    rsyslog_role_rules: "{{ rsyslog_viaq_rules }}"
-
+# Deploy configuration files
 - name: Install/Update role packages and generate role configuration files to role subdir
+  vars:
+    __rsyslog_packages: "{{ __rsyslog_viaq_prereq_packages + __rsyslog_viaq_packages }}"
```
sradco: @nhosoi does this work like this?

roles/rsyslog/roles/output_roles/elasticsearch/tasks/main.yaml
```diff
@@ -29,34 +29,35 @@
...
+        __rsyslog_elasticsearch_temp: >-
+          {{ __rsyslog_elasticsearch_temp | d([]) }} +
+          {{ [ item | combine( { 'ca_cert': item.ca_cert_src | basename ,
+          'cert': item.cert_src | basename, 'key': item.key_src | basename } ) ] }}
       with_items:
         - '{{ rsyslog_elasticsearch }}'
       when:
         - item.ca_cert_src is defined
-        - item.ca_cert_src != ""
+        - item.ca_cert_src
```
sradco: @nhosoi do they both mean the same
 item.ca_cert_src vs. item.ca_cert_src != ""   ?

roles/rsyslog/defaults/main.yaml
nhosoi: Thanks.  Fixed.

roles/rsyslog/roles/README.md
nhosoi: Thanks.  Fixed.

roles/rsyslog/roles/input_roles/basics/tasks/main.yaml
nhosoi: Basically, as long as one level list ['package0', 'package1', ...] is passed, flatten is not needed.  You might have noticed that the variables like `__rsyslog_basic_packages` start with `__` that means they are internal set in each vars/main.yaml files.  We do not allow users to set them.  So, it's guaranteed that the passed lists are one level.

But there is one exception. :)  We allow users to install/update extra rsyslog packages with rsyslog_extra_packages here.
https://github.com/linux-system-roles/logging/pull/87/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32R3
As you suggested, we'd better flatten rsyslog_extra_packages.  For instance, if this kind of nested list is given, it fails in the task.  I'm adding the flatten filter to rsyslog_extra_packages.  Thanks!!
rsyslog_extra_packages: [['rsyslog-mysql'], ['rsyslog-kafka']]

roles/rsyslog/roles/input_roles/files/tasks/cleanup.yaml
nhosoi: Well, we don't have a package cleanup code in cleanup.yaml. [1]  So, nothing happens even if we pass __rsyslog_packages...  Do you think we should uninstall packages?  If we uninstall rsyslog, it cleans up everything including the existing config files.  Is it preferable?  I checked linux-system-roles/{network,timesync,selinux}.  Network does not call "package"; the other 2 install packages with present, but I don't see absent...

[1] - https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/tasks/cleanup.yaml

roles/rsyslog/roles/input_roles/ovirt/tasks/main.yaml
nhosoi: Please see the comment https://github.com/linux-system-roles/logging/pull/87#discussion_r384707842

roles/rsyslog/roles/input_roles/viaq/tasks/main.yaml
nhosoi: Yes.  I'm adding a ci test using viaq next...

roles/rsyslog/roles/output_roles/elasticsearch/tasks/main.yaml
nhosoi: Yes.  This is part of adapting the ansible system roles best practices.

pcahyna: @nhosoi are you sure? Also which part of the best practices are you referring to?

pcahyna: see https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.8.html#bare-variables-in-conditionals

nhosoi: @sradco, @pcahyna, sorry, I obviously made a mistake.  Reverting this line.

roles/rsyslog/roles/input_roles/files/tasks/cleanup.yaml
sradco: @nhosoi No. You are right. I missed that this is the cleanup.yaml

roles/rsyslog/roles/output_roles/elasticsearch/tasks/main.yaml
pcahyna: @sradco I am afraid they do not mean the same - this is a complicated subject, see the discussion in https://github.com/ansible/ansible-lint/issues/457 and related issues. Bare variables used to be treated differently than expressions in conditionals. I.e. `something` is not the same as `not (not something)`, even if evaluated in a boolean context.

pcahyna: @nhosoi I think we should disable the problematic lint rule (E602), but there should be no problem with using `| length > 0` as well. Actually, I think now that `item.ca_cert_src` was actually correct, so @sradco was right - because it is not a bare variable - the dot in the middle makes it an expression. This is all so confusing :-(

roles/rsyslog/roles/input_roles/files/tasks/cleanup.yaml
pcahyna: FYI I don't think that we need an option to unistall the packages that we have installed. @nhosoi  As you observed, other roles don't do it either.

roles/rsyslog/roles/input_roles/basics/tasks/main.yaml
pcahyna: @nhosoi why should such a nested list appear there?

.travis.yml
```diff
@@ -10,7 +10,8 @@ env:
     - MOLECULE_DISTRO="fedora:30"
     - MOLECULE_DISTRO="fedora:31"
 install:
-  - pip install molecule docker
+  - pip uninstall molecule
```
pcahyna: @i386x @richm @tyll is the pip uninstall needed? Isn't `pip install 'molecule<3'` enough?

roles/rsyslog/defaults/main.yaml
```diff
@@ -7,6 +7,11 @@
 # Enable or disable ``rsyslog`` management.
 rsyslog_enabled: true
 
+# rsyslog_in_image
+#
+# true if the target host is a container and use rsyslog in the image.
+rsyslog_in_image: false
+
```
pcahyna: Is it a user-settable parameter? I don't see it in docs.

roles/rsyslog/tasks/deploy.yaml
```diff
@@ -1,20 +1,23 @@
...
   when:
-    - not use_rsyslog_image | default(false) |bool
+    - not rsyslog_in_image | default(false) | bool
```
pcahyna: why `default(false)` if you have already set the default in defaults?

roles/rsyslog/tasks/main.yaml
```diff
@@ -208,13 +215,13 @@
         state: stopped
       when:
         - not rsyslog_enabled | bool
-        - not use_rsyslog_image | default(false) | bool
+        - not rsyslog_in_image | default(false) | bool
 
-  when: rsyslog_failed_validation | d(false) == false
+  when: not __rsyslog_failed_validation | d(false) | bool
```
pcahyna: do you need the bool filter here? This filter converts strings to booleans. Is the value supposed to be a string?

roles/rsyslog/tasks/main.yaml
```diff
@@ -208,13 +215,13 @@
...
 
 - name: Check rsyslog version
   debug:
-    msg: "Rsyslog version must be >= 8.37.0-7.2.el7"
+    msg: "Rsyslog version must be >= 8.37.0-7.2"
   when:
-    - rsyslog_failed_validation | d(false) == true
+    - __rsyslog_failed_validation | d(false) | bool
```
pcahyna: same here.

roles/rsyslog/tasks/main.yaml
```diff
@@ -1,43 +1,34 @@
...
 
 - name: Get rsyslog version
-  package:
-    list: rsyslog
-  register: rsyslog_version
+  set_fact:
+    __rsyslog_version: >-
+      {{ ansible_facts.packages['rsyslog'][0].version
+      if ansible_facts.packages | selectattr('rsyslog', 'defined') | list | length > 0
+      else '0.0.0' }}
```
pcahyna: do you need set_fact or register? Can't you use vars: only on the block or tasks where the variables are actually used?

pcahyna: I would say it is using `ansible_facts.packages`. But my comment was not about the way to obtain the version, but about what is being done with the value. I don't think a fact or registered variable should be used.

roles/rsyslog/README.md
```diff
@@ -338,11 +338,11 @@ WARNING: Pre-existing rsyslog.conf and configuration files in /etc/rsyslog.d are
...
 Describing how the configuration files are defined to be deployed using the viaq case.
 
-Viaq configuration files are defined in {{ rsyslog_viaq_rules }} in /roles/input_rols/viaq/defaults/main.yaml.  The set is made from the generic modules{rsyslog_conf_global_options, rsyslog_conf_local_modules, rsyslog_conf_network_modules, rsyslog_conf_common} and viaq specific configurations.
+Viaq configuration files are defined in {{ __rsyslog_viaq_rules }} in /roles/input_roles/viaq/defaults/main.yaml.  The set is made from the generic modules{__rsyslog_conf_global_options, __rsyslog_conf_local_modules, __rsyslog_conf_common_defaults} and viaq specific configurations.
 
-To make a new configuration file installed in addition to the current {{ rsyslog_viaq_rules }}, create an rsyslog config item based on the following skelton and add the title {{ rsyslog_conf_yourname }} to {{ rsyslog_viaq_rules }}.
+To make a new configuration file installed in addition to the current {{ __rsyslog_viaq_rules }}, create an rsyslog config item based on the following skelton and add the title {{ __rsyslog_conf_yourname }} to {{ __rsyslog_viaq_rules }}.
 ```
-rsyslog_conf_yourname:
+__rsyslog_conf_yourname:
```
richm: Are these instructions intended for a role developer e.g. to add a new default configuration file to the role repo?  If so, then ok.

roles/rsyslog/defaults/main.yaml
richm: I think it is user-settable.  If `false` this means rsyslog is running on the host as a host package via systemd - the role needs to ensure that the rsyslog rpm packages are installed.  If `true`, this means rsyslog is running in a container - the role needs to ensure the right container is running - no rpm packages required.

roles/rsyslog/roles/input_roles/files/defaults/main.yaml
```diff
@@ -68,7 +68,7 @@ rsyslog_conf_imfile_inputs:
       - comment: 'Log messages from log files'
         options: |-
           # FIXME: Need to normalize a log before passing it to elasticsearch output
-          {% if rsyslog_elasticsearch is defined and rsyslog_elasticsearch|length > 0 %}
+          {% if rsyslog_elasticsearch is defined and rsyslog_elasticsearch | length > 0 %}
```
richm: if `rsyslog_elasticsearch` is a `list`, `dict`, or `string`, you do not need to test the length.  https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html and in particular https://jinja.palletsprojects.com/en/2.11.x/templates/#if - you can just use `if rsyslog_elasticsearch is defined and rsyslog_elasticsearch` - this means `if rsyslog_elasticsearch is defined and rsyslog_elasticsearch is not empty`

roles/rsyslog/roles/input_roles/basics/tasks/main.yaml
nhosoi: > @nhosoi why should such a nested list appear there?

@pcahyna, if we allow users to set a variable, any input could come in, so we'd better be prepared, I thought.

roles/rsyslog/roles/input_roles/files/tasks/cleanup.yaml
nhosoi: Thanks for your input, @pcahyna.

roles/rsyslog/roles/input_roles/viaq/tasks/main.yaml
richm: so we can use this instead of `flatten` (re: above comments regarding `flatten`)?

roles/rsyslog/defaults/main.yaml
nhosoi: @richm, do you think this parameter should be in the doc to advertise?  Please note that the efforts we made in the openshift logging for rsyslog at the last moment are not reflected to this logging role.  Do you think they should be?

roles/rsyslog/tasks/deploy.yaml
nhosoi: Fixed.

roles/rsyslog/defaults/main.yaml
richm: >  Do you think they should be?

Not sure, but let's worry about that in the future.  For now, for this PR, if `rsyslog_in_image` is settable by a user, and actual does something, let's doc it.

roles/rsyslog/tasks/main.yaml
nhosoi: @pcahyna, currently, __rsyslog_version is referred just in roles/rsyslog/tasks/main.yaml.  But I'm anticipating the value might be used in some other task files to absorb the version differences in the future.  What type of variable would be most appropriate for such a purpose?

roles/rsyslog/tasks/main.yaml
nhosoi: Fixed.

roles/rsyslog/README.md
nhosoi: Yes, I believe so.  I'm adding "to add a new default configuration file to the role repo" to the doc.  Thanks.

roles/rsyslog/tasks/main.yaml
richm: > What type of variable would be most appropriate for such a purpose?

I'm not sure, but let's do that in a separate commit in the future, when such functionality is required.

roles/rsyslog/defaults/main.yaml
nhosoi: Will do.  Thanks.

roles/rsyslog/tasks/main.yaml
nhosoi: Ok.  So, is it ok to keep this change in this PR or should I revert it?
https://github.com/linux-system-roles/logging/pull/87/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32L14-R17

richm: Hmm - for some reason that link isn't working for me - what is that change?

roles/rsyslog/roles/input_roles/viaq/tasks/main.yaml
nhosoi: Yes, I double-checked by running my test with viaq, which requires 2 package lists and they are combined at the line 5.
```
vars/main.yml:__rsyslog_viaq_prereq_packages: ['nmap-ncat', 'systemd-python', 'policycoreutils', 'checkpolicy', 'policycoreutils-python']
vars/main.yml:__rsyslog_viaq_packages: ['rsyslog-mmnormalize']
```
I experimentally removed rpm packages 'nmap-ncat' and 'rsyslog-mmnormalize' on the target host and ran ansible-playbook to see these 2 packages are installed as expected.

As long as `__rsyslog_.*_prereq_packages` and `__rsylog_.*_packages` are the clean 1 level list, there's no need to have `flatten`.  And lists starting with __ are internal and we manage them, so it's safe to assume they are clean. (if not, it is a bug.)  First I thought having `flatten` is a reasonable safeguard, but I was told to remove them unless they are really necessary, most likely not to hide a hidden bugs.

roles/rsyslog/tasks/main.yaml
nhosoi: Sorry, this is the raw diff.  (indeed, the link does not work for me, either ):
Actually, I realized there were 2 issues.  In the beginning, rsyslog version was retrieved using the yum module [1].  But we are to use `package` which is a higher level, platform dependent module.  Unfortunately, it does not return version.  So, I fixed it with the ansible_facts data [2].  That being said, we do not want to go back to `yum:`?

[1]
```
 - name: Get rsyslog version
-  yum:
+  package:
     list: rsyslog
   register: rsyslog_version
```
[2]
```
 - name: Get rsyslog version
-  package:
-    list: rsyslog
-  register: rsyslog_version
+  set_fact:
+    __rsyslog_version: >-
+      {{ ansible_facts.packages['rsyslog'][0].version
+      if ansible_facts.packages | selectattr('rsyslog', 'defined') | list | length > 0
+      else '0.0.0' }}
```

roles/rsyslog/roles/input_roles/viaq/tasks/main.yaml
richm: I think even with user supplied values for packages that are supposed to be in `list` format, it is ok to use them with `__default_packages + user_supplied_packages`.  If the user does not provide `user_supplied_packages` in `list` format, they will get an error, which IMO is ok for now. We can go back later and add some sort of input validation if we feel that the default error is not sufficient for some reason.  Also, any such public values should be listed in `defaults/main.yml` and have a default value of `[]`.

roles/rsyslog/tasks/main.yaml
richm: no, we do not want to go back to `yum` 
as pcahyna said, using `ansible_facts.packages['rsyslog'][0].version` is the canonical way to get the package version.

.travis.yml
tyll: the install command should be enough

roles/rsyslog/roles/input_roles/viaq/tasks/main.yaml
nhosoi: We already have a user supplied package list `rsyslog_extra_packages`.  It's initialized with [] in roles/rsyslog/defaults/main.yaml.
 roles/rsyslog/defaults/main.yaml
https://github.com/linux-system-roles/logging/pull/87/files#diff-def9bf9161e11520d46eea7aa85093e1R67-R70

And it's flattened before passing to the `package` module.  Hope it's ok...
https://github.com/linux-system-roles/logging/pull/87/files#diff-ed5a08a2e61430dc18e2d3820cbd6f32R3

.travis.yml
nhosoi: Thanks!  Removed uninstall.

richm:
lgtm - pcahyna?

roles/rsyslog/roles/input_roles/basics/tasks/main.yaml
pcahyna: @nhosoi I don't think we should expect any input, we should expect correct input. (Will we flatten all the lists passed to the role? If not, why is the package list special? If yes, will we do it in all the roles? We should as the roles should be consistent.)

nhosoi: @pcahyna, in my past experiences, I learned users could make any mistakes and ask supports for help; to reduce the calls, we are supposed to add any efforts as much as possible. :)  Let's assume a user put an invalid list.  I guess we need to choose one of these?
1. with flatten, logging role successfully works.
2. with some list inspection code (which does not exist now, I think), logging role fails with a descriptive error.
3. without any support, logging role fails with a cryptic error...
@richm, @sradco, can I have your thoughts?

richm: Having some sort of input normalization/validation for this parameter establishes a precedent, that we will have input normalization/validation for _all_ input parameters for _all_ system roles.  I don't think we want that.  I don't think we need that, except in very specific cases where:
* it is easy for users to make a mistake
* the mistake could be very costly (e.g. some sort of destructive operation that loses data)

I don't think this qualifies, so I would prefer not to have any extra code that makes this more complex.

pcahyna: I don't think we need a decision now. Please do not add new instances of `flatten`, but keep the existing ones. This is a topic for standards discussion.

pcahyna: @richm I agree. Moreover, input normalization for cases where the mistake could be very costly is even less justified, because it could convert user input to something that the user has not intended (OTOH input validation would be justified.)

roles/rsyslog/tasks/deploy.yaml
pcahyna: @nhosoi there are some remaining occurrences.

roles/rsyslog/roles/input_roles/basics/tasks/main.yaml
nhosoi: @pcahyna, @richm, thanks for your discussions.  Yes, it'd be nice to have some standards for handling possible user mistakes.

.travis.yml
pcahyna: @nhosoi please rebase to split the travis change into a separate commit and squash it with the fixup here: https://github.com/linux-system-roles/logging/pull/87/commits/07e3fa0d7317a29c62eb5d237c41623c0cb97766#diff-354f30a63fb0907d4ad57269548329e3

nhosoi: > @nhosoi please rebase to split the travis change into a separate commit and squash it with the fixup here: [07e3fa0#diff-354f30a63fb0907d4ad57269548329e3](https://github.com/linux-system-roles/logging/commit/07e3fa0d7317a29c62eb5d237c41623c0cb97766#diff-354f30a63fb0907d4ad57269548329e3)

@pcahyna, let me clarify your comment.  I was planning to squash 3 commits in this pr/#87 into one and merge.  Instead, are you suggesting to make 2 commits - one for travis changes and another for the rest?  Thanks.

pcahyna: @nhosoi sorry, I did not realize that you want to squash all the commits. But yes, I think that the Travis changes should be in a separate commits from the rest. By the way, I think other changes may also stay in separate commits, like renaming the variables and then adjusting the filters - @richm what do you think?

nhosoi: @pcahyna, it might be a bit difficult to make that many separate commits...  Travis is isolated, so it's easy.  For others, I'm afraid I may introduce some bugs (again...):  I don't trust myself. :p

pcahyna: @nhosoi what I meant, after splitting off the Travis changes, you could leave the three commits as they are otherwise, making 4 commits in total. @sradco @richm would you prefer the existing commits, or all squashed in one except the Travis changes? By the way, you should still address https://github.com/linux-system-roles/logging/pull/87/commits/69140401652cca448b476c923fe7535f5989ea57#r387825151

nhosoi: Ah, I see...  Thanks for the clarification, @pcahyna.
+1 to your suggestion.

richm: > @sradco @richm would you prefer the existing commits, or all squashed in one except the Travis changes?

It depends. If @sradco needs to cherry pick these changes to the stable branch, then fewer commits is better.  Otherwise, I think 4 commits is fine.

nhosoi: @richm, I don't think this pr is a candidate to backport to the stable branch except the travis/molecule change.  So, as long as the travis/molecule is isolated, we are good.

@sradco, do you want me to backport the travis fix to stable-0.1?

nhosoi:
Thank you all of you who reviewed this pr and answered to the questions.

Merging the pr.
Note: I created a separate commit from the travis/molecule change [3] and another commit addressing the reviews [2] (it used to be 2 commits which are merged into one [2]).  As shown in the commits page, 3 commits are involved.
https://github.com/linux-system-roles/logging/pull/87/commits
[1] Fixing package module related issues.
[2] Fixes to address the issues found in the review process.
[3] Workaround for the molecule3 upgrade - forcing to install molecule < 3.

roles/rsyslog/tasks/main.yaml
pcahyna: Here's a fix for a similar problem in timesync: https://github.com/linux-system-roles/timesync/pull/60

roles/rsyslog/tasks/main.yaml
pcahyna: By the way, do you really need `| length > 0` ?

nhosoi: > By the way, do you really need `| length > 0` ?

@pcahyna, no, it's not needed.

roles/rsyslog/tasks/main.yaml
nhosoi: Thank you so much, @pcahyna.  I tested this diff and verified it works as expected.
https://github.com/nhosoi/logging/compare/best_practices...nhosoi:package_version

Once pr/89 is done, I'm opening a pr with it.
https://github.com/linux-system-roles/logging/pull/89

pcahyna: @nhosoi I did not intend to say that you must follow the patch taken in timesync. Rather we should determine which approach is better and use it consistently.
(yet another approach would be to not support check mode on first run of the role until Ansible 2.7 is deprecated and we can solve the problem much more elegantly using the default filter).

roles/rsyslog/tasks/main.yaml
nhosoi: @pchayna
> By the way, do you really need `| length > 0` ?

I've opened a pr to remove `| length > 0`.  Could you please take a look?
https://github.com/linux-system-roles/logging/pull/91
Also, it turned out selectattr did not work for this case [1], thus I replaced it with json_query.  But if you think I should use the timesync method https://github.com/linux-system-roles/timesync/pull/60, please add your comment to https://github.com/linux-system-roles/logging/pull/91.  I'm going to update the pr.  Thanks.

[1] - Debug outputs from `ansible_facts.packages | json_query('rsyslog')` and `ansible_facts.packages | selectattr('rsyslog', 'defined') | list`.
```
TASK [linux-system-roles/logging/roles/rsyslog : debug] *******************************************
task path: linux-system-roles/logging/roles/rsyslog/tasks/main.yaml:12
ok: [momo7.localdomain] => {
    "msg": "json_query - [{'name': 'rsyslog', 'source': 'rpm', 'epoch': None, 'version': '8.24.0', 'release': '41.el7_7.2', 'arch': 'x86_64'}]"
}

TASK [linux-system-roles/logging/roles/rsyslog : debug] *******************************************
task path: linux-system-roles/logging/roles/rsyslog/tasks/main.yaml:15
ok: [momo7.localdomain] => {
    "msg": "selectattr - []"
}
```

roles/rsyslog/roles/output_roles/elasticsearch/tasks/main.yaml
pcahyna: @nhosoi by the way, I just disabled the lint rule 602, which disallows comparisons to empty string, globally in the template. https://github.com/linux-system-roles/template/pull/19 . You can safely use comparison to empty string when you believe it is the best option. (After this change is merged.)

nhosoi: Thank you for the update, @pcahyna !

roles/rsyslog/tasks/main.yaml
pcahyna: maybe I already said this but timesync now uses an even simpler approach: https://github.com/linux-system-roles/timesync/pull/67/files#diff-6d62b73b3f46cfa11f6466d08cf45303

PR #89, opened 2020-03-05
Following Ansible Best Practices 2.

nhosoi:
- Renaming logging_collector to logging_provider (== $ROLENAME_provider).
- In rules.conf.j2, replacing double quotes with single quotes.
- Changing file extension ".yaml" to ".yml".
- Changing rsyslog_elasticsearch to the internal variable __rsyslog_elasticsearch,
  which is a dictionary and the key-value pairs are from the elasticsearch outputs role.
- Changing rsyslog_enabled to the internal variable __rsyslog_enabled which value
  comes from logging_enabled.
- Adding and fixing comments and docs.
- Removing unnecessary bool filter from __rsyslog_enabled check.
- Removing unused variable logging_custom_config_files from  defaults/main.yml.
- Making the __rsyslog_enabled default value consistent.


tasks/main.yml
```diff
@@ -1,15 +1,14 @@
...
+- name: Set Rsyslog facts then include rsyslog role
+  block:
 
-- block:
-
-    - name: Set rsyslog_enabled
+    - name: Set __rsyslog_enabled
       set_fact:
-        rsyslog_enabled: "{{ logging_enabled | d(true) }}"
+        __rsyslog_enabled: "{{ logging_enabled }}"
```
pcahyna: if you remove the `| bool` filter on all places where `__rsyslog_enabled` is being used, you should add it here, as logging_enabled is user-supplied and not guaranteed to be a boolean - it can be a boolean-like string.

richm: @pcahyna that means we need to do input normalization on some of the input parameters?

pcahyna: For booleans this seems to be the usual practice, yes, but I don't have a reference for this. Again, to be discussed.

roles/rsyslog/tasks/main.yml
```diff
@@ -2,8 +2,7 @@
   package:
     name: "{{ (__rsyslog_base_packages) + (__rsyslog_tls_packages if rsyslog_pki else []) + (rsyslog_extra_packages | flatten) }}"
     state: latest
-  when:
-    - rsyslog_enabled | bool
+  when: __rsyslog_enabled
```
pcahyna: doesn't this lead to a warning about bare variable?

richm: yes - it does - so I guess the moral of the story is that you cannot use a bare variable in a `when` - if it is a boolean, you can use `when: varname | bool` - if it is some other type, I'm not sure what to do.

I created a test playbook
```yaml
---
- name: set parameters for test
  hosts: localhost
  tasks:
  - set_fact:
      junkdict: {'foo':'bar'}

  - debug:
      msg: hereherehere
    when: junkdict
```
running this - `ansible-playbook junk-playbook.yml` - gives
```
[DEPRECATION WARNING]: evaluating {'foo': 'bar'} as a bare variable, this 
behaviour will go away and you might need to add |bool to the expression in the
 future. Also see CONDITIONAL_BARE_VARS configuration toggle.. This feature 
will be removed in version 2.12. Deprecation warnings can be disabled by 
setting deprecation_warnings=False in ansible.cfg.
ok: [localhost] => {
    "msg": "hereherehere"
}
```
so it "works" - that is, it correctly evaluates the `dict` value as non-empty - but we get the warning - so I tried using `when: junkdict|bool` and I get this:
```
skipping: [localhost]
```
That doesn't work - if you filter a dict as a bool, it does not return a `true` value.  I tried this too - `when: "{{ junkdict }}"`:
```
[WARNING]: conditional statements should not include jinja2 templating
delimiters such as {{ }} or {% %}. Found: {{ junkdict }}
ok: [localhost] => {
    "msg": "hereherehere"
}
```
it also "works" but with a warning.  The only thing I have found that works consistently without a warning is this: `when: junkdict | length > 0` - but that only works with types that have a length (string, dict, list) - also note that with types of `int`, you can use `when: var|bool` but note that `0` evaluates to `false`, and *all other values* evaluate to `true` which may or may not be what you want

pcahyna: @richm the problem is, `| bool` is not a conversion to boolean in the Python sense. It is a conversion of string like `"yes"` -> true, `"no"` -> false. For a conversion in the Python sense this you would just use the variable and Jinja2 would convert it for you, except that in case it is just a variable and not a more complex expression you hit the "bare variable" rule (which is fortunately being deprecated). Plenty of discussions of this can be found in related Ansible issues. Didn't I say that this is confusing?

nhosoi: > doesn't this lead to a warning about bare variable? Also for the cases below.

@pcahyna, do we need to wrap variables in the when conditional?
https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html
According to this user guide,
`As the examples show, you don’t need to use {{ }} to use variables inside conditionals, as these are already implied.`
In our case, __rsyslog_enabled is considered "already implied" or not?  Anyway, I don't see any warning messages in terms of __rsyslog_enabled.

pcahyna: I think you can work around by using one of these which should be equivalent and not behave like bare variables:
```
not (not junkdict)
junkdict | d()
( junkdict )
```
I have not tested them though

pcahyna: or
```
vars.junkdict
```

pcahyna: > > doesn't this lead to a warning about bare variable? Also for the cases below.
> 
> @pcahyna, do we need to wrap variables in the when conditional?
> https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html

No.

> According to this user guide,
> `As the examples show, you don’t need to use {{ }} to use variables inside conditionals, as these are already implied.`
> In our case, __rsyslog_enabled is considered "already implied" or not? Anyway, I don't see any warning messages in terms of __rsyslog_enabled.

@richm says that it does lead to warning messages.

pcahyna: Anyway, I would say that we should not do this change yet, because it is not clear what the best practice is.

pcahyna: related discussion: https://github.com/linux-system-roles/logging/pull/87#pullrequestreview-364901931

pcahyna:
@nhosoi I would say this PR is again doing too much. Can you please do just the mechanical changes (like renaming of files and changes of the quoting style)?

roles/rsyslog/tasks/main.yml
nhosoi: > Anyway, I would say that we should not do this change yet, because it is not clear what the best practice is.

Ok, I'm going to put the bool filter back.  Thanks.

nhosoi:
> @nhosoi I would say this PR is again doing too much. Can you please do just the mechanical changes (like renaming of files and changes of the quoting style)?

@pcahyna, I'd love to do so...  That was my original intention, but the pr gets more and more complicated as usual... :(  Currently, there are 3 commits in this pr.  Is it ok to squash them into one?  Or do you want me to reshuffle the commits?

pcahyna:
@nhosoi , to the contrary, I think the first commit should be split to isolate the truly mechanical changes (file renames and perhaps quoting style changes) - other changes like variable renaming or adding variables to vars/defaults or adding/removing Jinja filters should be split into separate commits and kept on a branch, together with the two remaining commits.

nhosoi:
> @nhosoi , to the contrary, I think the first commit should be split to isolate the truly mechanical changes (file renames and perhaps quoting style changes) - other changes like variable renaming or adding variables to vars/defaults or adding/removing Jinja filters should be split into separate commits and kept on a branch, together with the two remaining commits.

Ok...  When I try to implement something defined in the logging roles tasks, issues due to my violating the best practices always come up and it stops me working on the tasks themselves...  So, I am hoping to have a solid ground asap...  That being said, in the list https://github.com/linux-system-roles/logging/pull/89#issue-384504681, do you recommend to retrieve just following 2 and revert the rest?  Unfortunately, it's going to be manual changes... :(
- In rules.conf.j2, replacing double quotes with single quotes.
- Changing file extension ".yaml" to ".yml".

nhosoi:
> @nhosoi , to the contrary, I think the first commit should be split to isolate the truly mechanical changes (file renames and perhaps quoting style changes) - other changes like variable renaming or adding variables to vars/defaults or adding/removing Jinja filters should be split into separate commits and kept on a branch, together with the two remaining commits.

@pcahyna, I've opened https://github.com/linux-system-roles/logging/pull/90, which is a subset of this pr/89 and contains just these items. 
- In rules.conf.j2, replacing double quotes with single quotes.
- Changing file extension ".yaml" to ".yml".
- Removing unused variable logging_custom_config_files from defaults/main.yml.
- Fixing docs in design_docs/*.md and README.md

I'm closing this pr in favor of pr/90.  In terms of the leftover of this pr (pr/89 ^ pr/90), what should I do?  Shall I leave them as is?

pcahyna:
for the remaining changes, can you leave them on a branch for now?

roles/rsyslog/roles/output_roles/elasticsearch/tasks/main.yml
```diff
@@ -38,19 +38,19 @@
...
+    - name: Set updated __rsyslog_elasticsearch
       set_fact:
-        rsyslog_elasticsearch: "{{ __rsyslog_elasticsearch_temp }}"
+        __rsyslog_elasticsearch: "{{ __rsyslog_elasticsearch_temp }}"
 
   when:
-    - rsyslog_elasticsearch is defined
-    - rsyslog_elasticsearch != []
+    - __rsyslog_elasticsearch is defined
+    - __rsyslog_elasticsearch != []
```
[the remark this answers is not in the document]
pcahyna: @richm are you sure?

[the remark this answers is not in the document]
richm: no, not sure - but it should be pretty easy to verify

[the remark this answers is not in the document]
pcahyna: I mean in the context of the bare variable discussion below. Your suggestion to change from `__rsyslog_elasticsearch != []` to just use `__rsyslog_elasticsearch` makes the expression a bare variable (if it was in when: - the commit is not available anymore) which is prone to change the meaning.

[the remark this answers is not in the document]
richm: Then maybe we should just use `__rsyslog_elasticsearch | d([])` for all of these sorts of tests?  This avoids the bare variable problem, and the need for `is defined`, and any tests involving `!=`

pcahyna:
@nhosoi  what is the name of the branch with the remaining changes?

pcahyna:
^ @nhosoi

nhosoi:
> ^ @nhosoi

Hi @pcahyna, this is the [branch](https://github.com/nhosoi/logging/tree/best_practices) associated with this PR.

I rescanned the diff and verified there are still 3 best practice issues left.
1. `rsyslog` should be set to `logging_provider` instead of `logging_collector`.
```
- # Default logging collector is Rsyslog
- logging_collector: rsyslog
+ # Default logging provider is Rsyslog
+ logging_provider: rsyslog
```
2. Internal variables should be started with `__` as follows.
```
-  when: rsyslog_enabled | bool
+  when: __rsyslog_enabled | bool
```
3. There could be some ansible vars precedence violations. (e.g., put vars which could be reset in the vars/*.yml files mistakenly)  I need to double check them.

I think that covers my current best practice related issues, but I'd appreciate if you could add any to the list.  I'm going to open a ticket for them with the above 3 items.  Thanks!

PR #91, opened 2020-03-07
Removing unnecessary "length > 0" checks.

nhosoi:
Replacing selectattr with json_query in __rsyslog_version retrieval.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ install:
   # workaround until https://github.com/ansible/molecule/issues/1727 is fixed
   - pip uninstall -y testinfra
   - pip install testinfra
+  - pip install jmespath
 script:
   - molecule test
 branches:
--- a/tests/tests_enabled.yml
+++ b/tests/tests_enabled.yml
@@ -20,3 +20,14 @@
     - name: Check file counts in rsyslog.d
       assert:
         that: rsyslog_d_file_count.matched < 3
+
+    - name: Get rsyslog_version_str
+      command: /bin/grep "^# rsyslog version is " /etc/rsyslog.conf
+      register: rsyslog_version_str
+      changed_when: false
+
+    - name: Check rsyslog_version
+      assert:
+        that: rsyslog_version | d() and rsyslog_version != "0.0.0"
+      vars:
+        rsyslog_version: "{{ rsyslog_version_str.stdout | regex_search('[0-9]*\\.[0-9]*\\.[0-9]*') }}"
--- a/roles/rsyslog/roles/output_roles/files/defaults/main.yml
+++ b/roles/rsyslog/roles/output_roles/files/defaults/main.yml
@@ -39,7 +39,7 @@ __rsyslog_conf_files_output_rules:
     sections:
 
       - options: |-
-          {% if rsyslog_files_actions is defined and rsyslog_files_actions %}
+          {% if rsyslog_files_actions | d([]) %}
           {% for element in rsyslog_files_actions %}
           {% if element.name is defined and element.path is defined %}
           {% if element.severity is defined %}
--- a/roles/rsyslog/roles/output_roles/files/defaults/main.yml
+++ b/roles/rsyslog/roles/output_roles/files/defaults/main.yml
@@ -39,7 +39,7 @@ __rsyslog_conf_files_output_rules:
     sections:
 
       - options: |-
-          {% if rsyslog_files_actions is defined and rsyslog_files_actions | length > 0 %}
+          {% if rsyslog_files_actions is defined and rsyslog_files_actions %}
           {% for element in rsyslog_files_actions %}
           {% if element.name is defined and element.path is defined %}
           {% if element.severity is defined %}
--- a/roles/rsyslog/tasks/main.yml
+++ b/roles/rsyslog/tasks/main.yml
@@ -13,7 +13,7 @@
   set_fact:
     __rsyslog_version: >-
       {{ ansible_facts.packages['rsyslog'][0].version
-      if ansible_facts.packages | selectattr('rsyslog', 'defined') | list | length > 0
+      if ansible_facts.packages | json_query('rsyslog')
       else '0.0.0' }}
 
 - debug:
--- a/roles/rsyslog/tasks/main.yml
+++ b/roles/rsyslog/tasks/main.yml
@@ -128,7 +128,7 @@
       loop: '{{ __rsyslog_common_rules | flatten }}'
       when:
         - rsyslog_enabled | bool
-        - rsyslog_logs_collections is defined and rsyslog_logs_collections != []
+        - rsyslog_logs_collections | d([])
         - item.filename | d() or item.name | d()
         - item.state is undefined or item.state != 'absent'
         - item.options | d() or item.sections | d()
@@ -173,7 +173,7 @@
         loop_var: input_item
       when:
         - rsyslog_enabled | bool
-        - input_item is defined and input_item != []
+        - input_item | d([])
         - input_item.state | d('present') != 'absent'
 
     - name: Run remove input sub-roles configs
@@ -184,7 +184,7 @@
       loop_control:
         loop_var: input_item
       when:
-        - input_item is defined and input_item != []
+        - input_item | d([])
         - input_item.state | d('present') == 'absent'
 
     - name: Run output sub-roles configs
@@ -195,7 +195,7 @@
         loop_var: output_item
       when:
         - rsyslog_enabled | bool
-        - output_item is defined and output_item != []
+        - output_item | d([])
       tags:
         - skip_ansible_lint
 
--- a/roles/rsyslog/templates/etc/rsyslog.conf.j2
+++ b/roles/rsyslog/templates/etc/rsyslog.conf.j2
@@ -1,5 +1,5 @@
 # {{ ansible_managed }}
-
+# rsyslog version is {{ __rsyslog_version }}.
 {% if not rsyslog_default | bool %}
 #
 # Include all config files in {{ rsyslog_config_dir }}


.travis.yml
```diff
@@ -14,6 +14,7 @@ install:
   # workaround until https://github.com/ansible/molecule/issues/1727 is fixed
   - pip uninstall -y testinfra
   - pip install testinfra
+  - pip install jmespath
```
richm: is there an rpm based package that provides this functionality?  We can't use pip installed packages in our products.

nhosoi: This "pip install jmespath" addition is just to make the travis CI test back to work.  The other CI tests are just fine (except rhel/centos-6 :) without it.  Should we still not use pip in .travis.yml?  I'm not sure how the CI test hosts are configured, but it looks to me jmespath is missing only in the host for the travis CI and it is available in the other test hosts...  Which is the valid initial status?

In terms of  the rpm packages, there are 2 of them:
```
python2-jmespath.noarch : JSON Matching Expressions
python3-jmespath.noarch : JSON Matching Expressions
```

pcahyna: @richm yes, there is a jmespath package and the rhel-system-roles package requires it in order to work - it is properly expressed as an RPM dependency. This is not related to Travis though - it does not even use a RPM distribution.

roles/rsyslog/roles/output_roles/files/defaults/main.yaml
```diff
@@ -39,7 +39,7 @@ __rsyslog_conf_files_output_rules:
     sections:
 
       - options: |-
-          {% if rsyslog_files_actions is defined and rsyslog_files_actions | length > 0 %}
+          {% if rsyslog_files_actions is defined and rsyslog_files_actions %}
```
pcahyna: would
```
rsyslog_files_actions | d([])
```
work? @richm what do you think?

roles/rsyslog/tasks/main.yaml
```diff
@@ -13,7 +13,7 @@
   set_fact:
     __rsyslog_version: >-
       {{ ansible_facts.packages['rsyslog'][0].version
-      if ansible_facts.packages | selectattr('rsyslog', 'defined') | list | length > 0
+      if ansible_facts.packages | json_query('rsyslog')
```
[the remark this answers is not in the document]
nhosoi: > is there a test for the problem being fixed here? I am surprised that the problem was not caught before merging #87

I will add a test case.

roles/rsyslog/roles/output_roles/files/defaults/main.yaml
nhosoi: Fixed.   I also fixed similar codes in roles/rsyslog/tasks/main.yml.

tests/tests_enabled.yml
```diff
@@ -20,3 +20,7 @@
     - name: Check file counts in rsyslog.d
       assert:
         that: rsyslog_d_file_count.matched < 3
+
+    - name: Check __rsyslog_version is set
+      assert:
+        that: __rsyslog_version != "0.0.0"
```
pcahyna: what does this change have in common with the rest of the commit?

roles/rsyslog/tasks/main.yaml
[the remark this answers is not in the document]
pcahyna: It does not seem like a good way to test this. It is an internal variable; what if it gets renamed or its scope restricted? This way you are basically relying on the unfortunate fact that variables leak out from the role. Can't you rather test for the actual functionality that needs this variable? I mean, before you fixed the problem, the role must have behaved differently in some situation, otherwise the whole code would be useless.

tests/tests_enabled.yml
nhosoi: > what does this change have in common with the rest of the commit?

Note: This change is a fix for "is there a test for the problem being fixed here? I am surprised that the problem was not caught before merging #87".  And the discussion is on-going in the thread https://github.com/linux-system-roles/logging/pull/91#discussion_r389896570.

roles/rsyslog/tasks/main.yaml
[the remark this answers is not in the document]
nhosoi: @pcahyna, currently, the version info is used only when rsyslog_in_image is true, which is not supported in the test-harness, I'd think.  Do you have any recommendation to do in such a case?  As you described above, the assertion to check `__rsyslog_version != "0.0.0"` is not ideal, but probably we could detect an error if some mismatch was introduced in a new pull request?

nhosoi:
@pcahyna, BTW, there are 2 commits in this pr for now.  Do we want to make more smaller commits from them or is it ok with 2 commits?  (Or do we want to open another pr with a part of them?)

https://github.com/linux-system-roles/logging/pull/91/commits
[1]
```
Removing unnecessary "length > 0" checks.
Replacing selectattr with json_query in __rsyslog_version retrieval.
pip install jmespath in .travis.yml to use json_query.
```
[2]
```
Simplifying jinja2 conditional expression:
      old> variable is defined and variable != []
      new> variable | d([])
Adding a test case to check __rsyslog_version is set properly.
```

tests/tests_enabled.yml
pcahyna: I know, but what does this new test have in common with removing length filters from `rsyslog_files_actions` etc ? I.e. why to have a commit
```
 Simplifying jinja2 conditional expression:

  old> variable is defined and variable != []
  new> variable | d([])

Adding a test case to check __rsyslog_version is set properly.
```
when the "Adding a test case to check __rsyslog_version is set properly." part has nothing to do with the "Simplifying jinja2 conditional expression:" part? Please split the commit in two. (You can keep both in this PR.)

pcahyna:
@nhosoi I would prefer to split [2] in two smaller commits and I don't think a separate PR is justified at this time.

tests/tests_enabled.yml
nhosoi: Understood your question.  This change was made before the agreement made in the meeting.  Instead of fixing the issues raised in the review process in the same pr, I will open a new pr for them.  Actually, that was my question https://github.com/linux-system-roles/logging/pull/91#issuecomment-597248067 :)

nhosoi:
> @nhosoi I would prefer to split [2] in two smaller commits and I don't think a separate PR is justified at this time.

Thanks, @pcahyna.  I dropped the test case change.

tests/tests_enabled.yml
pcahyna: But the issue of not having a test for the change in this PR is related to this PR, so please add it to this PR. How would you test that the change is correct otherwise?

roles/rsyslog/tasks/main.yaml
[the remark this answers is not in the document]
pcahyna: Ah. So there is even a test missing for `rsyslog_in_image`, so there is a bigger problem.

tests/tests_enabled.yml
pcahyna: We need to reach an agreement first on what is actually a good test, though.

tests/tests_enabled.yml
```diff
@@ -20,3 +20,16 @@
...
         that: rsyslog_d_file_count.matched < 3
+
+    - name: Get rsyslog_version_str
+      command: /bin/grep "^# rsyslog version is " /etc/rsyslog.conf
+      register: rsyslog_version_str
+      changed_when: false
+
+    - name: Retrieve rsyslog_Version
+      set_fact:
+        rsyslog_version: "{{ rsyslog_version_str.stdout | regex_search('[0-9]*\\.[0-9]*\\.[0-9]*') }}"
```
pcahyna: Why do you need to set a fact? Can't you just set a variable on the task which needs it, i.e. the task below?

pcahyna: Note: it is of course not extremely important to have this done properly in a test playbook, but I would still prefer it to set a good example.

richm: > set a variable on the task which needs it, i.e. the task below?

Not sure what you mean

pcahyna:
```yaml
    - name: Check rsyslog_version
      assert:
        that: rsyslog_version | d() and rsyslog_version != "0.0.0"
      vars:
        rsyslog_version: "{{ rsyslog_version_str.stdout | regex_search('[0-9]*\\.[0-9]*\\.[0-9]*') }}"
```

richm: cool - I didn't know about that feature

nhosoi: Thank you, @pcahyna and @richm.  I pushed the fix following the suggestion by Pavel.

pcahyna:
I propose to merge this with squashing the last two commits together. But I don't like the way how the version determination is tested and that the version is being added to the configuration file just to make the change testable. I hope it will be possible to improve the test.

nhosoi:
> I propose to merge this with squashing the last two commits together. But I don't like the way how the version determination is tested and that the version is being added to the configuration file just to make the change testable. I hope it will be possible to improve the test.

Indeed, I added the version string to the main rsyslog config file for checking it in the test.  But I'm now thinking it might be somehow more useful than I first thought...  The rsyslog version is the one when the logging role is executed.  Do we care the situation for the administrator to run "yum update" on the target hosts and the version of the system (rsyslog in this case) is more advanced than the one for the configuration?  I'd think the backward compatibility is pretty much guaranteed, so any sudden stop-working event won't occur.  But, getting deprecated warning level event might start at some point.  At that time, we may find the information useful for troubleshooting...

pcahyna:
> Putting my comment aside, shall I open an issue for "the version is being added to the configuration file" just for testing not to forget?

Yes please. Note though that it is not just about the version added to the configuration file. The problem is in general that it is not a functional test, i.e. it does not test that the functionality for which you need the version test is working properly.

By the way, does your test fail in the version before your fix?

nhosoi:
> Yes please.

All right.  Will do.

> Note though that it is not just about the version added to the configuration file. The problem is in general that it is not a functional test, i.e. it does not test that the functionality for which you need the version test is working properly.

In this case, the issue is lacking the test platform which pulls rsyslog from the container image.  (note: in the logging role code, it is in the rsyslog_in_image block.)  The part of the logging role was planed to use in the openshift logging and tested and verified it worked when it was implemented.  But rsyslog was dropped from the openshift logging.  Plus, the LSR logging role did not get used even before the drop.

> By the way, does your test fail in the version before your fix?

It was working in the version tagged with 0.1.0.  But I did not have a chance to test the code with rsyslog_in_image enabled after the fix -- instead of using yum, using package to install `rsyslog`, then the version info needed to be retrieved independently -- was made.

nhosoi:
Thank you for reviewing the pr, @pcahyna.

Merging...

PR #94, opened 2020-03-12
Create a configuration directory if it does not exist.

nhosoi:
Description: Create a configuration directory (rsyslog_config_dir) if it does not exist.

### DIFF
--- a/roles/rsyslog/tasks/main.yml
+++ b/roles/rsyslog/tasks/main.yml
@@ -54,6 +54,14 @@
         - rsyslog_unprivileged | bool
         - rsyslog_user != 'root'
 
+    - name: Create the config directory if it does not exist
+      file:
+        state: directory
+        path: '{{ rsyslog_config_dir }}'
+        owner: '{{ rsyslog_user }}'
+        group: '{{ rsyslog_file_group }}'
+        mode: 0755
+
     # Creating a work directory
     - name: Create a work directory
       file:


roles/rsyslog/tasks/main.yml
```diff
@@ -54,6 +54,22 @@
         - rsyslog_unprivileged | bool
         - rsyslog_user != 'root'
 
+    # Checking if the config directory exists or not
+    - name: Get config directory stat
+      stat:
+        path: "{{ rsyslog_config_dir }}"
+      register: stat_rsyslog_config_dir
```
richm: This part should not be necessary - the `file` module should only create the directory if it does not exist: https://docs.ansible.com/ansible/latest/modules/file_module.html#file-module
```yaml
- name: Create a directory if it does not exist
  file:
    path: /etc/some_directory
    state: directory
    mode: '0755'
```

richm: and, fyi - even though you are getting rid of this - the variable you created with `register` should be named appropriately according to the best practices - https://github.com/oasis-roles/meta_standards/pull/12/files#diff-04c6e90faac2675aa89e2176d2eec7d8R81 (soon to be merged) - so something like `register: __rsyslog_register_config_dir`

nhosoi: > This part should not be necessary - the `file` module should only create the directory if it does not exist: https://docs.ansible.com/ansible/latest/modules/file_module.html#file-module
> 
> ```yaml
> - name: Create a directory if it does not exist
>   file:
>     path: /etc/some_directory
>     state: directory
>     mode: '0755'
> ```

Thank you, @richm.  I removed the unnecessary stat task.

roles/rsyslog/tasks/main.yml
```diff
@@ -54,6 +54,15 @@
...
         - rsyslog_user != 'root'
 
+    - name: Create the config directory if it does not exist
+      file:
+        state: directory
+        path: '{{ rsyslog_config_dir }}'
+        owner: '{{ rsyslog_user }}'
+        group: '{{ rsyslog_file_group }}'
+        mode: 0755
+      changed_when: false
```
richm: Why use `changed_when: false` here?

nhosoi: A good question...  I copied an existing directory creation task.  I found this comment [1].  I guess we should consider "change" any changes made to the host, not just to the rsyslog configuration?  If so, should we remove changed_when in the file/directory creation tasks?  (I remove this "changed_when: false" in this PR with opening another for the rest.)
[1]
```
commit 85253790ea38f04829e68255353c0335d8165e27
Author: Shirly Radco <sradco@redhat.com>
Date:   Mon Jan 28 16:44:52 2019 +0200
    Updated backup task - added changed_when as False,
    Since the task does not actually changes Rsyslog configurations.
    Signed Of By: Shirly Radco
```

richm: Hmm - not sure what Shirly was doing there - but in general, you want to know when something changed - you want to know if the directory was created

Also, if you used `changed_when: false` in other places, yes, we should review them, in another PR

nhosoi: Ah, she put `changed_when: false` when creating a work dir `/var/lib/rsyslog` and a back up dir `/tmp/rsyslog.d-DATETIME` since they are not /etc/rsyslog.conf nor /etc/rsyslog.d related.  Probably, we could get rid of them?

nhosoi:
Thank you for your reviews, @richm!

roles/rsyslog/tasks/main.yml
richm: >  Probably, we could get rid of them?

yes

PR #95, opened 2020-03-13
Removing unnecessary rsyslog_unprivileged, rsyslog_user, rsyslog_group, rsyslog_allow variables.

nhosoi:
- rsyslog_unprivileged, rsyslog_user and rsyslog_group are prepared to support non-root ownership of rsyslog, which is not officially supported in RHEL. Thus dropping the implementation.
- rsyslog_allow and its siblings are prepared, but not being used.
- Removing the 'Set default permissions for all log files' rule.
   Note: without ownership and mode changes, the following settings are made.
        -r--------. 1 root root 134 Mar 15 15:48 rsyslog.conf
        drwxr-xr-x. 2 root root 277 Mar 15 15:48 rsyslog.d
        -r--------. 1 root root 118 Mar 15 15:48 rsyslog.d/##-filename.conf


richm:
What is the motivation for this change?  Is this yet another case where we are porting something from Shirly's code that she required for some special purpose?

nhosoi:
> What is the motivation for this change? Is this yet another case where we are porting something from Shirly's code that she required for some special purpose?

Well, my current motivation is adding the code to eliminate rsyslog_default (copying the original rsyslog.conf to /etc).  I already have a quite big patch for that and am cutting a small chunk out from it and opening a pr one by one.  This particular patch is not directly related to it.  But in my effort I noticed the functionality (using non-root owner for rsyslogd) is not tested, so I added it.  In that sense, yes, it's a side story.

richm:
I don't think we support changing the user for rsyslog on fedora/centos/rhel.  I suppose this is a holdover from the debops rsyslog role.  I would rather not have this in there at all.  None of the other roles that start system services support changing the user.  Usually that is controlled by the rpm package and we don't support changing it post install.  debian is quite a bit different than fedora/centos/rhel in this regard.

nhosoi:
> I don't think we support changing the user for rsyslog on fedora/centos/rhel. I suppose this is a holdover from the debops rsyslog role. I would rather not have this in there at all. None of the other roles that start system services support changing the user. Usually that is controlled by the rpm package and we don't support changing it post install. debian is quite a bit different than fedora/centos/rhel in this regard.

I thought the same thing and did a quick search to find such a requirement exists...  Do you think we can ignore it?
https://access.redhat.com/solutions/3710001
https://www.thegeekdiary.com/how-to-run-rsyslog-as-a-non-root-user-in-centos-rhel-7/

richm:
from https://access.redhat.com/solutions/3710001
"Please note: The following information has been provided by Red Hat, but is outside the scope of support."

so no, I don't think we want to support this.

nhosoi:
> from https://access.redhat.com/solutions/3710001
> "Please note: The following information has been provided by Red Hat, but is outside the scope of support."
> 
> so no, I don't think we want to support this.

Thanks.  Let me turn this pr to eliminating logging_unprivileged. :)

nhosoi:
@richm, I removed rsyslog_unprivileged, rsyslog_user, rsyslog_group and some other unused variables.  Could you please take a look at this pr one more time?  Thanks.

roles/rsyslog/defaults/main.yml
```diff
@@ -359,15 +301,11 @@ __rsyslog_conf_common_defaults:
 
       - comment: 'Set default permissions for all log files'
         options: |-
-          $FileOwner {{ rsyslog_file_owner }}
-          $FileGroup {{ rsyslog_file_group }}
+          $FileOwner root
+          $FileGroup root
```
richm: Are these parameters required?  If not, I think they can be omitted.

roles/rsyslog/defaults/main.yml
```diff
@@ -221,28 +185,6 @@ rsyslog_tcp_port: '514'
 # The incoming TCP TLS port used for remote logging.
 rsyslog_tcp_tls_port: '6514'
 
-# rsyslog_allow
```
richm: Did you mean to get rid of these?

roles/rsyslog/defaults/main.yml
nhosoi: Nope.  The default is root.  Removing them.  Thanks!
```
./tools/omfile.c:	{ "fileowner", eCmdHdlrUID, 0 },
./plugins/imptcp/imptcp.c:	{ "fileowner", eCmdHdlrUID, 0 },
./tools/omfile.c:	{ "filegroup", eCmdHdlrGID, 0 },
./plugins/imptcp/imptcp.c:	{ "filegroup", eCmdHdlrGID, 0 },
```

roles/rsyslog/defaults/main.yml
nhosoi: Yes, these rsyslog_.*allow variables are from debops and they are not used in the LSR logging for now.

roles/rsyslog/defaults/main.yml
nhosoi: Besides, they are all obsolete legacy directives for omfile...

`$FileOwner` - `process user`, by default
`$FileGroup` - `process user's primary group`, by default
`$FileCreateMode` - `0644`, by default
`$DirCreateMode` - 0700, by default
`$Umask` - rsyslogd's process umask; if in doubt, use `$umask 0000`

I'd think it's guaranteed the process user is root with the linux system role deployment.  And we want to set the default permission 0700 to /etc/rsyslog.d, don't we?  Lastly, can we expect root's umask is 0000, by default?  If all is clear, I'm going to get rid of this entire 'Set default permissions for all log files'.

richm: > And we want to set the default permission 0700 to /etc/rsyslog.d, don't we? 

Yes, if not set by rpm install

> Lastly, can we expect root's umask is 0000, by default?

I think we can assume umask is set appropriately, yes.

> If all is clear, I'm going to get rid of this entire 'Set default permissions for all log files'.

ok

nhosoi: Thanks, @richm.  I've removed the `Set default permissions for all log file`s rule.
Just as a record, here are the config file ownerships and modes.  By default, `/etc/bashrc` sets `umask 022` for the privileged users (but in this case, it did not matter).
```
Note: without ownership and mode changes, the following settings are made.
-r--------. 1 root root 134 Mar 15 15:48 rsyslog.conf
drwxr-xr-x. 2 root root 277 Mar 15 15:48 rsyslog.d
-r--------. 1 root root 118 Mar 15 15:48 rsyslog.d/##-filename.conf
```

richm: hmm - not sure why rsyslog.d is 0755 - but I don't think that should concern us - if anything, it should be changed in the rpm, but I don't think it is a problem

nhosoi: Ok, the mode 0755 for /etc/rsyslog.d is coming from the spec file like this.  
```
install -d -m 755 %{buildroot}%{_sysconfdir}/sysconfig
install -d -m 755 %{buildroot}%{_sysconfdir}/logrotate.d
install -d -m 755 %{buildroot}%{_sysconfdir}/rsyslog.d
install -d -m 700 %{buildroot}%{rsyslog_statedir}
install -d -m 700 %{buildroot}%{rsyslog_pkidir}
install -d -m 755 %{buildroot}%{rsyslog_docdir}/html
```
But I also found the logging role is setting more ownerships and modes which looks unnecessary.  https://github.com/linux-system-roles/logging/pull/95/commits
Could you please review this second commit, too?  Thanks!

richm:
lgtm

nhosoi:
Ahh, sorry, @richm.  One more...
I've noticed /etc/rsyslog.conf is set to 0644 in the rsyslog spec file.
`install -p -m 644 %{SOURCE2} %{buildroot}%{_sysconfdir}/rsyslog.conf`
Should we respect the permission and set 0644 to /etc/rsyslog.conf and files in /etc/rsyslog.d?

richm:
> Ahh, sorry, @richm. One more...
> I've noticed /etc/rsyslog.conf is set to 0644 in the rsyslog spec file.
> `install -p -m 644 %{SOURCE2} %{buildroot}%{_sysconfdir}/rsyslog.conf`
> Should we respect the permission and set 0644 to /etc/rsyslog.conf and files in /etc/rsyslog.d?

@jvymazal - Any reason why the default config files and directories should be world readable?

nhosoi:
I'm going to merge this PR for now.
If the file permission is supposed to be different from the one described in this comment: https://github.com/linux-system-roles/logging/pull/95, I'm going to open another pr.

jvymazal:
> > Ahh, sorry, @richm. One more...
> > I've noticed /etc/rsyslog.conf is set to 0644 in the rsyslog spec file.
> > `install -p -m 644 %{SOURCE2} %{buildroot}%{_sysconfdir}/rsyslog.conf`
> > Should we respect the permission and set 0644 to /etc/rsyslog.conf and files in /etc/rsyslog.d?
> 
> @jvymazal - Any reason why the default config files and directories should be world readable?

Probably just historical thing, AFAIK they have been this way forever and since everybody is used to it, it was never touched. Main issue is for the files to be writeable only by root, I would say that the rest is not important so much.

nhosoi:
> Probably just historical thing, AFAIK they have been this way forever and since everybody is used to it, it was never touched. Main issue is for the files to be writeable only by root, I would say that the rest is not important so much.

Thanks for your input, @jvymazal.  May not be so important, but having a policy for the permissions would be useful for the further development and maintenance.  I've opened https://github.com/linux-system-roles/logging/pull/98.  If you could review it, I'd greatly appreciate it.

PR #96, opened 2020-03-13
Removing "changed_when: false" from creating a work and backup direct…

nhosoi:
…ory.

Changed needs to be bumped even when the change is outside of the rsyslog
configuration.

### DIFF
--- a/roles/rsyslog/tasks/main.yml
+++ b/roles/rsyslog/tasks/main.yml
@@ -70,7 +70,6 @@
         owner: '{{ rsyslog_user }}'
         group: '{{ rsyslog_file_group }}'
         mode: 0700
-      changed_when: false
 
     # Creating a backup dir for rsyslog.d
     - name: Set backup dir name


nhosoi:
/hold
The reason why changed_when: false was set to creating a work dir and a backup dir was to work around this idempotence error reported by the travis molecule test.
```
--> Scenario: 'default'
--> Action: 'idempotence'
ERROR: Idempotence test failed because of the following tasks:
* [instance] => /home/travis/build/linux-system-roles/logging/roles/rsyslog : Create a backup dir
An error occurred during the test sequence action: 'idempotence'. Cleaning up.
--> Scenario: 'default'
--> Action: 'cleanup'
Skipping, cleanup playbook not configured.
--> Scenario: 'default'
--> Action: 'destroy'
```

nhosoi:
Sorry, @richm, please hold on...  I'm still investigating the idempotence error...

roles/rsyslog/tasks/main.yml
```diff
@@ -84,7 +83,6 @@
         owner: '{{ rsyslog_user }}'
         group: '{{ rsyslog_file_group }}'
         mode: 0755
-      changed_when: false
```
richm: Yes, you need the `changed_when: false` here - this is to create the backup dir which is set here:
```yaml
    # Creating a backup dir for rsyslog.d
    - name: Set backup dir name
      set_fact:
        rsyslog_backup_dir: '{{ rsyslog_backup_dir | default("/tmp") }}/rsyslog.d-{{ ansible_date_time.iso8601_basic_short }}'
```
This changes every time you run it, so the task `    - name: Create a backup dir` will never be idempotent - it will always report that something changed.

I think we may need to rethink this backup strategy - it isn't very ansible-ish.  There will also be an accumulation of backup directories that the user may not know what to do with.  That backup directory is not very persistent e.g. if you reboot the machine your backup is wiped out.  TBD in another PR - we may have to adopt a different strategy, something like make a single backup directory like `/etc/rsyslog/backup`, and only copy a file there if it doesn't already exist.  That would be idempotent, would be persistent, and would allow users to at least revert their configuration to some past known state, if necessary.

nhosoi: Thanks, @richm.  In this PR, I'm going to put "changed_when: false" back to the backup dir, but just remove the one from creating the work dir which has a fixed path.

In terms of the new strategy, do you think "only copy a file there if it doesn't already exist" is enough?  How about if there are some difference between the 2 files?  You may want to keep the newer one (current configuration)?  Or shall I introduce another variable to choose new or old (I'd think it's an overkill, though...)

richm: > Thanks, @richm. In this PR, I'm going to put "changed_when: false" back to the backup dir, but just remove the one from creating the work dir which has a fixed path.

ok

> In terms of the new strategy, do you think "only copy a file there if it doesn't already exist" is enough? How about if there are some difference between the 2 files? You may want to keep the newer one (current configuration)? Or shall I introduce another variable to choose new or old (I'd think it's an overkill, though...)

It depends on what was the original intention of creating the backup directories and backing up the files.  If the use case was "as an admin, I want to run the `cleanup` playbook and restore my original files", then we want to keep only the first files, so that we can restore the config back to the original config.  IMO that's the only case that makes sense - once you run ansible, your config is really in your ansible inventory, so you can go back to some previous version (assuming you keep your inventory in git or some sort of revision control system).  So it doesn't make sense to backup config that was generated by ansible.

nhosoi: > > In terms of the new strategy, do you think "only copy a file there if it doesn't already exist" is enough? How about if there are some difference between the 2 files? You may want to keep the newer one (current configuration)? Or shall I introduce another variable to choose new or old (I'd think it's an overkill, though...)
> 
> It depends on what was the original intention of creating the backup directories and backing up the files. If the use case was "as an admin, I want to run the `cleanup` playbook and restore my original files", then we want to keep only the first files, so that we can restore the config back to the original config. IMO that's the only case that makes sense - once you run ansible, your config is really in your ansible inventory, so you can go back to some previous version (assuming you keep your inventory in git or some sort of revision control system). So it doesn't make sense to backup config that was generated by ansible.

Ok, it makes sense except an issue that we do not have a `cleanup` task for now...  How about the admin runs ansible with the newer inventory to upgrade the configuration?  In that case, don't they want to keep the current, pre-upgrade configuration?

richm: > How about the admin runs ansible with the newer inventory to upgrade the configuration? In that case, don't they want to keep the current, pre-upgrade configuration?

In general, no, I do not think so. @pcahyna do any of the other system roles attempt to make a backup of the previous configuration?  I think the only reason to keep a backup is the case where a user manually changes the file, and the system somehow detects this (as is the case with rpm when a user changes a config file provided by rpm).  This is contrary to the ansible way of doing things - we put a banner in the file like `# File managed by Ansible - DO NOT EDIT` because if you do edit it outside of ansible there is no way in general we can preserve the contents and still allow ansible to work.

nhosoi: Ok.  Currently, we are putting this comment at the top of each generated config file.
`# Ansible managed`
Shall I change it with the line you proposed (in the separate pr)?
`# File managed by Ansible - DO NOT EDIT`

nhosoi:
Anyway, I'm merging this pr (one line change).

roles/rsyslog/tasks/main.yml
richm: >  Shall I change it with the line you proposed (in the separate pr)?
# File managed by Ansible - DO NOT EDIT

No, I just meant it as an example - if a user sees that, and edits the file anyway, their edits will be overwritten by the next ansible run

nhosoi: Ok, but I'd prefer to have DO NOT EDIT in the comment as in https://github.com/linux-system-roles/logging/pull/97.  What do you think?

PR #97, opened 2020-03-16
Adding " - DO NOT EDIT" to the ansible_managed comment in each genera…

nhosoi:
…ted config file.

`  # Ansible managed - DO NOT EDIT`

### DIFF
--- a/roles/rsyslog/README.md
+++ b/roles/rsyslog/README.md
@@ -359,7 +359,7 @@ Type is for adding prefix to the file name to manage the order of the configurat
 
 If the deploy destination is other than {{ rsyslog_config_dir }}, the path is to be set to path.
 
-By default, the generated configuration file starts with a comment "# Ansible managed".  It could break some type of configurations.  For instance, "version=2" must be the first line in a rulebase file.  To avoid having "# Ansible managed", set true to nocomment.
+By default, the generated configuration file starts with a comment "# Ansible managed - DO NOT EDIT".  It could break some type of configurations.  For instance, "version=2" must be the first line in a rulebase file.  To avoid having "# Ansible managed", set true to nocomment.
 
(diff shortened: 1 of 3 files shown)


richm:
In linux system roles we don't use `{{ ansible_managed }}` very much, but in the places we use it, it is used like this:
`# {{ ansible_managed }}`
I'm not sure what the default text is - if it doesn't include "do not edit" then I think we should override this in the `defaults/main.yml`, rather than edit each file that is ansible managed and add text - @pcahyna ?

nhosoi:
> In linux system roles we don't use `{{ ansible_managed }}` very much, but in the places we use it, it is used like this:
> `# {{ ansible_managed }}`
> I'm not sure what the default text is -

It is "Ansible managed".  The logging role has been putting the comment `# Ansible managed` by `# {{ ansible_managed }}`.

>  if it doesn't include "do not edit" then I think we should override this in the `defaults/main.yml`, rather than edit each file that is ansible managed and add text - @pcahyna ?

Ok.  Just fyi, I tried to directly override ansible_managed in defaults/main.yml as well as in roles/rsyslog/defaults/main.yml.  But it's cleaned up and only "Ansible managed" is printed.  I guess we cannot use ansible_managed, but we should define our own internal variable?  LIke __rsyslog_ansible_managed?

richm:
>  LIke __rsyslog_ansible_managed?

I think we should not do anything about it for now.  Just leave it as `# {{ ansible_managed }}` and use the default text for now.  If it becomes a problem we can fix it later.  I also think we should solve this globally, for all system roles, if we need a solution.

nhosoi:
There is no plan to put "DO NOT EDIT" to the configuration files generated by the linux-system-roles. Closing this pr for now.

PR #98, opened 2020-03-17
Adjusting the dir/file modes to the ones defined in the package.

nhosoi:
Policy:
- Logging role sets permissions only when it generates new dirs/files.
- Configuration Directory - 0755
  Configuration files - 0644
- Directory to store file outputs (omfile) - if the directory needs
  to be created, the permission is 0700.

Per comments from @jvymazal, I'm proposing to have the above policy in the logging role.
https://github.com/linux-system-roles/logging/pull/95#issuecomment-599932174

But any further discussions would be welcome.  Thanks, in advance.

### DIFF
--- a/roles/rsyslog/tasks/main.yml
+++ b/roles/rsyslog/tasks/main.yml
@@ -64,25 +64,18 @@
         remove: '{{ true if rsyslog_purge_original_conf | bool else false }}'
       changed_when: false
 
-    - name: Get directory stat of rsyslog_system_log_dir (/var/log)
-      stat:
-        path: "{{ rsyslog_system_log_dir }}"
-      register: __rsyslog_system_log_dir_stat
-
-    - name: Update directory and file permissions
+    - name: Create logging directory if it does not exist or the ownership and/or modes are different.
       file:
         state: directory
         path: '{{ rsyslog_system_log_dir }}'
+        owner: 'root'
+        group: 'root'
         mode: '0700'
-      when: (not __rsyslog_system_log_dir_stat.stat.isdir) or
-            (__rsyslog_system_log_dir_stat.stat.uid != 0 or __rsyslog_system_log_dir_stat.stat.gid != 0) or
-            (__rsyslog_system_log_dir_stat.stat.mode != '0700')
 
     - name: Generate main rsyslog configuration
       template:
         src: 'etc/rsyslog.conf.j2'
         dest: '/etc/rsyslog.conf'
-        mode: '0400'
       when:
         - rsyslog_enabled | bool
       notify: restart rsyslogd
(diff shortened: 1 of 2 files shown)


roles/rsyslog/tasks/main.yml
```diff
@@ -64,25 +64,26 @@
...
         path: '{{ rsyslog_system_log_dir }}'
         mode: '0700'
-      when: (not __rsyslog_system_log_dir_stat.stat.isdir) or
-            (__rsyslog_system_log_dir_stat.stat.uid != 0 or __rsyslog_system_log_dir_stat.stat.gid != 0) or
-            (__rsyslog_system_log_dir_stat.stat.mode != '0700')
+      when: ((not __rsyslog_system_log_dir_stat.stat.exists) or
+             (not __rsyslog_system_log_dir_stat.stat.isdir) or
+             (__rsyslog_system_log_dir_stat.stat.uid != 0) or
+             (__rsyslog_system_log_dir_stat.stat.gid != 0) or
+             (__rsyslog_system_log_dir_stat.stat.mode != '0700'))
```
richm: The `file` module is supposed to do all of this `when` stuff for you.  AFAICT, you can replace all of the `when` conditions with
```yaml
file:
  state: directory
  path: '{{ rsyslog_system_log_dir }}'
  mode: '0700'
  owner: root
  group: root
```

pcahyna: and then you can get rid of the `stat`, if you don't need the result elsewhere.

nhosoi: One question.  Do we care the path already exists as a file, the ansible-playbook fails with this message.  It fails with my original patch, too.  :p  So, my question would be if the path exists not as a directory, do we want to remove it prior to the task or it's ok to fail?
```
    "mode": "0666",
    "msg": "/tmp/var/log already exists as a file",
    "owner": "nhosoi",
    "path": "/tmp/var/log",
```

richm: > One question. Do we care the path already exists as a file,

No, I don't think so.  If we do, there are probably several dozen places in other roles where we would need to put such a check.

> the ansible-playbook fails with this message. It fails with my original patch, too. :p So, my question would be if the path exists not as a directory, do we want to remove it prior to the task or it's ok to fail?

I think it is ok to fail.

> 
> ```
>     "mode": "0666",
>     "msg": "/tmp/var/log already exists as a file",
>     "owner": "nhosoi",
>     "path": "/tmp/var/log",
> ```

nhosoi: Thank you, @richm and @pcahyna.  Pushed the fix.

richm:
lgtm

nhosoi:
Thanks for reviewing the patch, @richm and @pcahyna.
Merging...

PR #99, opened 2020-03-17
Instead of using set_fact to set internal variables, use role variables in the main task file

nhosoi:
- Variables rsyslog_enabled, rsyslog_purge_original_conf, and rsyslog_system_log_dir are internal variables - set through external variables logging_enabled, logging_purge_confs and logging_system_log_dir, respectively. And the variables are renamed to start with '__'.
- Adding a test case tests_default_files_log_dir.yml

### DIFF
--- a/roles/rsyslog/vars/outputs/files/main.yml
+++ b/roles/rsyslog/vars/outputs/files/main.yml
@@ -34,7 +34,7 @@ __rsyslog_conf_files_output_modules:
 # __rsyslog_conf_files_output_rules:
 #
 # A set of default ``rsyslog`` options which store local system logs in files
-# located in `{{ rsyslog_system_log_dir }}/` directory. This is mostly the
+# located in `{{ __rsyslog_system_log_dir }}/` directory. This is mostly the
 # same as the default ``rsyslogd`` configuration provided in the installations.
 __rsyslog_conf_files_output_rules:
 
@@ -62,25 +62,25 @@ __rsyslog_conf_files_output_rules:
 
               # Log anything (except mail) of level info or higher.
               # Don't log private authentication messages!
-              *.info;mail.none;authpriv.none;cron.none                {{ rsyslog_system_log_dir }}/messages
+              *.info;mail.none;authpriv.none;cron.none                {{ __rsyslog_system_log_dir }}/messages
 
               # The authpriv file has restricted access.
-              authpriv.*                                              {{ rsyslog_system_log_dir }}/secure
+              authpriv.*                                              {{ __rsyslog_system_log_dir }}/secure
 
               # Log all the mail messages in one place.
-              mail.*                                                  -{{ rsyslog_system_log_dir }}/maillog
+              mail.*                                                  -{{ __rsyslog_system_log_dir }}/maillog
 
               # Log cron stuff
-              cron.*                                                  -{{ rsyslog_system_log_dir }}/cron
+              cron.*                                                  -{{ __rsyslog_system_log_dir }}/cron
 
               # Everybody gets emergency messages
               *.emerg                                                  :omusrmsg:*
 
               # Save news errors of level crit and higher in a special file.
-              uucp,news.crit                                          {{ rsyslog_system_log_dir }}/spooler
+              uucp,news.crit                                          {{ __rsyslog_system_log_dir }}/spooler
 
               # Save boot messages also to boot.log
-              local7.*                                                {{ rsyslog_system_log_dir }}/boot.log
+              local7.*                                                {{ __rsyslog_system_log_dir }}/boot.log
           }
           {%   endif %}
           {% endfor %}
@@ -108,42 +108,42 @@ __rsyslog_conf_remote_templates:
           template(
             name="RemoteMessage"
             type="string"
-            string="{{ rsyslog_system_log_dir }}/remote/msg/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log"
+            string="{{ __rsyslog_system_log_dir }}/remote/msg/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log"
           )
 
           # Template for Remote host auth logs
           template(
             name="RemoteHostAuthLog"
             type="string"
-            string="{{ rsyslog_system_log_dir }}/remote/auth/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log"
+            string="{{ __rsyslog_system_log_dir }}/remote/auth/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log"
           )
 
           # Template for Remote host cron logs
           template(
             name="RemoteHostCronLog"
             type="string"
-            string="{{ rsyslog_system_log_dir }}/remote/cron/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log"
+            string="{{ __rsyslog_system_log_dir }}/remote/cron/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log"
           )
 
           # Template for Remote service auth logs
           template(
             name="RemoteServiceAuthLog"
             type="string"
-            string="{{ rsyslog_system_log_dir }}/remote/services/auth/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log"
+            string="{{ __rsyslog_system_log_dir }}/remote/services/auth/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log"
           )
 
           # Template for Remote service cron logs
           template(
             name="RemoteServiceCronLog"
             type="string"
-            string="{{ rsyslog_system_log_dir }}/remote/services/cron/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log"
+            string="{{ __rsyslog_system_log_dir }}/remote/services/cron/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log"
           )
 
           # Template for Remote service mail logs
           template(
             name="RemoteServiceMailLog"
             type="string"
-            string="{{ rsyslog_system_log_dir }}/remote/services/mail/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log"
+            string="{{ __rsyslog_system_log_dir }}/remote/services/mail/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log"
           )
 
           # Include custom templates
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -2,10 +2,6 @@
 - name: Set Rsyslog facts then include rsyslog role
   block:
 
-    - name: Set rsyslog_enabled
-      set_fact:
-        rsyslog_enabled: "{{ logging_enabled | d(true) }}"
-
     - name: Set rsyslog_output_elasticsearch for the omelasticsearch output
       set_fact:
         rsyslog_output_elasticsearch: "{{ logging_outputs | selectattr('name', 'defined') | selectattr('type', 'defined') | selectattr('type', '==', 'elasticsearch') | list }}"
@@ -28,11 +24,6 @@
     - name: Set rsyslog_outputs
       set_fact:
         rsyslog_outputs: '{{ rsyslog_output_elasticsearch | d([]) }} + {{ rsyslog_output_files | d([]) }} + {{ rsyslog_output_forwards | d([]) }}'
-
-    - name: Set rsyslog_purge_original_conf fact to purge all configuration files before saving the new ones.
-      set_fact:
-        rsyslog_purge_original_conf: yes
-      when: logging_purge_confs | d(false)
 
     - name: Set custom_config_files fact
       set_fact:
@@ -79,6 +70,10 @@
         - logging_debug | d(false)
 
     - name: Include Rsyslog role
+      vars:
+        __rsyslog_enabled: "{{ logging_enabled }}"
+        __rsyslog_purge_original_conf: "{{ logging_purge_confs }}"
+        __rsyslog_system_log_dir: "{{ logging_system_log_dir }}"
       include_role:
         name: "{{ role_path }}/roles/rsyslog"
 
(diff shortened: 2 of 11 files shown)


tasks/main.yml
```diff
@@ -65,6 +65,10 @@
         rsyslog_purge_original_conf: yes
       when: logging_purge_confs | d(false)
 
+    - name: Set the value of logging_system_log_dir to __rsyslog_system_log_dir
+      set_fact:
+        __rsyslog_system_log_dir: "{{ logging_system_log_dir }}"
```
richm: Future work for another PR - this and many of these could be rsyslog role parameters, instead of using `set_fact`.

nhosoi: @richm, is this diff close to what you suggested?
https://github.com/nhosoi/logging/compare/rsyslog_system_log_dir...nhosoi:remove_set_fact

If it looks ok, may I add the commit to this pr?  Or had I better open another pr just for that once this pr is merged?  Thanks.

richm: Looks ok to me - it seems a bit odd to have "private" parameters passed to the role as vars, but these are internal interfaces not meant for end users anyway.

nhosoi: Thank you, @richm.  Let me add the patch to this pr, then. :)

tasks/main.yml
```diff
@@ -110,6 +101,10 @@
         - logging_debug | d(false)
 
     - name: Include Rsyslog role
+      vars:
+        __rsyslog_enabled: "{{ logging_enabled | d(true) }}"
+        __rsyslog_purge_original_conf: "{{ logging_purge_confs | d(false) }}"
+        __rsyslog_system_log_dir: "{{ logging_system_log_dir | d('/var/log') }}"
```
richm: Not sure if we've had this discussion before, so forgive me and remind me if we have, or if this is different since rsyslog is a sub-role of logging, but . . .

. . . you don't need the `| d()` filter for these since they are defined in `logging/defaults/main.yml` - @pcahyna please correct me if I'm wrong, but you can do this:
```yaml

      vars:
        __rsyslog_enabled: "{{ logging_enabled }}"
        __rsyslog_purge_original_conf: "{{ logging_purge_confs }}"
        __rsyslog_system_log_dir: "{{ logging_system_log_dir }}"
```
again, if I'm forgetting some discussion we already had about this, let me know

nhosoi: @richm, @pcahyna, sorry, give me one sec.  I thought I should be able to remove the default filters safely, but I'm observing some test failures and need to debug it...
I set hold label.

nhosoi: @richm, @pcahyna, I pushed the [3rd patch](https://github.com/linux-system-roles/logging/pull/99/commits/d62a7dff24cd1810ad1fd9e886079d6426147d29
) following the suggestions from @richm.

Not just removing the default filter from the vars, I had to move the default setting `logging_system_log_dir: /var/log` from the top level defaults/main.yml to the rsyslog level roles/rsyslog/defaults/main.yml.  It was needed due to the logging specific role structure (please see also https://github.com/linux-system-roles/logging/blob/master/design_docs/logging_subsystem.md).

The top level tasks/main.yml includes the rsyslog role (https://github.com/nhosoi/logging/blob/rsyslog_system_log_dir/tasks/main.yml#L108-L109), then in the rsyslog role, eventually each sub-role (such as https://github.com/nhosoi/logging/tree/rsyslog_system_log_dir/roles/rsyslog/roles/input_roles/basics and https://github.com/nhosoi/logging/tree/rsyslog_system_log_dir/roles/rsyslog/roles/output_roles/files) includes https://github.com/nhosoi/logging/blob/rsyslog_system_log_dir/roles/rsyslog/tasks/deploy.yml like this (https://github.com/nhosoi/logging/blob/rsyslog_system_log_dir/roles/rsyslog/roles/output_roles/files/tasks/main.yml#L7-L9).  Usually, if the default value is assigned to a variable in the top level defaults/main.yml, it should be available in any place underneath, I suppose. But it's not after 2 step of including's.  Unless I move `logging_system_log_dir` to roles/rsyslog/defaults/main.yml, it fails with `logging_system_log_dir` undefined.

If the move is not acceptable, there are 2 alternatives.
1) moving back to set_fact.
2) giving up to have rsyslog internal vars, and exposing rsyslog vars as external and using them in the rsyslog roles.

Do you think what'd be the best?  Thanks!

nhosoi: Reminder to myself: I have to do the same thing on __rsyslog_purge_original_conf, too.  (maybe in another PR)

richm: So that means `logging_system_log_dir` set in the top level `defaults/main.yml` is undefined by the time the rsyslog role is called?  So at this point: 
```yaml
      vars:
        __rsyslog_enabled: "{{ logging_enabled }}"
        __rsyslog_purge_original_conf: "{{ logging_purge_confs }}"
        __rsyslog_system_log_dir: "{{ logging_system_log_dir }}"
```
this means that `logging_system_log_dir` is always undefined?  So what is the point of using it at all?  Are you saying that if you use `set_fact` at the top level to set `logging_system_log_dir`, it will be defined at the point above?  @greg-hellings have you ever seen anything like this before?  @maxamillion is this some sort of strange bug in ansible?

greg-hellings: `set_fact` has exceptionally high priority for values (I think only command-line defined parameters override it, if even that?) and are set globally throughout the entire rest of a playbook run. So it would make sense that using `set_fact` works. But also, because of that, using `set_fact` is, generally, a terrible idea. Especially for something that's just a parameter to a role.

I can't comment on what's "usual" when it comes to embedded roles. We steadfastly refuse to use them anywhere in OASIS, and I haven't yet encountered a reason to use them that wasn't covered just as well by promoting the subrole to a top level role. In the old days, before collections, that could create dependency problems. But now that we're in the Collections era, why not convert this to a collection, have this role and all its children as peer roles, and include them from within the same collection? If there really is a reason that these roles don't deserve to be first-class citizens, then why are they roles at all and not just task import/includes?

nhosoi: > So that means `logging_system_log_dir` set in the top level `defaults/main.yml` is undefined by the time the rsyslog role is called? So at this point:
<< snip >>
> this means that `logging_system_log_dir` is always undefined? So what is the point of using it at all?

If you set logging_system_log_dir in your inventory file, the value is respected.
https://github.com/nhosoi/logging/blob/rsyslog_system_log_dir/tests/tests_default_files_log_dir.yml#L6

nhosoi: Thank you,  @greg-hellings.
> But now that we're in the Collections era, why not convert this to a collection, have this role and all its children as peer roles, and include them from within the same collection? If there really is a reason that these roles don't deserve to be first-class citizens, then why are they roles at all and not just task import/includes?

I've never used collection yet...  Could you please give me pointer(s) to learn how to do the conversion?  Thanks!

richm: We are currently planning about how and when to convert all of the linux-system-roles to collections.  So don't worry about this for now for logging (although if you want to learn about ansible collections, please do).

As for the current problem, it seems that the only way to have `logging_system_log_dir` in scope when the rsyslog sub-role is called is to use `set_fact`, so please do that, and create another issue to revisit this.

nhosoi: I pushed this patch to use set_fact.  Could you review it one more time, @richm, @pcahyna?
Thanks!!
https://github.com/linux-system-roles/logging/pull/99/commits/b440c7d9504744fa1cd057fafe75b4f32c4c6742

richm:
lgtm - but since this is potentially controversial, I would prefer another reviewer

nhosoi:
> lgtm - but since this is potentially controversial, I would prefer another reviewer

Thanks, @richm.  Anyway, I opened #102 not to forget this issue...

roles/rsyslog/vars/main.yml
```diff
@@ -1,3 +1,8 @@
+# __rsyslog_enabled
+#
+# Enable or disable ``rsyslog`` management.
+__rsyslog_enabled: true
```
sradco, line 4: @nhosoi Hi, why did you move this to vars? Is it always enabled? And same question for the 
__rsyslog_system_log_dir: '/var/log' below.

I believe they should be defaults so that they can be updated.no?

nhosoi: > @nhosoi Hi, why did you move this to vars? Is it always enabled? And same question for the
> __rsyslog_system_log_dir: '/var/log' below.
> 
> I believe they should be defaults so that they can be updated.no?

The motivation is turning them (`rsyslog_enabled, rsyslog_system_log_dir`) to the internal variables (by starting them with '__' and moving them to the vars files) and only allowing to set them via corresponding the external variables.  For instance, use `logging_enabled` for `__rsyslog_enabled`; `logging_system_log_dir` for `__rsyslog_system_log_dir`.

Having them set in the vars file, I believe the values are just the default values.  So, __rsyslog_enabled is true, by default, and you can set logging_enabled to false to disable it.  But it was a very good point, @sradco.  All of the CI tests are setting logging_enabled to true.  They are not needed.  And I'm not testing `logging_enabled: false`.  I'm adding it.  Thanks!

richm: This is the `logging` role - whether to use `rsyslog` or not is supposed to be an implementation detail, so most of the parameters exposed to end users of the `logging` role will be `logging_*` parameters.  Of course, there will be a need for users to supply provider specific parameters e.g. as a user of the logging role, I know for a fact that I want to use `rsyslog`, and I must supply some `rsyslog` specific parameters.

nhosoi: > This is the `logging` role - whether to use `rsyslog` or not is supposed to be an implementation detail, so most of the parameters exposed to end users of the `logging` role will be `logging_*` parameters. Of course, there will be a need for users to supply provider specific parameters e.g. as a user of the logging role, I know for a fact that I want to use `rsyslog`, and I must supply some `rsyslog` specific parameters.

Thanks, @richm.  That is correct.  This is going to be the minimum configuration - input from imjournal, output to the local files.  No rsyslog indication in it. (of course, this is because logging_collector is set to rsyslog in the defaults/main.yml, though. ;)
```
      vars:
        logging_collections:
          - name: basic_input
            type: basics
```

pcahyna: @nhosoi 

> Having them set in the vars file, I believe the values are just the default values. So, __rsyslog_enabled is true, by default, and you can set logging_enabled to false to disable it.

No, variables set in vars are not default values. They are very high in the priority list, and it is less easy to override them. This is mentioned here: https://github.com/oasis-roles/meta_standards/blob/master/README.md#vars-vs-defaults, specifically: "Avoid giving default values in vars/main.yml as such values are very high in the precedence order and are difficult for users and consumers of a role to override." For details on variable precedence, see https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable . Note that defaults are on level 2, while role vars are on level 15. It may be the case that the way you are overriding them actually works, still, to conform to the best practices and reduce confusion, they should be given in defaults IMO.
Can you perhaps better explain the motivation for moving them from defaults to vars, please?

tasks/main.yml
pcahyna: > As for the current problem, it seems that the only way to have `logging_system_log_dir` in scope when the rsyslog sub-role is called is to use `set_fact`, so please do that, and create another issue to revisit this.

Would it work to keep the variable in defaults and not moving it to vars?

pcahyna: > If the move is not acceptable, there are 2 alternatives.
> 
>     1. moving back to set_fact.
> 
>     2. giving up to have rsyslog internal vars, and exposing rsyslog vars as external and using them in the rsyslog roles.

I am interested what option 2 means. Where would be rsyslog vars exposed? In which defaults.yml file? Of the `rsyslog` role?

pcahyna: @nhosoi @greg-hellings I found that the problem was not due to embedded roles (lexically nested), but in dynamic nesting in the role call chain, specifically, when a role calls indirectly itself (indirect recursion). So moving from embedded roles to a flat structure will not help. One will have to stop letting `rsyslog` call `basics` and `basics` call back `rsyslog`. This is shown in a minimal reproducer here: https://github.com/pcahyna/nested_roles_demo

maxamillion: Dropping this here to make sure it's been consulted, following the thread it seems the like the var precedence was called into question. https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#ansible-variable-precedence

@richm I'd need an example playbook and the corresponding roles that are showing the issue but it's entirely possible and I'd be interested in taking a look. We can follow-up off-thread if you like.

pcahyna: @maxamillion example playbook and corresponding roles is here: https://github.com/pcahyna/nested_roles_demo/blob/master/logging_flattened/testplaybook.yml . See https://github.com/pcahyna/nested_roles_demo/blob/master/README for a brief explanation.

pcahyna: @nhosoi I suppose the issue would be fixed by #108 and `set_fact` could be eliminated?

nhosoi: > @nhosoi I suppose the issue would be fixed by #108 and `set_fact` could be eliminated?

@pcahyna, let me test it...

nhosoi: > @nhosoi I suppose the issue would be fixed by #108 and `set_fact` could be eliminated?

@pcahyna, yes, it is fixed by https://github.com/linux-system-roles/logging/pull/108.  Let me apply the patch to the master branch.  Thanks.

roles/rsyslog/vars/outputs/files/main.yml
```diff
@@ -62,25 +62,25 @@ __rsyslog_conf_files_output_rules:
 
               # Log anything (except mail) of level info or higher.
               # Don't log private authentication messages!
-              *.info;mail.none;authpriv.none;cron.none                {{ rsyslog_system_log_dir }}/messages
```
richm, line 65: Proposal for another PR - Could this inline string template be replaced by roles/rsyslog/templates/etc/rsyslog.conf.j2 ?

richm:
lgtm

roles/rsyslog/vars/outputs/files/main.yml
pcahyna: sure, inline string templates would be better to have as template files (maybe avoid the `/etc`in the path according to https://github.com/oasis-roles/meta_standards/#ansible-best-practices "Avoid duplicating the remote full path in the role directory, however, as that creates unnecessary depth in the file tree for the role." )

pcahyna:
lgtm as well

nhosoi:
Thank you for the reviews, @richm, @pcahyna.  Merging this pr#99.

In terms of the templates, will be taken care in the next task "RHELPLAN-41563 - Logging - splitting an output config file into multiple ones that contains just one action".  Thanks!

PR #100, opened 2020-03-24
Support combining input and output roles to configure complex flows

nhosoi:
This is the second cut PR to implement the flow control in the rsyslog configuration and eliminate the rsyslog_default variable.  Thanks to @richm for the reviews and suggestions to the first cut.  I updated the PR following his advice especially on the rulesets.

For more details, please take a look at this proposal.
https://github.com/nhosoi/logging/blob/combination/design_docs/rsyslog_new_input_output_roles_proposal.md

Since this PR is large, I divided it into 7 implementation + 1 warning fix + 1 design doc commits.


nhosoi:
??? Is the CI test disabled ???

design_docs/rsyslog_new_input_output_roles_proposal.md
```diff
@@ -0,0 +1,254 @@
...
+ruleset(name="files_forwards") {
+  $IncludeConfig /etc/rsyslog.d/*.files
+  $IncludeConfig /etc/rsyslog.d/*.forwards
+}
+
+```
+In this PR, just files and forwards outputs are supported.  As described in [To Do's](#to-dos), elasticsearch is going to be added next.
+
+### Input roles
+Based upon the inventry file that users provide, input config jumps to the ruleset.
```
richm: inventory

design_docs/rsyslog_new_input_output_roles_proposal.md
```diff
@@ -0,0 +1,254 @@
...
+#
+ruleset(name="files") {
+  $IncludeConfig /etc/rsyslog.d/*.files
+}
+ruleset(name="forwards") {
+  $IncludeConfig /etc/rsyslog.d/*.forwards
+}
+ruleset(name="files_forwards") {
+  $IncludeConfig /etc/rsyslog.d/*.files
+  $IncludeConfig /etc/rsyslog.d/*.forwards
```
richm: Is this the convention that debops uses?
If not - I think it will be confusing to have rsyslog config files with different suffixes - .conf, .files, .forwards, etc. - but it can't be .conf because it will be picked up by the main rsyslog.conf `$IncludeConfig /etc/rsyslog.d/*.conf` . . . what about having separate subdirectories?  e.g. /etc/rsyslog.d/files/ - or did we already say that wasn't a good idea either?

design_docs/rsyslog_new_input_output_roles_proposal.md
```diff
@@ -0,0 +1,254 @@
...
+```
+module(load="imjournal"
+       StateFile="/var/lib/rsyslog/imjournal.state"
+       RateLimit.Burst="300000"
+       RateLimit.Interval="30"
+       PersistStateInterval="3000")
+call files
+```
+Please note that `imfile` input supports `ruleset` parameter, thus the parameter is used to specify which ruleset to call.
+Other plugins which do not have such a parameter, e.g., `imjournal`, use the `call` statement.
```
richm: The `call files` will be called for all inputs, not just the `imjournal` input - as opposed to the `input` with the `ruleset`.

richm: You could do something like this (not sure of the field names or the syntax):
```
if $inputname == "imjournal" then {
    call files
}
```

design_docs/rsyslog_new_input_output_roles_proposal.md
```diff
@@ -0,0 +1,254 @@
...
+  - name: forward-output
+    type: forwards
+    rsyslog_forwards_actions:
+      - name: to-remote
+        protocol: tcp
+        target: remote_host_name.remote_domain_name
+        port: 514
+  - name: file-output
+    type: files
+    logs_collections:
```
richm: typo here?

design_docs/rsyslog_new_input_output_roles_proposal.md
```diff
@@ -0,0 +1,254 @@
...
+logs_collections:
+  - name: basic-input
+    type: basics
+    rsyslog_use_files_ruleset: true
+  - name: file-input
+    type: files
+    rsyslog_use_files_ruleset: true
+    rsyslog_use_forwards_ruleset: true
+```
+In this PR, I propose to make the logs_collections (input roles) separate from logging_outputs (output roles) and specify the relationship using boolean parameters rsyslog_use_files_ruleset and/or rsyslog_use_forwards_ruleset.
```
richm: Is whether or not to use a `ruleset` an implementation detail?  That is, what difference does it make to the user if the input is sent to the output using a `ruleset` or some other mechanism?  That is, I would think the user would just want to specify "collect logs from these inputs, and send them to these outputs" - and that can be done without the `_ruleset` suffix e.g. `rsyslog_use_files` instead of `rsyslog_use_files_ruleset`

design_docs/rsyslog_new_input_output_roles_proposal.md
```diff
@@ -0,0 +1,254 @@
...
+ruleset(name="files_forwards_kafka") {
+  $IncludeConfig /etc/rsyslog.d/*.files
+  $IncludeConfig /etc/rsyslog.d/*.forwards
+  $IncludeConfig /etc/rsyslog.d/*.kafka
+}
+ruleset(name="elasticsearch_files_forwards_kafka") {
+  $IncludeConfig /etc/rsyslog.d/*.elasticsearch
+  $IncludeConfig /etc/rsyslog.d/*.files
+  $IncludeConfig /etc/rsyslog.d/*.forwards
+  $IncludeConfig /etc/rsyslog.d/*.kafka
```
richm: Does this mean, in the future, if we support 8 different output types, we will need `8!` (8 factorial) different rulesets?  That seems hard to scale and manage . . .
What about `files_elasticsearch`, `forwards_elasticsearch`, etc.?

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: debops uses *.system for the system logs and *.remote for the logs sent from the remote hosts.  I extended it.

No, not considered the subdirectory in this case.  We used to have /etc/rsyslog.d/{elasticsearch,viaq}, etc.  then flattened them.  That was the only opportunity we tried the subdirectories so far.  I'd think there's no difficulties to switch suffixes to subdirectories.

richm: > debops uses *.system for the system logs and *.remote for the logs sent from the remote hosts. I extended it.

Ok - works for me.

> 
> No, not considered the subdirectory in this case. We used to have /etc/rsyslog.d/{elasticsearch,viaq}, etc. then flattened them. That was the only opportunity we tried the subdirectories so far. I'd think there's no difficulties to switch suffixes to subdirectories.

I think the filename suffix `*.files`, `*.remote`, etc. is fine.  I think it is an implementation detail anyway - users will typically not need to know or care about this.

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: In the rsyslog config, if it is ordered as follows, do you think `call files` is effective just for imjournal?  Or log messages from other im-plugins take the path?
```
module(load="imjournal" ...)
call files
input(name="basics_imtcp" type="imtcp ... ruleset="remote")
input(type="imfile" file="/path/to/input_file" ... ruleset="files_forwards")
```
The 2 input plugins im{tcp,udp} and imfile except imjournal support the input function, with which the ruleset to jump could be managed.

If the above order satisfies our purpose, I think we are good...

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: > Is whether or not to use a `ruleset` an implementation detail? That is, what difference does it make to the user if the input is sent to the output using a `ruleset` or some other mechanism? 

Ahhh, there's no **other** mechanism for the output...  Only via rulesets.

> That is, I would think the user would just want to specify "collect logs from these inputs, and send them to these outputs" - and that can be done without the `_ruleset` suffix e.g. `rsyslog_use_files` instead of `rsyslog_use_files_ruleset`

I could add a code that if there is no rsyslog_use_.*_ruleset in the inventory, we enable rsyslog_use_files_ruleset, which means output to the local file (e.g., /var/log/messages).  Do you think that's most customers would like to have?  But to do so, the files output needs to be configured (i.e., 30-output.files has to be in /etc/rsyslog.d).  Thinking of it, it might be safer to deploy the files output role regardless of this files configuration is given or not...
```
logging_outputs:
  - name: files-output
    type: files
```

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: I know...  That's the ugliest part in this proposal...  In terms of the non-alphabetically ordered name, we should be able to sort or map to get the right ruleset name, I'd think.

If we could ignore the non-alphabetically ordered ruleset names, the number would be (1+2+3+...N) = (1+N) * N/2.  If we support 8 output roles, then there would be 36 of them... :(

design_docs/rsyslog_new_input_output_roles_proposal.md
richm: > > Is whether or not to use a `ruleset` an implementation detail? That is, what difference does it make to the user if the input is sent to the output using a `ruleset` or some other mechanism?
> 
> Ahhh, there's no **other** mechanism for the output... Only via rulesets.

There is no requirement to use rulesets.  If you do not specify a ruleset, the input goes into the default/general record flow (just like what happens when you use `imjournal`).  You would then have to use `if` statements to control the flow if you wanted a specific set of inputs to go to a specific set of outputs.

> 
> > That is, I would think the user would just want to specify "collect logs from these inputs, and send them to these outputs" - and that can be done without the `_ruleset` suffix e.g. `rsyslog_use_files` instead of `rsyslog_use_files_ruleset`
> 
> I could add a code that if there is no rsyslog_use_.*_ruleset in the inventory, we enable rsyslog_use_files_ruleset, which means output to the local file (e.g., /var/log/messages). Do you think that's most customers would like to have? But to do so, the files output needs to be configured (i.e., 30-output.files has to be in /etc/rsyslog.d). Thinking of it, it might be safer to deploy the files output role regardless of this files configuration is given or not...
> 
> ```
> logging_outputs:
>   - name: files-output
>     type: files
> ```

I'm not sure, but does it make a difference if you do not use rulesets at all?

design_docs/rsyslog_new_input_output_roles_proposal.md
richm: I guess that would work - we would have to be very careful about the order of modules/inputs

design_docs/rsyslog_new_input_output_roles_proposal.md
richm: That just seems wrong to me - what if you don't use rulesets so much, or figure out a way to only have a ruleset per output, and combine rulesets with config code e.g. something like
```
if outputs_string contains " kafka " then {
    call kafka
}
if outputs_string contains " elasticsearch " then {
    call elasticsearch
}
```
not sure about the syntax - but the idea is that you don't have inputs strictly bound to rulesets

nhosoi: > That just seems wrong to me - what if you don't use rulesets so much, or figure out a way to only have a ruleset per output, and combine rulesets with config code e.g. something like
> 
> ```
> if outputs_string contains " kafka " then {
>     call kafka
> }
> if outputs_string contains " elasticsearch " then {
>     call elasticsearch
> }
> ```

Thanks, @richm!!  My original code was as simple as like this...  Somehow my test failed to work back then, and I thought consecutive calls is not valid.  I tested it again and found it was my test problem... *shame shame* Let me update the code...
```
{% if rsyslog_use_files_ruleset %}
call files
{% endif %}
{% if rsyslog_use_forwards_ruleset %}
call forwards
{% endif %}
```

design_docs/rsyslog_new_input_output_roles_proposal.md
```diff
@@ -0,0 +1,255 @@
...
+  - name: forward-output
+    type: forwards
+    rsyslog_forwards_actions:
+      - name: to-remote
+        protocol: tcp
+        target: remote_host_name.remote_domain_name
+        port: 514
+  - name: file-output
+    type: files
+logs_collections:
```
pcahyna: should be `logging_collections`, no? According to standards. Also, wouldn't it be more intuitive to call it `inputs` for symmetry with `outputs`, instead of `collections`?

design_docs/rsyslog_new_input_output_roles_proposal.md
```diff
@@ -0,0 +1,255 @@
...
+logging_outputs:
+  - name: forward-output
+    type: forwards
+    rsyslog_forwards_actions:
+      - name: to-remote
+        protocol: tcp
+        target: remote_host_name.remote_domain_name
+        port: 514
+  - name: file-output
+    type: files
```
pcahyna: Can there be multiple file outputs, like this?
```
- name: file-output1
  type: files
- name: file-output2
  type: files
```
or, in general, multiple outputs with the same name? I guess at least for `forwards` it would make sense (define forwards to multiple hosts).

design_docs/rsyslog_new_input_output_roles_proposal.md
```diff
@@ -0,0 +1,255 @@
...
+logs_collections:
+  - name: basic-input
+    type: basics
+    rsyslog_use_files_ruleset: true
```
pcahyna: Isn't it better to refer to outputs by their name? I.e. if file output is defined as `name: file-output`, refer to it by the name `file-output`, rather than using a boolean. This would make sense especially if one can specify multiple outputs of the same type, because then how will you distinguish between them otherwise?

design_docs/rsyslog_new_input_output_roles_proposal.md
```diff
@@ -0,0 +1,255 @@
...
+  - name: file-input
+    type: files
+    rsyslog_use_files_ruleset: true
+    rsyslog_use_forwards_ruleset: true
+```
+There are some exceptions.
+- The files output is configured, by default. That is, even if there is no input roles that write logs to the local files, the files output config file is deployed. The filters and the actions in the configuration file (30-output.files) are equivalent to the ones in the original rsyslog.conf.
```
pcahyna: What does it mean "by default"? Does it mean that if you don't supply any `logging_outputs`, there will be a default one consisting of one files output? (That's what I understand by "default".) Or, do you mean that a files output will be always appended to user-supplied `logging_outputs`, if it does not have any?

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: I like the idea.
@sradco, is it ok to change the variable name?

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: > Can there be multiple file outputs, like this?
> 
> ```
> - name: file-output1
>   type: files
> - name: file-output2
>   type: files
>
> or, in general, multiple outputs with the same type? I guess at least for `forwards` it would make sense (define forwards to multiple hosts).
> ```

Actually, having multiple files declarations is not supported.  To configure multiple file outputs, you could use  `rsyslog_files_actions` like this.
https://github.com/linux-system-roles/logging/blob/master/tests/tests_files.yml#L13-L28

Forwarding also has the corresponding setting `rsyslog_forwards_actions`
https://github.com/linux-system-roles/logging/blob/master/tests/tests_forwards.yml#L14-L47

I have to confess I haven't tested this case (thanks for asking me!) to find the first file-output0 is just overwritten by file-outpu1 and no test0 and test1 actions are configured.  I will add this to the doc that output role is supposed to be one type each.
```
        logging_outputs:
          - name: file-output0
            type: files
            rsyslog_files_actions:
              - name: test0
                severity: info
                exclude: authpriv.none;auth.none;cron.none;mail.none
                path: /var/log/messages
              - name: test1
                facility: authpriv,auth
                path: /var/log/secure
          - name: file-output1
            type: files
            rsyslog_files_actions:
              - name: test2
                severity: emerg
                path: :omusrmsg:*
              - name: test3
                facility: local7
                path: /var/log/boot.log
```

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: The purpose of this implementation is to call `call` statement as simple as this.
https://github.com/nhosoi/logging/blob/combination/roles/rsyslog/roles/input_roles/files/defaults/main.yml#L66-L71

We need to have an ability to deliver the log messages to one particular output as well as multiple ones.  That being said, are you suggesting something like this?  (please note that instead of `name`, i'm putting `type` in the array since only one instance is available per one output role as described in https://github.com/linux-system-roles/logging/pull/100#discussion_r401131546.) If this looks ok, I will update this patch.
```
logs_collections:
  - name: basic-input
    type: basics
    rsyslog_outputs: ['files', 'forwards']
```

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: > What does it mean "by default"? Does it mean that if you don't supply any `logging_outputs`, there will be a default one consisting of one files output? (That's what I understand by "default".) 

Correct.

> Or, do you mean that a files output will be always appended to user-supplied `logging_outputs`, if it does not have any?

That's correct, too.  The files output is always appended.  I thought it provides the same functionality as the current rsyslog_default does - deploying the original rsyslog.conf.

design_docs/rsyslog_new_input_output_roles_proposal.md
sradco: @nhosoi Yes.it makes sense. I'll need to update this for ovirt.

nhosoi: Thanks, @sradco!
@pcahyna, is it ok to make the change in this PR or do you recommend to create a separate PR?

pcahyna: if you mean the change to the `logs_collections` -> `logging_inputs` variable name, then please do it in this PR. It would be just confusing to introduce an unwanted variable name in one PR and then change it in another one.

design_docs/rsyslog_new_input_output_roles_proposal.md
pcahyna: > Actually, having multiple files declarations is not supported.

What about multiple declarations for other types? Is it a general rule that every output type has only one declaration, or can there be declarations without this limitation?

design_docs/rsyslog_new_input_output_roles_proposal.md
pcahyna: yes, something like that. But I still think that putting `name` instead of `type` in the array makes more sense, because `name` is supposed to be the unique identifier. The role can resolve `name` to type internally. If the role starts supporting multiple output declaration for each type, it will be easy to support without changing the interface. (And if you think that this will never be supported, why having separate `name` and `type` at all?)

design_docs/rsyslog_new_input_output_roles_proposal.md
pcahyna: What about having the default only for situations where the user has not supplied any `logging_outputs`? It is quite irregular to have a magic value that gets always appended.

pcahyna: That's a problem with default values for lists in Ansible. it is not easy to have a list with some default values and then append to the list instead of overriding it.

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: ??? logs_collections has been in the logging role for a long time.  So, it's just a replacement.  If you don't mind, I'm more than happy to do it in this PR.

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: > > Actually, having multiple files declarations is not supported.
> 
> What about multiple declarations for other types?

That is supported.  This is a test case to have the files outputs and the forwards outputs.
https://github.com/nhosoi/logging/blob/combination/tests/tests_combination.yml#L19-L43
In the files outputs, 2 output files - test0 and test1 - are configured; In the forwards outputs, 2 forwards - severity_and_facility and facility_only are configured.

> Is it a general rule that every output type has only one declaration, or can there be declarations without this limitation?

The former.  It is a general rule.  To support output type specific configuration as `rsyslog_files_actions` for the files outputs and `rsyslog_forwards_action`s for the forwards outputs, we have to add the special vars.  I believe the elasticsearch outputs already have pretty good support on them.  But we need to enhance and test them.

design_docs/rsyslog_new_input_output_roles_proposal.md
pcahyna: I don't see it. I currently see `log_collections` only as an attribute of dicts. With the proposed change, you would turn it into a variable. I.e. introduce a new variable.

design_docs/rsyslog_new_input_output_roles_proposal.md
pcahyna: > > > Actually, having multiple files declarations is not supported.
> > 
> > 
> > What about multiple declarations for other types?
> 
> That is supported. This is a test case to have the files outputs and the forwards outputs.
> https://github.com/nhosoi/logging/blob/combination/tests/tests_combination.yml#L19-L43
> In the files outputs, 2 output files - test0 and test1 - are configured; In the forwards outputs, 2 forwards - severity_and_facility and facility_only are configured.

Well according to what you write below, it is a general rule that every output thype has only one declaration. Your example shows only one declaration for `files` and `forwards`.
> 
> > Is it a general rule that every output type has only one declaration, or can there be declarations without this limitation?
> 
> The former. It is a general rule. To support output type specific configuration as `rsyslog_files_actions` for the files outputs and `rsyslog_forwards_action`s for the forwards outputs, we have to add the special vars. I believe the elasticsearch outputs already have pretty good support on them. But we need to enhance and test them.

If this is the case, isn't the `name` a bit redundant? `type` seems to uniquely identify the entry. What is `name` used for?

pcahyna: It could be beneficial to have `name` to make it look more like the interface of other roles, but maybe it could be allowed to omit it and the role would derive it from `type`?

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: > yes, something like that. But I still think that putting `name` instead of `type` in the array makes more sense, because `name` is supposed to be the unique identifier. The role can resolve `name` to type internally. If the role starts supporting multiple output declaration for each type, it will be easy to support without changing the interface. (And if you think that this will never be supported, why having separate `name` and `type` at all?)

Hmmm, I thought the value of `name` could be anything that the customers name it in their configuration, on the other hand, the `type` could be restricted in the given types.  But it might have been a wrong impression.

Here's the summary:
1) only one instance per one type of input and output role is allowed.  (the rests are ignored.)
2) name of the input and output role has to be in [ 'basics', 'files', 'forwards' ] (note: 'files' is used for the input, imfile, as well aw for the output, omfile).

1 is a fact, if 2 is acceptable, I can eliminate the type subvariable and use name instead.

Then, the simplest configuration (imjournal + omfile) would look like this:
```
        logging_inputs:
          - name: basics
```
@sradco, @richm, I'd like to have your thoughts, too.  Thanks!

pcahyna: I would keep `type`, I think it expresses the meaning better. Would it be possible to have `name` default to `type` if omitted? What would customers use `name` for, is it reflected in any way in the resulting configuration? (I would think that it is used to construct the resulting configuration snippet name.)

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: > What about having the default only for situations where the user has not supplied any `logging_outputs`? It is quite irregular to have a magic value that gets always appended.

Sorry for my repeating this example...  This is the simplest configuration (imjournal + omfile).  We need some input info (in this case imjournal), but we'd like to allow to specify output to the /var/log/messages, etc. since it's too obvious for the UNIX users...
```
logging_inputs:
   - name: basics
```
> That's a problem with default values for lists in Ansible. it is not easy to have a list with some default values and then append to the list instead of overriding it.

This is going to be identical to the above.  Do you think it's a bad idea?
```
logging_outputs:
   - name: files
logging_inputs:
   - name: basics
```

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: Thank you for finding out the typo... :)  I pushed the patch with the typo fix.
https://github.com/linux-system-roles/logging/pull/100/commits/ca6eb843d284774df75a79bb1cdab2e0958083ce

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: > It could be beneficial to have `name` to make it look more like the interface of other roles, but maybe it could be allowed to omit it and the role would derive it from `type`?

+1!  I will eliminate the `name` key.

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: I will eliminate `name` and keep `type`.
```
        logging_inputs:
          - type: basics
```

design_docs/rsyslog_new_input_output_roles_proposal.md
pcahyna: ??? I did not find any typo, I just explained my reasoning for saying that `logs_collections` was a newly introduced variable in the PR.

design_docs/rsyslog_new_input_output_roles_proposal.md
pcahyna: Can you please answer what is/was `name` used for before eliminating it?

design_docs/rsyslog_new_input_output_roles_proposal.md
pcahyna: Please, what is the answer to
> What would customers use `name` for, is it reflected in any way in the resulting configuration?

richm: Could `name` be used for the `name` field of the `input()`?  If so, might be useful for `impstats` or flow control: https://www.rsyslog.com/doc/v8-stable/configuration/modules/imtcp.html#name

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: @pcahyna, I found a good use case which is a supporter of your original suggestion - using `name` for matching the input and output.  Plus we need to allow multiple instances in the output roles.

In this PR100 design, I did not allow to have multiple output instances to have separate input.  But it turned out it is essential.  This is an old configuration for elasticsearch.
```
logging_outputs:
  - name: viaq-elasticsearch
    type: elasticsearch
    logging_inputs:
      - name: viaq-input
        type: viaq
    server_host: logging-es
    <<snip>>
  - name: ovirt-elasticsearch
    type: elasticsearch
    logging_inputs:
      - name: ovirt-input
        type: ovirt
    server_host: logging-es-ovirt
    <<snip>>
```
So, what I have to do is ...
1) keeping both `name` and `type`.
2) `type` is used for the input/output type - it is the directory name in the input_roles & output_roles.
3) `name` is used for matching the input and the output.

I still have some concerns...
1) The outputs need to allow multiple instances as shown in the elasticsearch example, but the rest of the output roles do not require the ability.  (This is mainly because the files and the forwards output has rsyslog_files_action and rsyslog_forwards_action parameters.)  Is it ok to leave this asymmetry among outputs?
2) Do you think it's acceptable to enforce the `name` value has to start with the `type` value?  Like if it is the `files` type, the `name` should be `files-something`?

Then, the above example would be something like this.  (I added 'files-something' to show what we want to support in this PR.)
```
logging_outputs:
  - name: elasticsearch-viaq
    type: elasticsearch
    server_host: logging-es
    <<snip>>
  - name: elasticsearch-ovirt
    type: elasticsearch
    server_host: logging-es-ovirt
    <<snip>>
  - name: files-something
    type: files
logging_inputs:
   - name: viaq-input
     type: viaq
    rsyslog_outputs: ['elasticsearch-viaq']
  - name: ovirt-input
    type: ovirt
    rsyslog_outputs: ['files-something', 'elasticsearch-ovirt']
```

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: > Could `name` be used for the `name` field of the `input()`? If so, might be useful for `impstats` or flow control: https://www.rsyslog.com/doc/v8-stable/configuration/modules/imtcp.html#name

Yep, it is.  Also, in the action.  I'm not using it in the files action since it's in the old format.  I guess I should update them.  But breaking the elasticsearch use case is more destructive...  (please see https://github.com/linux-system-roles/logging/pull/100#discussion_r403132931)  I'm reworking on it.

nhosoi:
> You should make a note that in the previous style, `logging_inputs ` was actually called `logs_collections`.

+1
I'm putting it in README.md and the proposal.   Also, when this task is done, I'm going to write a changelog.

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: >> Do you think it's acceptable to enforce the name value has to start with the type value? Like if it is the files type, the name should be files-something?
> I don't think this is valuable, but such value for name could be the default if name is not provided.
>
> I am still interested what is name used for currently. Does each output instance under logging_outputs correspond to a configuration file or snippet? If so, is the name attribute used to determine the name of the file? If not, what is it used for?

It is just an identifier to distinguish the action in the state file, for instance.

> Finally, it seems strange to me that there is a rsyslog_outputs attribute. All the variables are prefixed with logging_ and attributes are not namespaced. This indicates that they are in principle generic. (I know that we don't have any other provider than rsyslog currently....) IMO, the attribute should then be called just outputs because it does not seem to be more rsyslog specific than the other attributes.

+1  I'm renaming `rsyslog_outputs` to `outputs`.

I have one ansible / jinja2 question...  I'm using the `rsyslog_use_files_ruleset: true` to make `call files` available in the rsyslog config file like this.
```
          {% if element.rsyslog_use_files_ruleset | d(false) %}
          call files
          {% endif %}
```
If we switch `rsyslog_use_files_ruleset: true|false` to `outputs: ['files-A', 'elasticsearch-B', elasticsearch-C']`, we could do something like this, I guess.  But are there any nice way to avoid to have `call elasticsearch` twice in this example?  Not having `break` always makes me sad...

{% for output in element.outputs %}
  {% if output.startswith("files")
    call files
  {% elif output.startswith("elasticsearch") %}
    call elasticsearch
  {% endif %}
{% endfor %}

I have another place I need to detect the input - output match.  In the last condition in `when` in the next code, I'd like to check logging_inputs which outputs list contains the elasticsearch instance name.  When it's true, I'd like to add the logging_input name to rsyslog_elasticsearch list.  Do you think it works as expected?  (Basically, this is very close to the implementation by @sradco.)
```
- name: Set rsyslog_elasticsearch
  set_fact:
    rsyslog_elasticsearch: "{{ rsyslog_elasticsearch | d([]) }} + {{ [ item.0 | combine( {'logging_input_name': item.1.name} )] }}"
  with_subelements:
    - "{{ logging_outputs }}"
    - logging_inputs
  when:
    - item.0.type | d('') == 'elasticsearch'
    - item.1.state | d('present') == 'present'
    - item.0.name in item.1.outputs
```

pcahyna: > >  I am still interested what is name used for currently. Does each output instance under logging_outputs correspond to a configuration file or snippet? If so, is the name attribute used to determine the name of the file? If not, what is it used for?

> It is just an identifier to distinguish the action in the state file, for instance.

If it is used for something, I would say it is a reason to keep it. Can you please give an example of state file with this identifier?

>  But are there any nice way to avoid to have `call elasticsearch` twice in this example? Not having `break` always makes me sad...
> 
> {% for output in element.outputs %}
> {% if output.startswith("files")
> call files
> {% elif output.startswith("elasticsearch") %}
> call elasticsearch
> {% endif %}
> {% endfor %}

I see `call elasticsearch` only once in the example.

nhosoi: > I see call elasticsearch only once in the example.

Well, since element.outputs contains elasticsearch-B and elasticsearch-C, output.startwith("elasticsearch") is true twice, isn't it?  Then, `call elasticsearch` would be printed in the rsyslog config file twice...

pcahyna: Ah, I see. But you should not use `output.startswith("elasticsearch")` anyway, because in `outputs`, you have names, not types, but the decision whether to include `call elasticsearch` in the file should be done according to the type, not the name. (Assuming that an output is of type `elasticsearch` if and only if it does start with "elasticsearch" does not seem like a good idea.) Therefore, I think you should convert the names to types, and remove duplicates from the list of types, and in the loop iterate over the list of types. Then in this example `call elasticsearch` will be added only once.

richm: If you have a list of dicts, and you need a unique list based on some key of the dict:
```yaml
---
- name: set parameters for test
  hosts: localhost
  tasks:
  - set_fact:
      logging_outputs:
      - name: name-files-foo
        type: files
      - name: name-files-bar
        type: files
      - name: name-fwd-foo
        type: fwd
      - name: name-fwd-bar
        type: fwd

  - debug:
      msg: "uniquelist = {{ logging_outputs | items2dict(key_name='type', value_name='name') }}"
```
If I run this like `ansible-playbook -vv test.yml` I get this output:
```
TASK [set_fact] ****************************************************************
task path: /home/rmeggins/linux-system-roles/kernel_settings/junk-playbook.yml:5
ok: [localhost] => {"ansible_facts": {"logging_outputs": [{"name": "name-files-foo", "type": "files"}, {"name": "name-files-bar", "type": "files"}, {"name": "name-fwd-foo", "type": "fwd"}, {"name": "name-fwd-bar", "type": "fwd"}]}, "changed": false}

task path: /home/rmeggins/linux-system-roles/kernel_settings/junk-playbook.yml:16
ok: [localhost] => {
    "msg": "uniquelist = {'files': 'name-files-bar', 'fwd': 'name-fwd-bar'}"
}
```
so I'm assuming you could do something like this:
```
{% for output in logging_outputs | items2dict(key_name='type', value_name='name') %}
call {{ output }}
{% endfor %}
```
or something like that - not sure if you can use the filter directly in the `for` loop, or if you need to set the result of `items2dict` to some other intermediate variable.

richm: Looks like that works - I added the following task:
```yaml
  - template:
      src: junk.j2
      dest: junk.rsyslog
```
where junk.j2 is 
```
{% for output in logging_outputs | items2dict(key_name='type', value_name='name'
) %}
call {{ output }}
{% endfor %}
```
and junk.rsyslog contains
```
call files
call fwd
```

README.md
```diff
@@ -150,54 +166,81 @@ logging_outputs:
...
+      - `input_type`: Specifying the input type. Type `ovirt` and `viaq` are supported. Default to `ovirt`.
+      - `retryfailures`: Specifying whether retries or not in case of failure. on or off.  Default to on.
       - `ca_cert`: Path to CA cert for ElasticSearch.  Default to '/etc/rsyslog.d/es-ca.crt'
       - `cert`: Path to cert for ElasticSearch.  Default to '/etc/rsyslog.d/es-cert.pem'
       - `key`: Path to key for ElasticSearch.  Default to "/etc/rsyslog.d/es-key.pem"
+   -  ** `type: files`**
+      - `facility`: facility; default to `*`
+      - `severity`: severity; default to `*`
+	  - `exclude`: exclude; default to none.
+	  - `path`: path to the output file.  Must have.  If `path` is not defined, the files instance is dropped.
```
richm, line 225: indentation - I think `path` is supposed to be at the same level as `severity`

README.md
```diff
@@ -150,54 +166,81 @@ logging_outputs:
...
       - `index_prefix`: Elasticsearch index prefix the particular log will be indexed to.
+      - `input_type`: Specifying the input type. Type `ovirt` and `viaq` are supported. Default to `ovirt`.
+      - `retryfailures`: Specifying whether retries or not in case of failure. on or off.  Default to on.
       - `ca_cert`: Path to CA cert for ElasticSearch.  Default to '/etc/rsyslog.d/es-ca.crt'
       - `cert`: Path to cert for ElasticSearch.  Default to '/etc/rsyslog.d/es-cert.pem'
       - `key`: Path to key for ElasticSearch.  Default to "/etc/rsyslog.d/es-key.pem"
+   -  ** `type: files`**
+      - `facility`: facility; default to `*`
+      - `severity`: severity; default to `*`
+	  - `exclude`: exclude; default to none.
```
richm, line 224: should `exclude` be indented here?

README.md
nhosoi: For some reason, 'tab' had sneaked into... :(  I'm checking the other files, too.  Thanks!

roles/rsyslog/handlers/main.yml
```diff
@@ -1,5 +1,10 @@
 ---
-- name: restart rsyslogd
+- name: reset failure
+  listen: restart rsyslogd
+  command: systemctl reset-failed rsyslog
```
nhosoi, line 4: @richm, @pcahyna, I know this is not a right thing to do - we should not call systemctl in the command module directly, but by this resetting the failure in the repeated restarts by the following handler, the CI test failure [1] disappears.
[1]
```
   Apr 13 17:47:43 ibm-p8-kvm-03-guest-02 systemd[1]: rsyslog.service: Start request repeated too quickly.
   Apr 13 17:47:43 ibm-p8-kvm-03-guest-02 systemd[1]: rsyslog.service: Failed with result 'start-limit-hit'.
   Apr 13 17:47:43 ibm-p8-kvm-03-guest-02 systemd[1]: Failed to start System Logging Service. 
```

README.md
```diff
@@ -102,16 +111,28 @@ logging_enabled: true
 logging_purge_confs: true
 ```
 
-2) Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the local system and outputs into the local files.
+2) Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the local system and outputs into the local files, which actions are predefined.
```
pcahyna, line 114: that sounds strange, what does it mean? @richm could you help with improving the wording, please?

design_docs/rsyslog_new_input_output_roles_proposal.md
```diff
@@ -0,0 +1,240 @@
...
+Each output action is in the ruleset which name is the logging_outputs name.
+
+Elasticsearch, Files and Forwards output configuration files in /etc/rsyslog.d:
+```
+  10-output-modules.conf
+  30-output-elasticsearch.conf
+  30-output-files.conf
+  30-output-forwards.conf
+```
+With configuring the outputs (in this example, `files`) with a unique name (e.g., "files_output1", etc.), it will generate an rsyslog configuraion in the ruleset which name is the unique name.
```
pcahyna, line 33: what is the example variable structure to generate this?
In the exam,ples above, you have `name: unique_output_name`, but here the name is `files_output1`. Can you please make the examples consistent? I.e. if you show an example of variables, show later what would be the configuration produced by that very same example?

design_docs/rsyslog_new_input_output_roles_proposal.md
```diff
@@ -0,0 +1,240 @@
...
+```
+ruleset(name="files_output1") {
+  *.info;authpriv.none;auth.none;cron.none;mail.none /var/log/messages
+}
+ruleset(name="files_output2") {
+  *.emerg :omusrmsg:*
+}
```
pcahyna, line 40: would it work to put each of those into a separate configuration file?
I.e. have e.g. `30-output-files_output1.conf` containing
```yaml
ruleset(name="files_output1") {
  *.info;authpriv.none;auth.none;cron.none;mail.none /var/log/messages
}
```
and `30-output-files_output2.conf` containing
```yaml
ruleset(name="files_output2") {
  *.emerg :omusrmsg:*
}
```
This way, each unique `name` corresponds to a separate configuration file. And it is easy to implement `state: absent` by simply removing the corresponding file. But I am not an rsyslog expert and I don't know whether rsyslog would allow this.

README.md
richm: Maybe something like `... handle inputs from the local system (e.g. systemd journald), and outputs into local files (e.g. /var/log/messages).` and omit the `, which actions are predefined.` clause.

README.md
```diff
@@ -102,16 +111,28 @@ logging_enabled: true
...
-        type: basics
+logging_inputs:
+  - name: system-input
+    type: basics
+logging_flows:
+  - name: flow0
+    inputs: [system-input]
+    outputs: [local-files]
+```
+The local files `type files` is a special output.  The predefined file type `logging_outputs` and specifying the logging_flows could be skippable.  So, the following configuration is the same as above.
```
richm, line 129: > The local files `type files` is a special output.

Do you mean that it is the default output, if no output is specified?

> The predefined file type `logging_outputs` and specifying the logging_flows could be skippable.

Do you mean that the default output are files such as `/var/log/messages`?  And the default flow is to write records from `system-input` to the default output?

README.md
nhosoi: Thank you, @richm.  Replacing the line.

README.md
nhosoi: > > The local files `type files` is a special output.
> 
> Do you mean that it is the default output, if no output is specified?

Yes.

> > The predefined file type `logging_outputs` and specifying the logging_flows could be skippable.
> 
> Do you mean that the default output are files such as `/var/log/messages`? And the default flow is to write records from `system-input` to the default output?

Yes.  Correct.

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: Ok.  Let me put an example first and explain based on it.

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: No, there is only one config file per output.  I think it'd be quite difficult to support the multiple files per output due to our usage of the jinja templates...

README.md
richm: Then I think it would be clearer to say something like `If inputs are specified, but no flows or outputs are specified, the default is to write the input to the predefined system log files e.g. /var/log/messages.`

design_docs/rsyslog_new_input_output_roles_proposal.md
pcahyna: So it is not a rsyslog limitation, just a limitation of the role? Why does our use of jinja2 templates make it difficult? It would be so logical to have one file for each `name` ...

jvymazal: From rsyslog POV you can have as many files for output as you want, either put more PRI-FAC simplified rules after each other in the ruleset (the way its done in default config), or use traditional action() statements after the prio/facility specifier if you want to specify more options for the file output. 
You can have it inside one config file, or nest another includes within the ruleset and have it separate each file in its own config file. The second way also means you can chain the outputs, if that is something you would like.

As for the prior question, if you delete the config file defining a ruleset you are calling elsewhere rsyslog will run and sort-of ignore the call, but there will be an error upon startup as rsyslog expects everything that is used to also be defined, so I would not recommend this approach.

nhosoi: Thank you, @pcahyna, @richm, @jvymazal, for your inputs.  Ok, rethinking about the implementation, it may not that hard to generate as many config files per ansible input config vars from customers (but it'll be in the separate pr :).

README.md
nhosoi: Done.

design_docs/rsyslog_new_input_output_roles_proposal.md
nhosoi: Done.

README.md
nhosoi: Done.

nhosoi:
[retest]

nhosoi:
[citest]

nhosoi:
[citest]

README.md
```diff
@@ -279,6 +322,32 @@ can test against a different image/tag like so:
 
   `$ MOLECULE_DISTRO="fedora:30" molecule test`
 
+CI tests
+========
+In the tests directory, currently the following test scripts are available.
+tests_combination.yml    tests_default.yml              tests_files_files.yml              tests_files.yml
+tests_combination2.yml   tests_enabled.yml              tests_files_forwards_implicit.yml  tests_forwards.yml
+tests_default_files.yml  tests_files_elasticsearch.yml  tests_files_forwards.yml           tests_listen.yml
```
richm, line 330: You can omit the list of tests, and just say "The tests are `tests/tests*.yml`" since this list may change in the future and we don't want to have to keep this doc in sync.

roles/rsyslog/roles/input_roles/basics/defaults/main.yml
```diff
@@ -29,45 +27,99 @@ __rsyslog_basics_rules:
...
+          # Read from journald
+          {% set done = "false" %}
+          {% for input in logging_inputs %}
+          {%   if done == "false" and input.type == "basics" %}
+          module(load="imjournal"
+                 StateFile="{{ rsyslog_work_dir }}/imjournal.state"
+                 RateLimit.Burst="{{ input.rsyslog_imjournal_ratelimit_burst | d(20000) }}"
+                 RateLimit.Interval="{{ input.rsyslog_imjournal_ratelimit_interval | d(600) }}"
+                 PersistStateInterval="{{ input.rsyslog_imjournal_persist_state_interval | d(10) }}")
+          {%     set done = "true" %}
```
richm, line 65: Does this actually work?  according to https://jinja.palletsprojects.com/en/2.11.x/templates/#assignments you aren't supposed to do this.  If you need to use the value of `done` later on in the code, use a namespace.  Otherwise, can you use or one of the other loop special variables, like `loop.first` or `loop.last`, listed in the table under https://jinja.palletsprojects.com/en/2.11.x/templates/#for?

nhosoi: It may not be.  When I tested this, I thought it fixed my problem (and it looked it made the CI test pass...)  But in the later coding in other places, I found I had to use namespace.  Fixing this.  Thanks!

roles/rsyslog/roles/input_roles/basics/defaults/main.yml
```diff
@@ -29,45 +27,99 @@ __rsyslog_basics_rules:
...
+          {%   if output.type == "files" %}
+          {%     set ns.files_output = true %}
+          {%   endif %}
+          {% endfor %}
+          {% if not ns.files_output %}
+          {%   set rsyslog_outputs = rsyslog_outputs + [{"name": "files", "type": "files"}] %}
+          {% endif %}
+          {% set indict = {} %}
+          {% set input_list = [] %}
+          {% for input in logging_inputs %}
```
richm, line 80: is `logging_inputs` guaranteed to be set?  Is it set in `defaults/main.yml` or something like that?

nhosoi: Thanks.  I should add it to the defaults file.  (it used to be rsyslog_input_TYPE copied from logging_inputs, where empty list was set if logging_inputs is not given.  But not any more...

roles/rsyslog/roles/input_roles/files/defaults/main.yml
```diff
@@ -64,45 +46,52 @@ __rsyslog_conf_imfile_inputs:
...
+          {% set rsyslog_outputs = logging_outputs | d([]) %}
+          {% for output in rsyslog_outputs %}
+          {%   if output.type == "files" %}
+          {%     set ns.files_output = true %}
+          {%   endif %}
           {% endfor %}
-
+          {% if not ns.files_output %}
+          {%   set rsyslog_outputs = rsyslog_outputs + [{"name": "files", "type": "files"}] %}
           {% endif %}
```
richm, line 63: You don't need the temporary variable here.  You can search the list directly like this:
```
          {% if not rsyslog_outputs|selectattr("type","==","files")|list %}
          {%   set rsyslog_outputs = rsyslog_outputs +[{"name": "files", "type": "files"}] %}
          {% endif %}
```
`selectattr` is horribly documented :-( - found through trial and error
also found through trial and error - you cannot modify `logging_outputs` - the copy `{% set rsyslog_outputs = logging_outputs | d([]) %}` is necessary in order to have something to append to

roles/rsyslog/roles/output_roles/files/defaults/main.yml
```diff
@@ -34,72 +34,49 @@ __rsyslog_conf_files_output_rules:
...
-    path: '{{ rsyslog_config_dir }}'
     sections:
 
       - options: |-
-          {% if rsyslog_files_actions | d([]) %}
-          {% for element in rsyslog_files_actions %}
-          {% if element.name is defined and element.path is defined %}
-          {% if element.severity is defined %}
-          {% if element.facility is defined %}
+          {% if rsyslog_output_files | d([]) and rsyslog_output_files.0.path | d() %}
```
richm, line 40: Is the logic here `if rsyslog_output_files has some elements, and the first element has the path attribute, output the configuration for each element, else output the default configuration`?  Does that mean it is possible for `rsyslog_output_files` to contain 1 element that has no `path` attribute?

Because you could get rid of one level of `if` by doing something like this: https://jinja.palletsprojects.com/en/2.11.x/templates/#for
```
{% for element in rsyslog_output_files | d([]) %}
... output config for each element ...
{% else %}
.... output default config ...
{% endfor %}
```
if you need to loop only for those elements that have a `path` attribute:
```
{% for element in rsyslog_output_files | d([]) | selectattr("path","defined") %}
... output config for each element that has a "path" attribute ...
{% else %}
.... output default config when no list elements have a "path" attribute ...
{% endfor %}
```

tasks/main.yml
```diff
@@ -6,59 +6,41 @@
...
 
-    - name: Set rsyslog_elasticsearch
+    - name: Set rsyslog_output_elasticsearch for the omelasticsearch output
       set_fact:
-        rsyslog_elasticsearch: "{{ rsyslog_elasticsearch | d([]) }} + {{ [ item.0 | combine( {'logs_collections_name': item.1.name} )] }}"
-      with_subelements:
-        - "{{ logging_outputs }}"
-        - logs_collections
+        rsyslog_output_elasticsearch: "{{ rsyslog_output_elasticsearch | d([]) }} + {{ [ item ] }}"
+      loop: "{{ logging_outputs }}"
```
richm, line 12:
```
set_fact:
  rsyslog_output_elasticsearch: "{{ logging_outputs | selectattr('name','defined') | selectattr('type','defined') | selectattr('type','==','elasticsearch') }}"
```

tasks/main.yml
```diff
@@ -6,59 +6,41 @@
...
 
-    - name: Set rsyslog_files_actions
+    - name: Set rsyslog_output_files for the omfile outputs
       set_fact:
-        rsyslog_files_actions: "{{ rsyslog_files_actions | d([]) }} + {{ item.0.rsyslog_files_actions }}"
-      with_subelements:
-        - "{{ logging_outputs }}"
-        - logs_collections
+        rsyslog_output_files: "{{ rsyslog_output_files | d([]) }} + {{ [ item ] }}"
+      loop: "{{ logging_outputs }}"
```
richm, line 20: see above for `rsyslog_output_elasticsearch`:
```
set_fact:
  rsyslog_output_files: "{{ logging_outputs | selectattr('name','defined') | selectattr('type','defined') | selectattr('type','==','files') }}"
```

tasks/main.yml
```diff
@@ -6,59 +6,41 @@
...
 
-    - name: Set rsyslog_forwards_actions
+    - name: Set rsyslog_output_forwards for the omfwd output
       set_fact:
-        rsyslog_forwards_actions: "{{ rsyslog_forwards_actions | d([]) }} + {{ item.0.rsyslog_forwards_actions }}"
-      with_subelements:
-        - "{{ logging_outputs }}"
-        - logs_collections
+        rsyslog_output_forwards: "{{ rsyslog_output_forwards | d([]) }} + {{ [ item ] }}"
+      loop: "{{ logging_outputs }}"
```
richm, line 28: see above

tasks/main.yml
```diff
@@ -6,59 +6,41 @@
...
+        - item.name is defined
+        - item.target is defined
+        - item.type | d('') == 'forwards'
 
     - name: Set rsyslog_outputs
       set_fact:
-        rsyslog_outputs: "{{ rsyslog_outputs | d([]) }} + {{ [ item.type ] }}"
-      with_items: "{{ logging_outputs }}"
-      when: item.type is defined
+        rsyslog_outputs: '{{ rsyslog_output_elasticsearch | d([]) }} + {{ rsyslog_output_files | d([]) }} + {{ rsyslog_output_forwards | d([]) }}'
```
richm, line 36: is `rsyslog_outputs` the same as `logging_outputs`?

roles/rsyslog/roles/output_roles/files/defaults/main.yml
jvymazal: > Is the logic here `if rsyslog_output_files has some elements, and the first element has the path attribute, output the configuration for each element, else output the default configuration`? Does that mean it is possible for `rsyslog_output_files` to contain 1 element that has no `path` attribute?
> ...

Well, I do not know from system roles perspective, but from rsyslog side file output with no path does not make sense and is actually not allowed (you will get an error in log complaining about it, if you write it in config).

roles/rsyslog/roles/input_roles/files/defaults/main.yml
nhosoi: Thanks, @richm!  Verified your proposal.

roles/rsyslog/roles/output_roles/files/defaults/main.yml
nhosoi: This logic - check the first rsyslog_output_files item to have the path or not - supports this simplest configuration.  The files output - writing to the local files - is implicitly configured.  
```
        logging_enabled: true
        logging_inputs:
          - name: basic_input
            type: basics
```
The following input code @richm suggested me to enhance is the exact case.  Here, the files output is not configured, and we add it without a path parameter.  Then, in the files output code, it goes to {% else %} and the default actions - rsyslog.conf in the rpm package - are deployed.
```
          {% if not rsyslog_outputs|selectattr("type","==","files")|list %}
          {%   set rsyslog_outputs = rsyslog_outputs + [{"name": "files", "type": "files"}] %}
          {% endif %}
```

tasks/main.yml
nhosoi: Ok. In this case, we can get rid of loop by using selectattr calls.  It simplifies the code.  I'd like to learn if there could be more benefits like performance?

tasks/main.yml
nhosoi: Basically, yes.  I expect it's a cleaned up version of logging_outputs. :)

tasks/main.yml
richm: >  I'd like to learn if there could be more benefits like performance?

I don't know.

tasks/main.yml
richm: Then can you just do 
```
rsyslog_outputs: "{{ logging_outputs }}"
```
?  I guess I don't understand what you mean by "cleaned up" here - is it that `rsyslog_output_elasticsearch | d([]) }} + {{ rsyslog_output_files | d([]) }} + {{ rsyslog_output_forwards | d([])` is somehow different than `logging_outputs`?

nhosoi: For instance, this removes an elasiticsearch type with no name parameter, doesn't it?
`rsyslog_output_elasticsearch: "{{ logging_outputs | selectattr('name','defined') | selectattr('type','defined') | selectattr('type','==','elasticsearch') | list }}"`

richm: Yes, it does.

Under what conditions is `logging_outputs` not the same as `rsyslog_output_elasticsearch + rsyslog_output_files + rsyslog_output_forwards`?  Is it when the user specifies some bogus values in `logging_outputs`?  Is it when the user specifies some outputs that are not yet supported by the `rsyslog` role?

nhosoi: > Under what conditions is `logging_outputs` not the same as `rsyslog_output_elasticsearch + rsyslog_output_files + rsyslog_output_forwards`? Is it when the user specifies some bogus values in `logging_outputs`? 

I think bogus values would go through.  It just removes `name` is not defined.

> Is it when the user specifies some outputs that are not yet supported by the `rsyslog` role?

Yes, they will be dropped.

nhosoi:
[citest]

richm:
latest commit looks good

nhosoi:
Thank you for reviewing the commit, @richm!

@richm, @pcahyna, This PR is made from 16 commits including obsolete implementations.  Is it ok to leave them as it is for the history when merge them?  Or do you prefer to squash into one or several?

richm:
> Thank you for reviewing the commit, @richm!
> 
> @richm, @pcahyna, This PR is made from 16 commits including obsolete implementations. Is it ok to leave them as it is for the history when merge them? Or do you prefer to squash into one or several?

If the reviews are done (@pcahyna @sradco ?) then please squash into a single commit.

richm:
lgtm

roles/rsyslog/roles/output_roles/files/defaults/main.yml
```diff
@@ -34,73 +34,50 @@ __rsyslog_conf_files_output_rules:
...
+              cron.*                                                  -{{ rsyslog_system_log_dir }}/cron
+
+              # Everybody gets emergency messages
+              *.emerg                                                  :omusrmsg:*
+
+              # Save news errors of level crit and higher in a special file.
+              uucp,news.crit                                          {{ rsyslog_system_log_dir }}/spooler
+
+              # Save boot messages also to boot.log
+              local7.*                                                {{ rsyslog_system_log_dir }}/boot.log
```
pcahyna, line 77: I hope this huge inline of a config file will get changed in some subsequent PR

nhosoi: > I hope this huge inline of a config file will get changed in some subsequent PR

Yes, this is going to be in a template file in the following pr.  (I.e.,  https://github.com/nhosoi/logging/pull/10, which will be submitted to this repo once this pr is merged.)

richm:
squash lgtm

nhosoi:
Thank you @richm.  Finally, the CI tests have finished.  Merging...

PR #101, opened 2020-04-01
do not use json_query to test for rsyslog package

richm:
We can just use the `in` test to see if `rsyslog` is
in `ansible_facts.packages` - we do not have to use
`json_query`.  I believe this also gets rid of the
dependency on `jmespath`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ install:
   # workaround until https://github.com/ansible/molecule/issues/1727 is fixed
   - pip uninstall -y testinfra
   - pip install testinfra
-  - pip install jmespath
 script:
   - molecule test
 branches:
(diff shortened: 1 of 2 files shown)


nhosoi:
Thanks, @richm!  (I did not know it...)
LGTM

richm:
I don't think this change has caused the fedora 31 failure which seems unrelated but retesting anyway
[citest bad]

pcahyna:
> I believe this also gets rid of the
> dependency on `jmespath`

It does, but there are other places in system roles where json_query needs to be used. Anyway, I like the change just because of the simplicity it brings.

richm:
> > I believe this also gets rid of the
> > dependency on `jmespath`
> 
> It does, but there are other places in system roles where json_query needs to be used. Anyway, I like the change just because of the simplicity it brings.

I haven't found a place yet where we needed to have
```
install:
  - pip install jmespath
```
in `.travis.yml` - if you find such a place, we will need to add it to the template

pcahyna:
I suppose it is an accident. Only the default test (which usually does not do much) is being executed by Molecule and in other roles the code path executed by the default test probably does not hit the json_query filter.

PR #103, opened 2020-04-16
do not use ../ in path when including roles

nhosoi:
For some reason, when including parent roles from a sub-role
like this:
```yaml
  include_role:
    name: "{{ role_path }}/../../../"
    tasks_from: deploy.yml
```
and there are handlers to be included and executed from the included
role, ansible thinks the handler tasks are separate, and executes
them separately, leading to a problem that the same handler is
executed multiple times in a row.  Using the role path without
`../` allows ansible to correctly determine that the included
handlers are all the same task, and only executes them once.

### DIFF
--- a/roles/rsyslog/roles/input_roles/basics/tasks/main.yml
+++ b/roles/rsyslog/roles/input_roles/basics/tasks/main.yml
@@ -5,5 +5,5 @@
     __rsyslog_packages: "{{ __rsyslog_basics_packages }}"
     __rsyslog_rules: "{{ __rsyslog_basics_rules }}"
   include_role:
-    name: "{{ role_path }}/../../../"
+    name: "{{ role_path | dirname | dirname | dirname }}"
     tasks_from: deploy.yml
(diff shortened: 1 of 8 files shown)


nhosoi:
@richm, sadly this change (using `dirname` instead of `../`) did not solve our problem.
The failure of starting restarting occurs when more than or equal to 4 subroles (input roles + output roles) are involved.  The following is the systemctl status rsyslog output when tests_combination.yml in https://github.com/linux-system-roles/logging/pull/100/ failed.
```
Apr 17 16:17:16 ibm-p8-kvm-03-guest-02 systemd[1]: rsyslog.service: Start request repeated too quickly.
Apr 17 16:17:16 ibm-p8-kvm-03-guest-02 systemd[1]: rsyslog.service: Failed with result 'start-limit-hit'.
Apr 17 16:17:16 ibm-p8-kvm-03-guest-02 systemd[1]: Failed to start System Logging Service.
```
Counting "RUNNING HANDLER" in the output from the tests_combination.yml run on Fedora-31, we see extra 4 restart rsyslogd from the subroles.
```
$ egrep "RUNNING HANDLER" /tmp/combination31.out
RUNNING HANDLER [/home/nhosoi/linux-system-roles/logging/roles/rsyslog : restart rsyslogd] ****************************************
RUNNING HANDLER [/home/nhosoi/linux-system-roles/logging/roles/rsyslog : restart rsyslogd] ****************************************
RUNNING HANDLER [/home/nhosoi/linux-system-roles/logging/roles/rsyslog : restart rsyslogd] ****************************************
RUNNING HANDLER [/home/nhosoi/linux-system-roles/logging/roles/rsyslog : restart rsyslogd] ****************************************

```

PR #106, opened 2020-04-20
Workaround to execute "RUNNING HANDLER" from include_role just once.

nhosoi:
Add a condition to restart rsyslogd handler when __rsyslogd_restarted is false.
Once the handler is executed, it sets __rsyslogd_restarted to true.

Revert "do not use ../ in path when including roles 3"
This reverts commit ff5fd47c8df1f5531e775ee8d98a6fde2bad0cf6.

### DIFF
--- a/roles/rsyslog/handlers/main.yml
+++ b/roles/rsyslog/handlers/main.yml
@@ -3,3 +3,10 @@
   service:
     name: rsyslog
     state: restarted
+  when: not __rsyslogd_restarted | d(false)
+  listen: 'restart rsyslogd'
+
+- name: set rsyslogd restarted
+  set_fact:
+    __rsyslogd_restarted: true
+  listen: 'restart rsyslogd'
--- a/roles/rsyslog/roles/input_roles/basics/tasks/cleanup.yml
+++ b/roles/rsyslog/roles/input_roles/basics/tasks/cleanup.yml
@@ -4,5 +4,5 @@
   vars:
     __rsyslog_rules: "{{ __rsyslog_basics_rules }}"
   include_role:
-    name: "{{ role_path | dirname | dirname | dirname ~ '/rsyslog' }}"
+    name: "{{ role_path | dirname | dirname | dirname }}"
     tasks_from: cleanup.yml
--- a/roles/rsyslog/roles/input_roles/basics/tasks/main.yml
+++ b/roles/rsyslog/roles/input_roles/basics/tasks/main.yml
@@ -5,5 +5,5 @@
     __rsyslog_packages: "{{ __rsyslog_basics_packages }}"
     __rsyslog_rules: "{{ __rsyslog_basics_rules }}"
   include_role:
-    name: "{{ role_path | dirname | dirname | dirname ~ '/rsyslog' }}"
+    name: "{{ role_path | dirname | dirname | dirname }}"
     tasks_from: deploy.yml
--- a/roles/rsyslog/roles/input_roles/files/tasks/cleanup.yml
+++ b/roles/rsyslog/roles/input_roles/files/tasks/cleanup.yml
@@ -4,5 +4,5 @@
   vars:
     __rsyslog_rules: "{{ __rsyslog_files_input_rules }}"
   include_role:
-    name: "{{ role_path | dirname | dirname | dirname ~ '/rsyslog' }}"
+    name: "{{ role_path | dirname | dirname | dirname }}"
     tasks_from: cleanup.yml
--- a/roles/rsyslog/roles/input_roles/files/tasks/main.yml
+++ b/roles/rsyslog/roles/input_roles/files/tasks/main.yml
@@ -5,5 +5,5 @@
     __rsyslog_packages: "{{ __rsyslog_files_input_packages }}"
     __rsyslog_rules: "{{ __rsyslog_files_input_rules }}"
   include_role:
-    name: "{{ role_path | dirname | dirname | dirname ~ '/rsyslog' }}"
+    name: "{{ role_path | dirname | dirname | dirname }}"
     tasks_from: deploy.yml
--- a/roles/rsyslog/roles/input_roles/ovirt/tasks/cleanup.yml
+++ b/roles/rsyslog/roles/input_roles/ovirt/tasks/cleanup.yml
@@ -4,5 +4,5 @@
   vars:
     __rsyslog_rules: "{{ __rsyslog_ovirt_rules }}"
   include_role:
-    name: "{{ role_path | dirname | dirname | dirname ~ '/rsyslog' }}"
+    name: "{{ role_path | dirname | dirname | dirname }}"
     tasks_from: cleanup.yml
--- a/roles/rsyslog/roles/input_roles/ovirt/tasks/main.yml
+++ b/roles/rsyslog/roles/input_roles/ovirt/tasks/main.yml
@@ -5,7 +5,7 @@
     __rsyslog_packages: "{{ __rsyslog_ovirt_prereq_packages + __rsyslog_ovirt_packages }}"
     __rsyslog_rules: "{{ __rsyslog_ovirt_rules }}"
   include_role:
-    name: "{{ role_path | dirname | dirname | dirname ~ '/rsyslog' }}"
+    name: "{{ role_path | dirname | dirname | dirname }}"
     tasks_from: deploy.yml
 
 - name: Allow rsyslog to listen on collectd port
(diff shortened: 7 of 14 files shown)


richm:
lgtm - @pcahyna ?

roles/rsyslog/handlers/main.yml
```diff
@@ -3,3 +3,10 @@
   service:
     name: rsyslog
     state: restarted
+  when: not __rsyslogd_restarted | d(false)
+  listen: 'restart rsyslogd'
+
+- name: set rsyslogd restarted
+  set_fact:
+    __rsyslogd_restarted: true
```
pcahyna, line 11: how will this work if the role gets called multiple times in the same playbook?

pcahyna:
I believe it would be better to get rid of the "call-back" to the parent role instead of inventing more and more workarounds. Not sure how difficult would that be though...

roles/rsyslog/handlers/main.yml
richm: > how will this work if the role gets called multiple times in the same playbook?

I don't think it will work, so we will need a workaround for the workaround e.g. reset `__rsyslogd_restarted: false` somehow

richm:
> I believe it would be better to get rid of the "call-back" to the parent role instead of inventing more and more workarounds. Not sure how difficult would that be though...

Right.  IMO this PR is just a temporary patch to the larger problem of refactoring the roles not to have  the recursive role calling.  Solving the latter problem will take some time, and in the meantime, this PR gives us at least a working CI with which we can rebase and very no regressions caused by rebasing.

nhosoi:
> > I believe it would be better to get rid of the "call-back" to the parent role instead of inventing more and more workarounds. Not sure how difficult would that be though...
> 
> Right. IMO this PR is just a temporary patch to the larger problem of refactoring the roles not to have the recursive role calling. Solving the latter problem will take some time, and in the meantime, this PR gives us at least a working CI with which we can refactor and verify no regressions caused by refactoring.

Maybe...  Do you have any idea from where I should start?  From scratch?  Getting rid of the call-back but we could keep the input_/output_roles?

richm:
> > > I believe it would be better to get rid of the "call-back" to the parent role instead of inventing more and more workarounds. Not sure how difficult would that be though...
> > 
> > 
> > Right. IMO this PR is just a temporary patch to the larger problem of refactoring the roles not to have the recursive role calling. Solving the latter problem will take some time, and in the meantime, this PR gives us at least a working CI with which we can refactor and verify no regressions caused by refactoring.
> 
> Maybe... Do you have any idea from where I should start? From scratch? Getting rid of the call-back but we could keep the input_/output_roles?

I'm not sure - I initially thought if we had a "flat" role structure - all roles in the same level under `rsyslog/roles` - and a `rsyslog/roles/common` role for common things like handlers - then the "call stack" would have no loops e.g. `logging -> rsyslog -> input_roles -> rolename -> common`

but https://github.com/ansible/ansible/issues/47374 seems to be more or less flat and exhibits the problem - so I'm not sure how to get around the problem with a refactoring, unless we get rid of nested roles completely . . .

pcahyna:
> I'm not sure - I initially thought if we had a "flat" role structure - all roles in the same level under `rsyslog/roles` - and a `rsyslog/roles/common` role for common things like handlers - then the "call stack" would have no loops e.g. `logging -> rsyslog -> input_roles -> rolename -> common`

That's a good point. I also thought that it would be fixed by eliminating indirect recursion, but thinking more about the problem, it probably won't.

I also thought that it would be easy to replace the code with something less convoluted, given that `deploy.yml` and `cleanup.yml` are very short task files. But now I don't see an easy and at the same time generic way to do it. So proceeding with this workaround is the best thing to do, if it helps with CI testing.

nhosoi:
> > I'm not sure - I initially thought if we had a "flat" role structure - all roles in the same level under `rsyslog/roles` - and a `rsyslog/roles/common` role for common things like handlers - then the "call stack" would have no loops e.g. `logging -> rsyslog -> input_roles -> rolename -> common`
> 
> That's a good point. I also thought that it would be fixed by eliminating indirect recursion, but thinking more about the problem, it probably won't.
> 
> I also thought that it would be easy to replace the code with something less convoluted, given that `deploy.yml` and `cleanup.yml` are very short task files. But now I don't see an easy and at the same time generic way to do it. So proceeding with this workaround is the best thing to do, if it helps with CI testing.

Of course, not in this PR :), but could making this change ease other logging role path issues?  If so, it'd be worth updating, I think.
`logging -> rsyslog -> input_roles -> rolename -> common`

>> how will this work if the role gets called multiple times in the same playbook?
> I don't think it will work, so we will need a workaround for the workaround e.g. reset __rsyslogd_restarted: false somehow

In terms of this issue, I thought the handler runs and is supposed to run at the end of the play.  Is it wrong?  Sorry, I haven't understood the proposal fully... :(

If such a condition is not a good idea, we have this uglier workaround, too...
```
   diff --git a/roles/rsyslog/handlers/main.yml
   b/roles/rsyslog/handlers/main.yml
   index 367746b..58fe2cc 100644
   --- a/roles/rsyslog/handlers/main.yml
   +++ b/roles/rsyslog/handlers/main.yml
   @@ -1,5 +1,9 @@
     ---
   +- name: reset failure
   +  command: systemctl reset-failed rsyslog.service
   +  listen: "restart rsyslogd"
```

pcahyna:
> > > I'm not sure - I initially thought if we had a "flat" role structure - all roles in the same level under `rsyslog/roles` - and a `rsyslog/roles/common` role for common things like handlers - then the "call stack" would have no loops e.g. `logging -> rsyslog -> input_roles -> rolename -> common`
> > 
> > 
> > That's a good point. I also thought that it would be fixed by eliminating indirect recursion, but thinking more about the problem, it probably won't.
> > I also thought that it would be easy to replace the code with something less convoluted, given that `deploy.yml` and `cleanup.yml` are very short task files. But now I don't see an easy and at the same time generic way to do it. So proceeding with this workaround is the best thing to do, if it helps with CI testing.
> 
> Of course, not in this PR :), but could making this change ease other logging role path issues? If so, it'd be worth updating, I think.
> `logging -> rsyslog -> input_roles -> rolename -> common`

This would help with the other issue you found, yes.

> > > how will this work if the role gets called multiple times in the same playbook?
> > > I don't think it will work, so we will need a workaround for the workaround e.g. reset __rsyslogd_restarted: false somehow
> 
> In terms of this issue, I thought the handler runs and is supposed to run at the end of the play. Is it wrong? Sorry, I haven't understood the proposal fully... :(

Until the end of the play or `meta: flush_handlers` is called. I suppose that if one wants to call the role multiple times, one could want to separate the calls by `meta: flush_handlers` to make the calls independent.
Also, what if one calls the role in multiple plays in the same playbook? I believe facts persist from one play to another, don't they?

nhosoi:
> This would help with the other issue you found, yes.

+1 Thanks!

> Until the end of the play or `meta: flush_handlers` is called. I suppose that if one wants to call the role multiple times, one could want to separate the calls by `meta: flush_handlers` to make the calls independent.

I see. Yes, that's what I am doing in some of the logging role CI tests.

> Also, what if one calls the role in multiple plays in the same playbook? I believe facts persist from one play to another, don't they?

Hmmm, then shall we set `__rsyslogd_restarted: false` at the end of the logging role?  That way if the handler had been notified, it will be called and executed at the end of the logging role?

nhosoi:
> > Also, what if one calls the role in multiple plays in the same playbook? I believe facts persist from one play to another, don't they?
> 
> Hmmm, then shall we set `__rsyslogd_restarted: false` at the end of the logging role? That way if the handler had been notified, it will be called and executed at the end of the logging role?

@pcahyna, do you think `__rsyslogd_restarted: false` should be appended to this pr?  Or it does not help at all?  If it's supposed to be added, I'm updating this pr.

nhosoi:
Just to make the CI test run again, merging this PR.
NOTE: Need to revisit the multiple handler call issue #41560

pcahyna:
> Hmmm, then shall we set `__rsyslogd_restarted: false` at the end of the logging role? That way if the handler had been notified, it will be called and executed at the end of the logging role?

@nhosoi If you do that, you need to flush handlers before setting it, otherwise the handler will be called after you set the variable to false, and will not do the right thing.

nhosoi:
> > Hmmm, then shall we set `__rsyslogd_restarted: false` at the end of the logging role? That way if the handler had been notified, it will be called and executed at the end of the logging role?
> 
> @nhosoi If you do that, you need to flush handlers before setting it, otherwise the handler will be called after you set the variable to false, and will not do the right thing.

@pcahyna, Ok.  I merged this PR without adding the reset code.  So, this is just for my education (I don't plan to make any further changes for this issue :), but to do the right thing, where would be the right places to reset the flag?  Every time we call the notify handler?

PR #107, opened 2020-04-24
Fixing the inaccurate task names in the sub roles.

nhosoi:


nhosoi:
Thanks for reviewing the pr, @richm.  Merging...

PR #110, opened 2020-05-01
RHELPLAN-41773 - Logging - Best Practices - rsyslog should be set to …

nhosoi:
…logging_provider instead of logging_collector

Replacing logging_collector with logging_provider.

Note: Removing the default filter from "when: logging_collector | d('rsyslog') == 'rsyslog'"
in tasks/main.yml causes an error while evaluating conditional (logging_provider == 'rsyslog'):
'logging_provider' is undefined, unless including defaults/main.yml in roles/rsyslog/tasks/main.yml.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -69,12 +69,13 @@
         - item.key is match("rsyslog*")
         - logging_debug | d(false)
 
-    - name: Include Rsyslog role
-      vars:
-        __rsyslog_enabled: "{{ logging_enabled }}"
-        __rsyslog_purge_original_conf: "{{ logging_purge_confs }}"
-        __rsyslog_system_log_dir: "{{ logging_system_log_dir }}"
-      include_role:
-        name: "{{ role_path }}/roles/rsyslog"
+  when: logging_provider == 'rsyslog'
 
-  when: logging_collector | d('rsyslog') == 'rsyslog'
+- name: Include Rsyslog role
+  vars:
+    __rsyslog_enabled: "{{ logging_enabled }}"
+    __rsyslog_purge_original_conf: "{{ logging_purge_confs }}"
+    __rsyslog_system_log_dir: "{{ logging_system_log_dir }}"
+  include_role:
+    name: "{{ role_path }}/roles/rsyslog"
+  when: logging_provider == 'rsyslog'
--- a/README.md
+++ b/README.md
@@ -201,7 +201,6 @@ For more details, see also [roles/rsyslog/README.md](https://github.com/linux-sy
 Variables in vars.yml
 ----------------------
 
-- `logging_collector`: The logs collector to use for the logs collection. Currently Rsyslog is the only supported logs collector. Defaults to `rsyslog`. WARNING: this is going to be renamed to logging_provider.
 - `logging_enabled` : When 'true', logging role will deploy specified configuration file set. Default to 'true'.
 - `logging_mmk8s_token`: Path to token for kubernetes.  Default to "/etc/rsyslog.d/mmk8s.token".
 - `logging_mmk8s_ca_cert`: Path to CA cert for kubernetes.  Default to "/etc/rsyslog.d/mmk8s.ca.crt".
@@ -241,6 +240,7 @@ Variables in vars.yml
           `type`: The type of the pre-configured logs to collect. **Note:** Currently ['viaq', 'viaq-k8s', 'ovirt'] are supported for the elasticsearch output.
           `state`: The state of the configuration files states if they should be `present` or `absent`. Default to `present`.
 
+- `logging_provider`: The logging collector to use for the logging provider. Currently Rsyslog is the only supported logging provider. Defaults to `rsyslog`.
 - `logging_purge_confs`: By default, the Rsyslog configuration files are applied on top of pre-existing configuration files. To purge local files prior to setting new ones, set logging_purge_confs variable to 'true', it will move all Rsyslog configuration files to a backup directory, `/tmp/rsyslog.d-XXXXXX/backup.tgz`, before deploying the new configuration files. Defaults to 'false'.
 - `logging_system_log_dir`: System log directory.  Default to '/var/log'.
 
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,6 +1,6 @@
 ---
-# Default logging collector is Rsyslog
-logging_collector: rsyslog
+# Default logging provider is Rsyslog
+logging_provider: rsyslog
 
 # Unless logging_enabled is set to true, logging role is no-op
 logging_enabled: true


roles/rsyslog/tasks/main.yml
```diff
@@ -1,3 +1,7 @@
+- name: Include top vars
+  include_vars:
+    file: "{{ role_path | dirname | dirname }}/defaults/main.yml"
```
richm, line 3: Did we discuss this before?  That using `include_vars` makes the vars defined in the file a different precedence than `defaults`?  https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#ansible-variable-precedence - role defaults have a much lower precedence than anything else - not sure if it makes a difference here.  See also https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html#role-default-variables

The important thing is that the user can, with an inventory, or with `-e var` or `-e@vars.yml`, override any of the vars defined in `defaults/main.yml`.  But the "Using Variables" page says this:
"If you are writing a role and want to ensure the value in the role is absolutely used in that role, and is not going to be overridden by inventory, you should put it in roles/x/vars/main.yml like so, and inventory values cannot override it. -e however, still will"

I guess the problem is that the variables defined in `defaults/vars.yml` are not available in all contexts in the child roles?  For example, with the current code, if I don't include the vars, and remove the `| d('rsyslog')` below, I get this:
```
RUNNING HANDLER [/home/rmeggins/linux-system-roles/logging/tests/roles/linux-system-roles.logging/roles/rsyslog : restart rsyslogd] ***
task path: /home/rmeggins/linux-system-roles/logging/roles/rsyslog/handlers/main.yml:2
fatal: [/var/lib/libvirt/images/f31.qcow2]: FAILED! => {}

MSG:

The conditional check 'logging_collector == 'rsyslog'' failed. The error was: error while evaluating conditional (logging_collector == 'rsyslog'): 'logging_collector' is undefined
```

richm: I did a google search for `ansible role default variables defined in parent role are not defined in subroles`
This is what I found: https://www.samdarwin.com/ansible-default-vars/ - looks very hacky, but seems to work . . .

@sradco - ovirt uses nested roles quite heavily - but even with this logging role that is much simpler than the whole of ovirt ansible, we have run into several issues - has ovirt run into issues like these with the use of nested roles?  I'm wondering why we keep running into issues like this, and if we should just assume ansible doesn't like nested roles, and rewrite the logging role not to use them . . .

nhosoi: @richm, not sure if I understood the hack correctly, but with this [commit](2790ba189840d890fa984f4fd6e0f46cb061eb9f), I could verify that logging_provider could be overriden if it's set in the command line with -e as well as in the inventory.

richm: > @richm, not sure if I understood the hack correctly, but with this [commit](2790ba189840d890fa984f4fd6e0f46cb061eb9f), I could verify that logging_provider could be overriden if it's set in the command line with -e as well as in the inventory.

Ok - if it is working as-is, then let's go with it.

richm: ah, I see - never mind - are you able to override the defaults with inventory with commit https://github.com/linux-system-roles/logging/pull/110/commits/15d37bbf9560e6c59b8e3bd1832e9f3c4089b3f3 ?

nhosoi: > ah, I see - never mind - are you able to override the defaults with inventory with commit [15d37bb](https://github.com/linux-system-roles/logging/commit/15d37bbf9560e6c59b8e3bd1832e9f3c4089b3f3) ?

@richm, Yes, I was.  But rethinking about this, probably, we should be able to convert the rsyslog role into a task/var pair (e.g., logging/{tasks,vars}/rsyslog/main.yml) as we did for the call-back?  If we could, it'd reduce the future potential headaches...  Let me play with it a bit.

richm: Ok - but I wouldn't worry about it too much, as long as with https://github.com/linux-system-roles/logging/commit/15d37bbf9560e6c59b8e3bd1832e9f3c4089b3f3 you can override the defaults with settings from your inventory.

nhosoi: All right. Then, I won't spend too much time on it :). Thanks!

nhosoi: I tried to converting the include_role rsyslog to include_tasks/vars, but the attempt failed since I could not figure out how to include the defaults file. If I put the default variables in the included vars file, the variable could not be overridden. :(

So, I have to give up the idea and live with this PR...

richm: > I tried to converting the include_role rsyslog to include_tasks/vars, but the attempt failed since I could not figure out how to include the defaults file. If I put the default variables in the included vars file, the variable could not be overridden. :(
> 
> So, I have to give up the idea and live with this PR...

ok, I think that's fine for now

nhosoi: @richm, @pcahyna, can I consider the review is done for this PR? Thanks!

roles/rsyslog/tasks/main.yml
```diff
@@ -1,3 +1,7 @@
+- name: Include top vars
+  include_vars:
+    file: "{{ role_path | dirname | dirname }}/vars/main.yml"
```
pcahyna, line 3: What's the point of this change? `vars/main.yml` gets included automatically in any role.

tasks/main.yml
```diff
@@ -77,4 +77,4 @@
       include_role:
         name: "{{ role_path }}/roles/rsyslog"
 
-  when: logging_collector | d('rsyslog') == 'rsyslog'
+  when: logging_provider | d(default_logging_provider) == 'rsyslog'
```
pcahyna, line 80: why the `d(...)` ? `logging_provider` is set in defaults, so it should be always defined.

pcahyna:
> Removing the default filter from "when: logging_collector | d('rsyslog') == 'rsyslog'"
> in tasks/main.yml causes an error while evaluating conditional (logging_provider == 'rsyslog'):
> 'logging_provider' is undefined, unless including defaults/main.yml in roles/rsyslog/tasks/main.yml.

I don't see any mention of `defaults/main.yml` in https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/tasks/main.yml . Anyway, where exactly does it fail and how?

tasks/main.yml
nhosoi: > why the `d(...)` ? `logging_provider` is set in defaults, so it should be always defined.

This is the point.  Since we do include_role above at the line 78, logging_provider is _not_ found in the included role.

```
task path: /home/nhosoi/linux-system-roles/logging/roles/rsyslog/handlers/main.yml:2
fatal: [/home/nhosoi/Downloads/CentOS-Stream-GenericCloud-8-20200113.0.x86_64.qcow2]: FAILED! => {
    "msg": "The conditional check 'logging_provider == 'rsyslog'' failed. The error was: error while evaluating conditional (logging_provider == 'rsyslog'): 'logging_provider' is undefined\n\nThe error appears to be in '/home/nhosoi/linux-system-roles/logging/roles/rsyslog/handlers/main.yml': line 2, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- name: restart rsyslogd\n  ^ here\n"
}
META: ran handlers
```
And please note that I tried to convert the include_role with include_tasks, but it did not work well. (See also https://github.com/linux-system-roles/logging/pull/110#discussion_r419182270)

roles/rsyslog/tasks/main.yml
nhosoi: Without this include_vars, it fails as follows. And default_logging_provide is needed as described in https://github.com/linux-system-roles/logging/pull/110#discussion_r419597869.
```
task path: /home/nhosoi/linux-system-roles/logging/roles/rsyslog/handlers/main.yml:2
fatal: [/home/nhosoi/Downloads/CentOS-Stream-GenericCloud-8-20200113.0.x86_64.qcow2]: FAILED! => {
    "msg": "The conditional check 'logging_provider | d(default_logging_provider) == 'rsyslog'' failed. The error was: error while evaluating conditional (logging_provider | d(default_logging_provider) == 'rsyslog'): 'default_logging_provider' is undefined\n\nThe error appears to be in '/home/nhosoi/linux-system-roles/logging/roles/rsyslog/handlers/main.yml': line 2, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- name: restart rsyslogd\n  ^ here\n"
}

```

nhosoi:
> > Removing the default filter from "when: logging_collector | d('rsyslog') == 'rsyslog'"
> > in tasks/main.yml causes an error while evaluating conditional (logging_provider == 'rsyslog'):
> > 'logging_provider' is undefined, unless including defaults/main.yml in roles/rsyslog/tasks/main.yml.
> 
> I don't see any mention of `defaults/main.yml` in https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/tasks/main.yml . Anyway, where exactly does it fail and how?

Please see the error in https://github.com/linux-system-roles/logging/pull/110#discussion_r419586944.

tasks/main.yml
pcahyna: Well, you should have said that the problem is in a handler, this makes a bit more sense. Still quite surprising though because the handler does not actually use `logging_provider`. Let me experiment a bit.

nhosoi: > Well, you should have said that the problem is in a handler, ...

Sorry, @pcahyna. I tried including the var in the handler, but it did not work, either... (;_;)

pcahyna: Try this workaround:
```yaml
  - name: Include Rsyslog role
    vars:
      __rsyslog_enabled: "{{ logging_enabled }}"
      __rsyslog_purge_original_conf: "{{ logging_purge_confs }}"
      __rsyslog_system_log_dir: "{{ logging_system_log_dir }}"
    include_role:
      name: "{{ role_path }}/roles/rsyslog"
    when: logging_provider == 'rsyslog'
```
I.e. move the `include_role` out of the `block`. For some reason it is the `block` statement that causes the problem if enclosing the `include_role` statement.

nhosoi: Thank you, @pcahyna.  I've updated the PR.

richm:
lgtm

nhosoi:
Thank you for the reviews and the advice, @richm, @pcahyna.  Merging the pr.

PR #111, opened 2020-05-02
RHELPLAN-41563 - Logging - splitting an input/output config file into…

nhosoi:
… multiple ones that contains just one operation 2

Sharing one input_template.j2 between the basics and files inputs.


roles/rsyslog/tasks/inputs/basics/main.yml
```diff
@@ -9,9 +9,12 @@
 
 - name: Create basics input configuration files in /etc/rsyslog.d
   vars:
+    __line_0: ""
```
richm, line 12: please rename this to something like `__rsyslog_input_decl`

roles/rsyslog/tasks/inputs/basics/main.yml
```diff
@@ -9,9 +9,12 @@
...
+    __my_name: "{{ basics_item.name }}"
+    __my_type: "{{ basics_item.type }}"
```
richm, line 14: please use `__rsyslog_input_name` and `__rsyslog_input_type`

richm:
other than a couple of minor nits, looks good

nhosoi:
Thanks for your reviews and advice, @richm.  Merging the PR...

PR #114, opened 2020-05-08
Changing the exclude value from string to list.

nhosoi:
Files and forwards output uses a filter
  `<facility>.<severity>;<exclude0>;<exclude1>;... <output>`
E.g.,
  `*.info;authpriv.none;cron.none    /var/log/messages`
Files and forwards logging_output takes exclude subvariable, which used to be
  `exclude: <exclude0>;<exclude1>;...`
but by changing it to
  `exclude: ;<exclude0>;<exclude1>;...`
by adding `;` at the first column, `output_{files,forwards}.j2` template could
be simplified as follows.
  `{{ files_item.facility | d('*') }}.{{ files_item.severity | d('*') }}{{ files_item.exclude | d() }} {{ files_item.path }}`

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,11 +219,11 @@ Variables in vars.yml
    -  ** `type: files`**
       - `facility`: facility; default to `*`
       - `severity`: severity; default to `*`
-      - `exclude`: exclude; default to none.
+      - `exclude`: exclude list; default to none.
       - `path`: path to the output file.  Must have.  If `path` is not defined, the files instance is dropped.
       These values are used in the omfile action as follows:
       ```
-      facility.severity;exclude path
+      facility.severity;<semicolon separated exclude list> path
       ```
    -  ** `type: forwards`**
       - `facility`: facility; default to `*`
--- a/design_docs/rsyslog_inputs_outputs_flows.md
+++ b/design_docs/rsyslog_inputs_outputs_flows.md
@@ -40,7 +40,7 @@ logging_outputs
   - name: files_output
     type: files
     severity: info
-    exclude: authpriv.none;auth.none;cron.none;mail.none
+    exclude: [authpriv.none, auth.none, cron.none, mail.none]
     path: /var/log/messages
 ```
 
--- a/roles/rsyslog/README.md
+++ b/roles/rsyslog/README.md
@@ -110,14 +110,42 @@ logging_flows:
 ```
 If inputs are specified, but no flows or outputs are specified, the default is to write the input to the predefined system log files e.g. /var/log/messages.
 ```
-logging_enabled: true
 logging_purge_confs: true
 logging_inputs:
   - name: system-input
     type: basics
 ```
 
-**3. Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the local system and remote rsyslog and outputs into the local files.**
+**3. Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the local system and outputs into the local files, which each acrion is defined.
+```
+logging_purge_confs: true
+rsyslog_backup_dir: /tmp/rsyslog_backup
+logging_outputs:
+  - name: files_output0
+    type: files
+    severity: info
+    exclude:
+      - authpriv.none
+      - auth.none
+      - cron.none
+      - mail.none
+    path: /var/log/messages
+  - name: files_output1
+    type: files
+    severity: emerg
+    path: :omusrmsg:*
+  - name: files_output2
+    <<snip>>
+logging_inputs:
+  - name: system-input
+    type: basics
+logging_flows:
+  - name: flow0
+    inputs: [system-input]
+    outputs: [files_output0, files_output1, files_output2, ...]
+```
+
+**4. Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the local system and remote rsyslog and outputs into the local files.**
 ```
 logging_enabled: true
 rsyslog_capabilities: [ 'network', 'remote-files' ]
@@ -134,7 +162,7 @@ logging_flows:
     outputs: [local-files]
 ```
 
-**4. Deploying basic LSR/Logging config files in /etc/rsyslog.d, which forwards the local system logs to the remote rsyslog.
+**5. Deploying basic LSR/Logging config files in /etc/rsyslog.d, which forwards the local system logs to the remote rsyslog.
 ```
 logging_enabled: true
 rsyslog_default: false
@@ -161,7 +189,7 @@ logging_flows:
     outputs: [output-forwards0, output-forwards1]
 ```
 
-**5. Sample vars.yml file for the viaq case. (not implemented yet) **
+**6. Sample vars.yml file for the viaq case. (not implemented yet) **
 ```
 logging_enabled: true
 logging_outputs:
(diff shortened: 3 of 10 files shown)


README.md
```diff
@@ -219,7 +219,7 @@ Variables in vars.yml
    -  ** `type: files`**
       - `facility`: facility; default to `*`
       - `severity`: severity; default to `*`
-      - `exclude`: exclude; default to none.
+      - `exclude`: exclude; default to none. exclude list must start with ';' and is delimited by ';'. E.g., ';cron.none;mail.none'
```
pcahyna, line 222: Why to introduce this format? We are using YAML, so we don't have to use strings with embedded delimiters and parse them into lists. We can use lists directly:
```yaml
exclude:
  - cron.none
  - mail.none
```

design_docs/rsyslog_inputs_outputs_flows.md
```diff
@@ -40,7 +40,7 @@ logging_outputs
   - name: files_output
     type: files
     severity: info
-    exclude: authpriv.none;auth.none;cron.none;mail.none
+    exclude: ;authpriv.none;auth.none;cron.none;mail.none
```
pcahyna, line 43: see above

roles/rsyslog/README.md
```diff
@@ -233,7 +233,7 @@ Files output format
      type: files
      facility: <facility_in_text, e.g., "mail"; default to "*">
      severity: <severity_in_text, e.g., "info"; default to "*">
-     exclude: <excluded facility list separated by ';', e.g., "mail.none;auth.none"; default to nil>
+     exclude: <excluded facility list started with and delimited by ';', e.g., ";mail.none;auth.none"; default to nil>
```
pcahyna, line 236: see above

roles/rsyslog/README.md
```diff
@@ -243,7 +243,7 @@ Forwards output format
      type: forwards
      facility: <facility_in_text, e.g., "mail"; default to "*">
      severity: <severity_in_text, e.g., "info"; default to "*">
-     exclude: <excluded facility list separated by ';', e.g., "mail.none;auth.none"; default to nil>
+     exclude: <excluded facility list separated by ';', e.g., ";mail.none;auth.none"; default to nil>
```
pcahyna, line 246: see above

roles/rsyslog/templates/output_files.j2
```diff
@@ -1,13 +1,7 @@
 {% if files_item.path is defined %}
-{%   if files_item.exclude is defined %}
 ruleset(name="{{ files_item.name }}") {
-    {{ files_item.facility | d('*') }}.{{ files_item.severity | d('*') }};{{ files_item.exclude }} {{ files_item.path }}
+    {{ files_item.facility | d('*') }}.{{ files_item.severity | d('*') }}{{ files_item.exclude | d() }} {{ files_item.path }}
```
pcahyna, line 3: if `exclude` becomes a true list, use the join filter: https://jinja.palletsprojects.com/en/2.9.x/templates/#join

README.md
nhosoi: Ok, let me update the patch according to your proposal. Thanks!

pcahyna:
I disagree with the intent of the PR, it is a bad idea to introduce a less intuitive interface (which user would think to prefix a list with a semicolon? `exclude` is a user-settable parameter, right?) in order to simplify the implementation. I.e. the previous version was better. Moreover, the whole thing should be expressed even more naturally as a YAML list. Why are strings with embedded delimiters even used here when we have true lists in YAML?

nhosoi:
> I disagree with the intent of the PR, it is a bad idea to introduce a less intuitive interface (which user would think to prefix a list with a semicolon? `exclude` is a user-settable parameter, right?) in order to simplify the implementation. I.e. the previous version was better. Moreover, the whole thing should be expressed even more naturally as a YAML list. Why are strings with embedded delimiters even used here when we have true lists in YAML?

I think your proposal solves the problem. The semicolon is a delimiter. By setting the exclude list as you suggested, users do not have to worry about what the delimiter is. So, I'd like to go with your suggestions. That is, the following setting will be converted to `;cron.none;main.one`. Is it ok?
```
exclude:
  - cron.none
  - mail.none
```

richm:
> I disagree with the intent of the PR, it is a bad idea to introduce a less intuitive interface (which user would think to prefix a list with a semicolon? `exclude` is a user-settable parameter, right?) in order to simplify the implementation. I.e. the previous version was better. Moreover, the whole thing should be expressed even more naturally as a YAML list. Why are strings with embedded delimiters even used here when we have true lists in YAML?

The intent is to allow to input the values in rsyslog format.  Note that the names such as "severity", "facility", "mail.none", "auth.none", etc. are very rsyslog-centric - that is, someone who knows rsyslog configuration would easily be able to grok them - someone using the logging role to configure e.g. fluentd or beats would not.

pcahyna:
> > I disagree with the intent of the PR, it is a bad idea to introduce a less intuitive interface (which user would think to prefix a list with a semicolon? `exclude` is a user-settable parameter, right?) in order to simplify the implementation. I.e. the previous version was better. Moreover, the whole thing should be expressed even more naturally as a YAML list. Why are strings with embedded delimiters even used here when we have true lists in YAML?
> 
> The intent is to allow to input the values in rsyslog format. 

But is it usual in the rsyslog format to start lists with a semicolon? If not (I don't remember ever seeing it) then the previous format was better, since it was closer to what one sees in rsyslog configuration files. (It is still up to discussion whether to keep the previous format closer to rsyslog.conf or whether use the more regular YAML list syntax.)

> Note that the names such as "severity", "facility", "mail.none", "auth.none", etc. are very rsyslog-centric - that is, someone who knows rsyslog configuration would easily be able to grok them - someone using the logging role to configure e.g. fluentd or beats would not.

Are they? They seem to come from IEEE Std 1003.1 and the representation like `mail.none` from the original BSD syslog, see syslog(3). I.e. they are part of the protocol, not of the implementation. I would suppose that other logging daemons must know about them, because those facilities and levels are generated at the source, by the application that produces the log messages using syslog(3). (I've never used fluentd though.)

richm:
> by the application that produces the log messages using syslog(3). (I've never used fluentd though.)

Right.  The terminology is still used (somewhat) by the syslog (rfc3164, rfc5424) plugins used by fluentd, beats, etc. other logging systems.  But if you aren't using a syslog plugin, e.g. if you are using a systemd input, you'll have to know how to map `auth.none` into systemd `SYSLOG_IDENTIFIER` or `PRIORITY` or some other systemd field.

As far as beginning a string value with a semicolon - the older rsyslog format (still supported by EL7), uses this format a lot, so it simplifies the code handling the values.

I guess what I'm saying is - if we want to get completely away from rsyslog syntax and schema, and generalize everything, there is a *lot* of information coded into the terse, inscrutable rsyslog config format, that would have to be expanded dramatically to provide a truly implementation independent interface easily consumable by a fluentd, beats, etc. provider.

OTOH, perhaps using a format like
```yaml
exclude:
  - cron.none
  - mail.none
```
strikes a good balance that is not too terse but still expressive enough to know what's going on if not familiar with rsyslog.

nhosoi:
`<<snip>>`
> OTOH, perhaps using a format like
> 
> ```yaml
> exclude:
>   - cron.none
>   - mail.none
> ```
> strikes a good balance that is not too terse but still expressive enough to know what's going on if not familiar with rsyslog.

+1
Plus, if users who are not familiar with rsyslog and are willing to configure files output, they can put these 2 lines without path parameter, then the default config file would be placed.
```
        logging_outputs:
          - name: files_output
            type: files
```

design_docs/rsyslog_inputs_outputs_flows.md
richm: I think if you fix this to use the `list` format, this PR is good to go

nhosoi: Thank you, @richm. I've fixed the line 43, squashed commits and pushed.

nhosoi:
Thank you, @richm. Merging...

PR #115, opened 2020-05-08
Adding an rsyslog service error check to the CI tests.

nhosoi:

PR #116, opened 2020-05-11
RHELPLAN-32453 - Add remote inputs (imtcp, imptcp, imudp) and their r…

nhosoi:
…emote_files outputs supports

- Separating the inputs from the remote hosts from the basic input type and
  making it a remote input type.
- Separating the output into the local file from the files output type and
  making it a remote_file output type, which uses an output_remote_files.j2
  template.

- CI tests
  . adding tests_remote_default_remote.yml for testing the default output rules.
  . adding tests_remote_remote.yml for testing the configured output rules.
  . removing an obsolete tests_listen.yml.

- Updating docs:
  . README.md
  . roles/rsyslog/README.md
  . design_docs/logging_subsystem.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Definitions
   - [`Viaq`](https://docs.okd.io/latest/install_config/aggregate_logging.html)- Common Logging based on OpenShift Aggregated Logging (OCP/Origin).
   - [`Elasticsearch`](https://www.elastic.co/) - Non-OpenShift standalone Elasticsearch.
   - `Local` - Output the collected logs to a local File / Journal (Not yet implemented). Supported only for default and basics Rsyslog data at this point.
-  - `Remote Rsyslog` - Output logs to a remote Rsyslog server.
+  - `Remote Rsyslog` - Input logs from or output logs to a remote Rsyslog server.
   - `Message Queue` (kafka, amqp) - Not yet implemented
 
 Deploy Configuration Files
@@ -70,7 +70,7 @@ vars.yml stores variables which are passed to ansible to control the tasks.
 Initial configuration will be supplied by default.
 User can supply further configuration to be used.
 
-Currently, the logging role supports four types of logs collections ([inputs](https://github.com/linux-system-roles/logging/tree/master/roles/rsyslog/tasks/inputs/)): `basics`, `files`, `ovirt`, and `viaq`.  And 3 types of log outputs ([outputs](https://github.com/linux-system-roles/logging/tree/master/roles/rsyslog/tasks/outputs/)): `elasticsearch`, `files`, and `forwards`.  To deploy configuration files with these inputs and outputs, first specify the outputs as `logging_outputs`, then inputs as `logging_inputs`.  For define the flow from inputs to outputs, use `logging_flows`.  The `logging_flows` is made from, `name`, `inputs`, and `outputs`, where `inputs` is a list of `logging_inputs name` values and `outputs` is a list of `logging_outputs name` values.
+Currently, the logging role supports 5 types of logging inputs ([inputs](https://github.com/linux-system-roles/logging/tree/master/roles/rsyslog/tasks/inputs/)): `basics`, `files`, `ovirt`, `remote`, and `viaq`.  And 4 types of outputs ([outputs](https://github.com/linux-system-roles/logging/tree/master/roles/rsyslog/tasks/outputs/)): `elasticsearch`, `files`, `forwards`, and `remote_files`.  To deploy configuration files with these inputs and outputs, specify the inputs as `logging_inputs` and the outputs as `logging_outputs`. For definining the flow from inputs to outputs, use `logging_flows`.  The `logging_flows` has 3 sub variables `name`, `inputs`, and `outputs`, where `inputs` is a list of `logging_inputs name` values and `outputs` is a list of `logging_outputs name` values.
 
 To make an effect with the following setting, vars.yml has to have `logging_enabled: true`.  Unless logging_enabled is set to true, LSR/Logging does not deploy logging systems.
 
@@ -135,21 +135,46 @@ logging_inputs:
     type: basics
 ```
 
-3) Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the local system and remote rsyslog and outputs into the local files.
+3) Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the remote rsyslog and outputs into the local files per host.
 ```
-logging_enabled: true
-rsyslog_capabilities: [ 'network', 'remote-files' ]
-logging_purge_confs: true
 logging_outputs:
-  - name: local-files
-    type: files
+  - name: remote_files_output
+    type: remote_files
 logging_inputs:
-  - name: system-and-remote-input
-    type: basics
+  - name: remote_udp_input
+    type: remote
+    udp_port: 11514
+  - name: remote_tcp_input
+    type: remote
+    tcp_port: 22514
 logging_flows:
-  - name: flow0
-    inputs: [system-and-remote-input]
-    outputs: [local-files]
+  - name: flow_0
+    inputs: [remote_udp_input, remote_tcp_input]
+    outputs: [remote_files_output]
+```
+If more detailed outputs to be configured instead of using the default paths, they are configurable as follows.
+```
+logging_outputs:
+  - name: remote_files_output0
+    type: remote_files
+    remote_log_path: /var/log/remote/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log
+    severity: info
+    exclude: [authpriv.none]
+  - name: remote_files_output1
+    type: remote_files
+    remote_sub_path: others/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log
+    facility: authpriv
+logging_inputs:
+  - name: remote_udp_input
+    type: remote
+    udp_port: 11514
+  - name: remote_tcp_input
+    type: remote
+    tcp_port: 22514
+logging_flows:
+  - name: flow_0
+    inputs: [remote_udp_input, remote_tcp_input]
+    outputs: [remote_files_output0, remote_files_output1]
 ```
 
 4) Deploying config files for collecting logs from OpenShift pods as well as RHV and forwarding them to elasticsearch.
@@ -206,7 +231,7 @@ Variables in vars.yml
 - `logging_mmk8s_ca_cert`: Path to CA cert for kubernetes.  Default to "/etc/rsyslog.d/mmk8s.ca.crt".
 - `logging_outputs`: A set of following variables to specify output configurations.  It could be an list if multiple outputs that should to be configured.
    - `name`: Unique name of the output
-   - `type`: Type of the output element. Currently, `elasticsearch`, `files`, and `forwards` are supported. The `type` is used to specify a task type which corresponds to a directory name in roles/rsyslog/tasks/outputs/.
+   - `type`: Type of the output element. Currently, `elasticsearch`, `files`, `forwards`, and `remote_files` are supported. The `type` is used to specify a task type which corresponds to a directory name in roles/rsyslog/{tasks,vars}/outputs/.
    -  ** `type: elasticsearch`**
       - `server_host`: Hostname elasticsearch is running on.
       - `server_port`: Port number elasticsearch is listening to.
@@ -235,10 +260,27 @@ Variables in vars.yml
       - `name`: Name of the custom file output element.
       - `type`: Type of the output element.
       - `custom_config_files`: List of custom configuration files are deployed to /etc/rsyslog.d. [ '/path/to/custom_A.conf', '/path/to/custom_B.conf' ]. Default to none.
+   -  ** `type: remote_files`**
+      - `facility`: facility; default to `*`
+      - `severity`: severity; default to `*`
+      - `exclude`: exclude list; default to none.
+      - `remote_log_path`: full path to store the filtered logs.
+      - `remote_sub_path`: relative path to logging_system_log_dir to store the filtered logs.
+	                       if `remote_log_path` nor `remote_sub_path` are not specified, the remote_file output configured with the default settings.
 - `logging_inputs` : List of optional logs collections, dictionaries with `name`, `type` and `state` attributes, that were pre-configured.
         - `name`: Unique name of the input.
-          `type`: The type of the pre-configured logs to collect. **Note:** Currently ['viaq', 'viaq-k8s', 'ovirt'] are supported for the elasticsearch output.
+          `type`: The type of the log inputs. **Note:** Currently [`basics`, `files`, `ovirt`, `remote`, and `viaq`] are supported.
           `state`: The state of the configuration files states if they should be `present` or `absent`. Default to `present`.
+   -  ** `type: basics`**
+      - `rsyslog_imjournal_ratelimit_burst`: set to imjournal RateLimit.Burst. Default to 20000.
+      - `rsyslog_imjournal_ratelimit_interval`: set to imjournal RateLimit.Interval. Default to 600.
+      - `rsyslog_imjournal_persist_state_interval`: set to imjournal PersistStateInterval. Default to 10.
+   -  ** `type: files`**
+      - `rsyslog_input_log_path`: File name to be read by the imfile plugin. The value should be full path. Wildcard '*' is allowed in the path.  Default to `/var/log/containers/*.log`.
+   -  ** `type: remote_files`**
+      - `udp_port`: if the port number is given, rsyslog is configured to listen at the udp port number. Default to 514.
+      - `tcp_port`: if the port number is given, rsyslog is configured to listen at the tcp port number. Default to 514.
+      - `tcp_tls_port`: if the port number is given, rsyslog is configured to listen at the tls tcp port number. Default to 6514.
 
 - `logging_provider`: The logging collector to use for the logging provider. Currently Rsyslog is the only supported logging provider. Defaults to `rsyslog`.
 - `logging_purge_confs`: By default, the Rsyslog configuration files are applied on top of pre-existing configuration files. To purge local files prior to setting new ones, set logging_purge_confs variable to 'true', it will move all Rsyslog configuration files to a backup directory, `/tmp/rsyslog.d-XXXXXX/backup.tgz`, before deploying the new configuration files. Defaults to 'false'.
@@ -283,7 +325,7 @@ Additional output will be added.
 
 Planned Flows:
 --------------
-  - `Rsyslog` -> `Local` (RHEL Default) / `Viaq` [1] / `Elasticsearch` / `Remote Rsyslog` / `message Queue (kafka, amqp)`
+  - `Rsyslog` -> `Local` (RHEL Default) / `Viaq` [1] / `message Queue (kafka, amqp)`
 
 [1] Rsyslog to Viaq currently means doing output to the OCP Elasticsearch using client cert auth.
     In the future we want to support Rsyslog to OCP rsyslog using RELP, or Rsyslog to mux using fluent relp input plugin, or message queue.
--- a/design_docs/logging_subsystem.md
+++ b/design_docs/logging_subsystem.md
@@ -8,7 +8,7 @@ It is designed to satisfy following requirements.
 - The existing scenario, especially an input ovirt + output elasticsearch should not be affected by the updates.
 
 ```
-logging/
+logging
 ├── defaults
 │   └── main.yml
 ├── design_docs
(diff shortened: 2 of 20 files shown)


README.md
```diff
@@ -135,21 +135,46 @@ logging_inputs:
...
-    type: files
-logging_inputs:
-  - name: system-and-remote-input
-    type: basics
-logging_flows:
-  - name: flow0
-    inputs: [system-and-remote-input]
-    outputs: [local-files]
+        logging_outputs:
+          - name: default_remote_files
```
pcahyna, line 141: Why is it called `default_remote_files`? Is it a special name or just an arbitrary name that you introduced in the example? If it is just an arbitrary name, why is it called `default_something`? To me `default` sounds like something predefined.

README.md
```diff
@@ -135,21 +135,46 @@ logging_inputs:
...
-logging_outputs:
-  - name: local-files
-    type: files
-logging_inputs:
-  - name: system-and-remote-input
-    type: basics
-logging_flows:
-  - name: flow0
-    inputs: [system-and-remote-input]
-    outputs: [local-files]
```
pcahyna, line 152: is there a reason to increase the indent of those lines?

tests/tests_remote_default_remote.yml
```diff
@@ -27,6 +30,22 @@
...
+      assert:
+        that: rsyslog_conf_line_count.stdout.0 | int <= 7
+
+    - name: Check file counts in rsyslog.d
+      assert:
+        that: rsyslog_d_file_count.matched >= 8
+
+    - name: Check rsyslog errors
+      shell: |
+        systemctl status rsyslog | grep error
```
pcahyna, line 43: please no `shell` and `grep`.

tests/tests_remote_default_remote.yml
```diff
@@ -27,6 +30,22 @@
...
+      register: result
+      failed_when: result.rc == 0
+      tags:
+        - skip_ansible_lint
```
pcahyna, line 47: what does this tag do?

README.md
nhosoi: > Why is it called `default_remote_files`? Is it a special name or just an arbitrary name that you introduced in the example? If it is just an arbitrary name, why is it called `default_something`? To me `default` sounds like something predefined.

The name can be any string, but since there is no `remote_log_path` nor `remote_sub_path` is defined in the remote_files output, the [default configuration](https://github.com/linux-system-roles/logging/pull/116/files#diff-b23a18f65552498ad31bf927da164818R20-R55) is used.

README.md
nhosoi: No, I don't think so...

README.md
nhosoi: Replaced `default_remote_files` with `remote_files_output`.

roles/rsyslog/tasks/inputs/remote/main.yml
```diff
@@ -9,15 +9,26 @@
...
   vars:
-    __rsyslog_input_decl: ""
-    __udp_input_calls: "# Log messages from remote hosts over UDP\ninput(name=\"{{ remote_item.name }}\" type=\"imudp\" port=\"{{ remote_item.udp_port | d(__rsyslog_default_port) }}\")\n"
-    __ptcp_input_calls: "# Log messages from remote hosts over plain TCP\ninput(name=\"{{ remote_item.name }}\" type=\"imptcp\" port=\"{{ remote_item.tcp_port | d(__rsyslog_default_port) }}\")\n"
-    __tls_input_calls: "# Log messages from remote hosts over TLS\ninput(name=\"{{ remote_item.name }}\" type=\"imtcp\" port=\"{{ remote_item.tcp_tls_port | d(__rsyslog_default_tls_port) }}\")\n"
     __rsyslog_input_name: "{{ remote_item.name }}"
     __rsyslog_input_type: "{{ remote_item.type }}"
     __rsyslog_packages: []
-    __rsyslog_rules: [ { "name": "input-remote-{{ remote_item.name }}", "type": "input", "weight": "11",
-                         "sections": [ { "options": "{{ __udp_input_calls if (remote_item.udp_port | d() ) else ( __ptcp_input_calls if remote_item.tcp_port | d() else __tls_input_calls ) }}{{ lookup('template', 'input_template.j2') }}" } ] } ]
```
pcahyna: @nhosoi You say "__rsyslog_rules value from a string to the dictionary" but __rsyslog_rules was a dictionary before and it is a dictionary after the change.

nhosoi: @pcahyna, indeed :p. I'll fix the comment to `modifying the __rsyslog_rules value to YAML friendly manner.`

README.md
```diff
@@ -135,21 +135,46 @@ logging_inputs:
     type: basics
 ```
 
-3) Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the local system and remote rsyslog and outputs into the local files.
+3) Deploying basic LSR/Logging config files in /etc/rsyslog.d, which handle inputs from the remote rsyslog and outputs into the per host base local files.
```
richm, line 138: did you mean "per host"?  not sure what "base" means in this context

README.md
```diff
@@ -135,21 +135,46 @@ logging_inputs:
...
+    outputs: [remote_files_output]
+```
+If more detailed outputs to be configured instead of using the default paths, they are configurable as follows.
+```
+logging_outputs:
+  - name: remote_files_output0
+    type: remote_files
+    remote_log_path: /var/log/remote/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log
+    severity: info
+    exclude: ;authpriv.none
```
richm, line 162: this will change if https://github.com/linux-system-roles/logging/pull/114 merges first

README.md
```diff
@@ -235,10 +260,27 @@ Variables in vars.yml
       - `name`: Name of the custom file output element.
       - `type`: Type of the output element.
       - `custom_config_files`: List of custom configuration files are deployed to /etc/rsyslog.d. [ '/path/to/custom_A.conf', '/path/to/custom_B.conf' ]. Default to none.
+   -  ** `type: remote_files`**
+      - `facility`: facility; default to `*`
+      - `severity`: severity; default to `*`
+      - `exclude`: exclude; default to none.
```
richm, line 266: this will change too

README.md
```diff
@@ -235,10 +260,27 @@ Variables in vars.yml
...
+   -  ** `type: basics`**
+      - `rsyslog_imjournal_ratelimit_burst`: set to imjournal RateLimit.Burst. Default to 20000.
+      - `rsyslog_imjournal_ratelimit_interval`: set to imjournal RateLimit.Interval. Default to 600.
+      - `rsyslog_imjournal_persist_state_interval`: set to imjournal PersistStateInterval. Default to 10.
+   -  ** `type: files`**
+      - `rsyslog_input_log_path`: File name to be read by the imfile plugin. The value should be full path. Wildcard '*' is allowed in the path.  Default to `/var/log/containers/*.log`.
+   -  ** `type: remote_files`**
+      - `udp_port`: if the port number is given, rsyslog is configured to listen at the udp port number. Default to 514.
+      - `tcp_port`: if the port number is given, rsyslog is configured to listen at the tcp port number. Default to 514.
+      - `tcp_tls_port`: if the port number is given, rsyslog is configured to listen at the tls tcp port number. Default to 6514.
```
richm, line 283: this needs certs, key?

tests/tests_remote_default_remote.yml
```diff
@@ -27,6 +30,19 @@
...
+        that: rsyslog_conf_line_count.stdout.0 | int <= 7
+
+    - name: Check file counts in rsyslog.d
+      assert:
+        that: rsyslog_d_file_count.matched >= 8
+
+    - name: Check rsyslog errors
+      command: systemctl status rsyslog
+      register: systemctl_result
+      failed_when: "'error' in systemctl_result.stderr"
```
richm, line 44: `or systemctl_result.rc != 0`

README.md
nhosoi: Yes, I wanted to convert this line into a plain English... :p
`/var/log/remote/%HOSTNAME%/%PROGRAMNAME:::secpath-replace%.log`
With this path, remote logs are stored in the local files.
```
ls /var/log/remote/msg/_hostname_
abrt-hook-ccpp.log journal.log         newtag.log         setroubleshoot.log
auditd.log               kernel.log          nm-dispatcher.log  systemd.log
testTag0.log           ...
```
Is "outputs into the local files per host" better?

README.md
nhosoi: Right. I'll fix the conflicts...

README.md
richm: > Is "outputs into the local files per host" better?

Yes, works for me

README.md
nhosoi: Right. The tls part hasn't been touched yet. :(

tests/tests_remote_default_remote.yml
nhosoi: Actually, the return value is 0...
Please see this comment and the commit in pr115.
https://github.com/linux-system-roles/logging/pull/115#issuecomment-627882173
https://github.com/linux-system-roles/logging/pull/115/files#diff-b66e58c3692b490cd622af3414c88ab1R75-R81

richm: Right - I get that in some error cases, `rc` will be `0` and there will be the string `error` somewhere in the output.  But for some error cases, maybe not even ones that you are expecting, it is possible that `systemctl status rsyslog` will return an non-zero exit code, and the string `error` might not be in the output in that case (e.g. rsyslog crashed for some reason before being able to write out the full error message).  In that case, rsyslog will have failed, but you will have masked the error by using `failed_when`.

nhosoi: Thanks, @richm. Fixed. Also, rebased with master.

roles/rsyslog/tasks/inputs/remote/main.yml
```diff
@@ -0,0 +1,37 @@
...
+              # Log messages from remote hosts over UDP
+              input(name="{{ remote_item.name }}" type="imudp" port="{{ remote_item.udp_port | d(__rsyslog_default_port) }}")
+              {% elif remote_item.tcp_port | d() %}
+              # Log messages from remote hosts over plain TCP
+              input(name="{{ remote_item.name }}" type="imptcp" port="{{ remote_item.tcp_port | d(__rsyslog_default_port) }}")
+              {% else %}
+              # Log messages from remote hosts over TLS
+              input(name="{{ remote_item.name }}" type="imtcp" port="{{ remote_item.tcp_tls_port | d(__rsyslog_default_tls_port) }}")
+              {% endif %}
+              {{ lookup('template', 'input_template.j2') }}
```
richm, line 31: I suggest putting this into a template file `input_remote_template.j2` or something like that

roles/rsyslog/tasks/outputs/remote_files/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+- name: Create files output configuration files in /etc/rsyslog.d
+  vars:
+    __rsyslog_packages: "{{ __rsyslog_remote_files_output_packages }}"
+    __rsyslog_rules: [ { "name": "output-files-{{ remote_files_item.name }}", "type": "output",
+                         "sections": [ { "options": "{{ lookup('template', 'output_remote_files.j2') }}" } ] } ]
```
richm, line 6: I suggest cleaning this up to put it into a more yaml/ansible friendly structure

roles/rsyslog/tasks/inputs/remote/main.yml
nhosoi: Thanks, @richm. I'm updating the pr.

I was wondering which is preferable - sharing templates and making the number of templates as small as possible vs simplifying the task code. It sounds latter, doesn't it? It's not like an analogy like a template is a library function to be shared so that it's supposed to be generic. :)

richm: > I was wondering which is preferable - sharing templates and making the number of templates as small as possible vs simplifying the task code. It sounds latter, doesn't it?

In this case, yes, but like anything, it is a judgement call - if the tasks code is "too complex", it should be moved into a template file, or otherwise refactored - for sure the yaml and ansible linters don't like really long lines, and trying to follow rules of indentation can sometimes reduce readability.  On the other hand, moving too much code to template files can also reduce readability if you have to constantly switch back and forth between the tasks and the template to find out what's going on.

roles/rsyslog/templates/input_remote.j2
```diff
@@ -0,0 +1,46 @@
+{% if remote_item.udp_port | d() %}
+# Log messages from remote hosts over UDP
+input(name="{{ remote_item.name }}" type="imudp" port="{{ remote_item.udp_port | d(__rsyslog_default_port) }}")
+{% elif remote_item.tcp_port | d() %}
+# Log messages from remote hosts over plain TCP
+input(name="{{ remote_item.name }}" type="imptcp" port="{{ remote_item.tcp_port | d(__rsyslog_default_port) }}")
+{% else %}
+# Log messages from remote hosts over TLS
+input(name="{{ remote_item.name }}" type="imtcp" port="{{ remote_item.tcp_tls_port | d(__rsyslog_default_tls_port) }}")
+{% endif %}
```
richm, line 10: I think you can use `{{ lookup('template', 'input_template.j2') }}` here to avoid having to copy/paste the contents of input_template.j2 here.  In that sense, yes, you could think of a template as a "library" of "functions" that you can "include" in multiple places.

https://docs.ansible.com/ansible/latest/plugins/lookup/template.html

nhosoi: Ah, I see. We can template in a template file... :O

tests/tests_remote_default_remote.yml
pcahyna: Indeed - `or` was part of the suggested code :-)

roles/rsyslog/tasks/inputs/remote/main.yml
pcahyna: looking up a template from another template gives you the best of both, doesn't it?

pcahyna: Also, would template inheritance help as an alternative to the template lookup? https://jinja.palletsprojects.com/en/2.9.x/templates/#template-inheritance

roles/rsyslog/templates/input_remote.j2
pcahyna: Or inherit from templates, not sure whether it helps here though: https://jinja.palletsprojects.com/en/2.9.x/templates/#template-inheritance

roles/rsyslog/defaults/main.yml
```diff
@@ -154,20 +154,15 @@ rsyslog_send_permitted_peers: '{{ rsyslog_permitted_peers }}'
 # Firewall, UDP, TCP ports
 # ------------------------
 
-# rsyslog_udp_port
+# default rsyslog port
 #
-# The incoming UDP port used for remote logging.
-rsyslog_udp_port: '514'
+# The incoming TCP/UDP port used for remote logging.
+__rsyslog_default_port: '514'
```
richm, line 160: private variables not settable by users should go into `vars/main.yml` - https://github.com/oasis-roles/meta_standards#vars-vs-defaults

roles/rsyslog/vars/inputs/remote/main.yml
```diff
@@ -0,0 +1,75 @@
...
+# List of ``rsyslogd`` modules that enable logs from the local system to be
+# received and parsed by the ``rsyslogd`` daemon.
+__rsyslog_conf_remote_modules:
+
+  - name: 'input-remote-modules'
+    type: 'modules'
+    sections:
+
+      - options: |-
+          {% set ns0 = namespace(use_tls = false) %}
```
richm, line 38: This should be in a template file

nhosoi:
Rebased against the master branch.

roles/rsyslog/defaults/main.yml
nhosoi: This is a fallback port number used like this. Yes, I should move it to roles/rsyslog/vars/main.yml...
`input(name="{{ remote_item.name }}" type="imudp" port="{{ remote_item.udp_port | d(__rsyslog_default_port) }}")`

.travis.yml
```diff
@@ -7,8 +7,8 @@ env:
   matrix:
     - MOLECULE_DISTRO="centos:7"
     - MOLECULE_DISTRO="centos:8"
-    - MOLECULE_DISTRO="fedora:30"
     - MOLECULE_DISTRO="fedora:31"
+    - MOLECULE_DISTRO="fedora:32"
```
richm, line 11: please do this in a separate PR - unless this PR needs to specifically run the molecule tests on fedora 32

nhosoi: Sure. Is it ok to remove fedora:30 in this PR? Otherwise, continuous-integration/travis-ci/pr does not pass...

richm:
lgtm

richm:
>  @nhosoi nhosoi force-pushed the nhosoi:32453 branch from cb3b9ae to 528565d 16 minutes ago 

What changed?

nhosoi:
> > @nhosoi nhosoi force-pushed the nhosoi:32453 branch from [cb3b9ae](https://github.com/linux-system-roles/logging/commit/cb3b9aed48fc49db51e8ba75aa853b406886de38) to [528565d](https://github.com/linux-system-roles/logging/commit/528565da28bb570f1b00c229b26f1994b9774799) 16 minutes ago
> 
> What changed?

Squashed and cleaned up the github comment...

richm:
> > > @nhosoi nhosoi force-pushed the nhosoi:32453 branch from [cb3b9ae](https://github.com/linux-system-roles/logging/commit/cb3b9aed48fc49db51e8ba75aa853b406886de38) to [528565d](https://github.com/linux-system-roles/logging/commit/528565da28bb570f1b00c229b26f1994b9774799) 16 minutes ago
> > 
> > 
> > What changed?
> 
> Squashed and cleaned up the github comment...
> I don't think it introduced the travis-ci failure, but somehow it failed on CentOS-8
> 
> ```
> TASK [/home/travis/build/linux-system-roles/logging/roles/rsyslog : Create a work directory] ***
> fatal: [instance]: FAILED! => {"changed": false, "module_stderr": "", "module_stdout": "OCI runtime exec failed: exec failed: container_linux.go:348: starting container process caused \"process_linux.go:90: adding pid 7145 to cgroups caused \\\"failed to write 7145 to cgroup.procs: write /sys/fs/cgroup/cpu,cpuacct/docker/a459ded0f0c7ec8fce720c5fdb33b490b89993d24f226aa11afa09fb7582682e/cgroup.procs: invalid argument\\\"\": unknown\r\n", "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error", "rc": 126}
> ```

Can you reproduce locally?  It looks like it is failing on this task:
```yaml
    # Creating a work directory
    - name: Create a work directory
      file:
        state: directory
        path: '{{ rsyslog_work_dir }}'
```
where the default for `rsyslog_work_dir` is `/var/lib/rsyslog` - is it not allowed to create `/var/lib/rsyslog` in a centos8 container?

nhosoi:
> Can you reproduce locally? It looks like it is failing on this task:
> 
> ```yaml
>     # Creating a work directory
>     - name: Create a work directory
>       file:
>         state: directory
>         path: '{{ rsyslog_work_dir }}'
> ```
> 
> where the default for `rsyslog_work_dir` is `/var/lib/rsyslog` - is it not allowed to create `/var/lib/rsyslog` in a centos8 container?

Well, the path is taken in every CI-test and there's no failure observed. Not sure why, but I feel the travis-ci randomly fails in these couple of days...

richm:
> Well, the path is taken in every CI-test and there's no failure observed. Not sure why, but I feel the travis-ci randomly fails in these couple of days...

Ok.  Not sure what's going on.  Also, I'm not aware of a way to tell travis-ci to re-run a failed test.  If you can't figure out a way to do this, then submit a "dummy" commit i.e. make a very minor change to the README or something like that, to trigger a new travis-ci run.

nhosoi:
> > Well, the path is taken in every CI-test and there's no failure observed. Not sure why, but I feel the travis-ci randomly fails in these couple of days...
> 
> Ok. Not sure what's going on. Also, I'm not aware of a way to tell travis-ci to re-run a failed test. If you can't figure out a way to do this, then submit a "dummy" commit i.e. make a very minor change to the README or something like that, to trigger a new travis-ci run.

I changed one character in the github comment and forced to push :). Glad the travis-ci passed this time.

nhosoi:
Thank you for reviewing this PR, @richm. Merging.

design_docs/logging_subsystem.md
```diff
@@ -8,7 +8,7 @@ It is designed to satisfy following requirements.
 - The existing scenario, especially an input ovirt + output elasticsearch should not be affected by the updates.
 
 ```
-logging/
+logging
```
pcahyna, line 11: Is it useful to have this long directory tree listing in this file that you must update every time you change the role structure? Maybe consider removing it in some future PR? @jvymazal @richm do you find useful to have this in the document?

jvymazal: For me, probably not that much, if I need it, I can always browse it either here or locally.

richm: maybe a description of the general layout, but not the specific files e.g. logging -> roles -> rsyslog -> tasks -> output -> task file for each output

PR #117, opened 2020-05-11
Fixing the key/cert setting parameter use_cert for the elasticsearch …

nhosoi:
…output

The key/certs for the elasticsearch output is controled by an elasticsearch
boolean sub variable use_cert. If on, ca_cert, cert and key sub variables
are supposed to be configured. Alternatively, if a global variable
use_local_omelasticsearch_cert is set to true, key/certs files on the local
host are copied to rsyslog_config_dir (/etc/rsyslog.d, by default) on the
target hosts and set them.

Adding the test case tests/tests_files_elasticsearch_use_local_cert.yml

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,9 +238,13 @@ Variables in vars.yml
       - `index_prefix`: Elasticsearch index prefix the particular log will be indexed to.
       - `input_type`: Specifying the input type. Type `ovirt` and `viaq` are supported. Default to `ovirt`.
       - `retryfailures`: Specifying whether retries or not in case of failure. on or off.  Default to on.
-      - `ca_cert`: Path to CA cert for ElasticSearch.  Default to '/etc/rsyslog.d/es-ca.crt'
-      - `cert`: Path to cert for ElasticSearch.  Default to '/etc/rsyslog.d/es-cert.pem'
-      - `key`: Path to key for ElasticSearch.  Default to "/etc/rsyslog.d/es-key.pem"
+      - `use_cert`: If true, key/certificates are used to access Elasticsearch. Triplets {`ca_cert`, `cert`, key`} and/or {`ca_cert_src`, `cert_src`, `key_src`} should be configured. Default to true.
+      - `ca_cert`: Path to CA cert for Elasticsearch.  Default to '/etc/rsyslog.d/es-ca.crt'
+      - `cert`: Path to cert for Elasticsearch.  Default to '/etc/rsyslog.d/es-cert.pem'
+      - `key`: Path to key for Elasticsearch.  Default to "/etc/rsyslog.d/es-key.pem"
+      - `ca_cert_src`: Path to the CA cert file on the local host to copy to the target host. If `ca_cert` is specified, copied to the location. Otherwise, to rsyslog_config_dir.
+      - `cert_src`: Path to the cert file on the local host to copy to the target host. If `cert` is specified, copied to the location. Otherwise, to rsyslog_config_dir.
+      - `key_src`: Path to the key file on the local host to copy to the target host. If `key` is specified, copied to the location. Otherwise, to rsyslog_config_dir.
    -  ** `type: files`**
       - `facility`: facility; default to `*`
       - `severity`: severity; default to `*`
--- a/roles/rsyslog/README.md
+++ b/roles/rsyslog/README.md
@@ -260,15 +260,23 @@ Elasticsearch, Files, Remote_files, and Forwards outputs sub-variables
 - `elasticsearch`: array of dictionary to specify the parameters to forward the log messages to elasticsearch.
    ```
    - name: <unique_name>
-     type: elasticsearch
-     server_host: <elasticsearch hostname>
-     server_port: <elasticsearch port, default to 9200>
-     index_prefix: <prefix to use in the elasticsearch index file>
-     input_type: <input type, e.g., ovirt>
-     retryfailures: on|off, default to on
-     ca_cert: <path to ca.crt>
-     cert: <path to es-cert.pem>
-     key: <path to es-key.pem>
+     type:          elasticsearch
+     server_host:   Hostname elasticsearch is running on.
+     server_port:   Port number elasticsearch is listening to. Default to 9200.
+     index_prefix:  Elasticsearch index prefix the particular log is to be indexed.
+     input_type:    Specifying the input type. Type ovirt and viaq are supported. Default to ovirt.
+     retryfailures: Specifying whether retries or not in case of failure. on or off.  Default to on.
+     use_cert:      If true, access to Elasticsearch using ca_cert, cert and key. Default to true.
+	                Triplets {ca_cert, cert, key} and/or {ca_cert_src, cert_src, key_src} are required.
+     ca_cert:       Path to CA cert for Elasticsearch.  Default to '/etc/rsyslog.d/es-ca.crt'
+     cert:          Path to cert for Elasticsearch.  Default to '/etc/rsyslog.d/es-cert.pem'
+     key:           Path to key for Elasticsearch.  Default to "/etc/rsyslog.d/es-key.pem"
+     ca_cert_src:   Path to the CA cert file on the local host to copy to the target host.
+	                If ca_cert is specified, copied to the location. Otherwise, to rsyslog_config_dir.
+     cert_src:      Path to the cert file on the local host to copy to the target host.
+	                If cert is specified, copied to the location. Otherwise, to rsyslog_config_dir.
+     key_src:       Path to the key file on the local host to copy to the target host.
+	                If key is specified, copied to the location. Otherwise, to rsyslog_config_dir.
    ```
 - `files`: array of dictionary to specify the facility and severity filter and the full path to store logs satisfying the filter.  It takes the sub-variables - `name`, `facility`, `severity`, `exclude`, and `path`.  Unless the name and the path are given, the element is skipped.
 Files output format
@@ -324,23 +332,6 @@ Viaq inputs sub-variables
 - `rsyslog_viaq_log_dir`: Viaq log directory.  Default to '/var/log/containers'.
 - `logging_mmk8s_token`: Path to token for kubernetes.  Default to "/etc/rsyslog.d/mmk8s.token"
 - `logging_mmk8s_ca_cert`: Path to CA cert for kubernetes.  Default to "/etc/rsyslog.d/mmk8s.ca.crt"
-- `use_omelasticsearch_cert` : If set to 'true', omelasticsearch is configured to use the certificates specified in the elasticsearch type logging_outputs.  Default to 'false'.
-- `use_local_omelasticsearch_cert` : If set to 'true', local files ca_cert_src, cert_src and key_src in the elasticsearch type logging_outputs will be deployed to the remote host.
-
-- For the elasticsearch type logging_outputs, a set of following variables are to specify output elasticsearch configurations. It could be an array if multiple elasticsearch clusters to be configured.
-   If set to 'true', ca_cert_src, cert_src and key_src must be set in each elasticsearch element. Otherwise, the deployment fails. Default to 'false'.
-  - `name`: Name of the elasticsearch element.
-  - `server_host`: Hostname elasticsearch is running on.
-  - `server_port`: Port number elasticsearch is listening to.
-  - `index_prefix`: Elasticsearch index prefix the particular log is to be indexed.
-  - `input_type`: Specifying the input type. Type `ovirt` and `viaq` are supported. Default to `ovirt`.
-  - `retryfailures`: Specifying whether retries or not in case of failure. on or off.  Default to on.
-  - `ca_cert`: Path to CA cert for Elasticsearch.  Default to '/etc/rsyslog.d/es-ca.crt'
-  - `cert`: Path to cert for Elasticsearch.  Default to '/etc/rsyslog.d/es-cert.pem'
-  - `key`: Path to key for Elasticsearch.  Default to "/etc/rsyslog.d/es-key.pem"
-  - `ca_cert_src`: Path to the local CA cert file to deploy for Elasticsearch.
-  - `cert_src`: Path to the local cert file to deploy for Elasticsearch.
-  - `key_src`: Path to the local key file to deploy for Elasticsearch.
 
 Contents of Roles
 =================
--- a/roles/rsyslog/tasks/outputs/elasticsearch/main.yml
+++ b/roles/rsyslog/tasks/outputs/elasticsearch/main.yml
@@ -25,6 +25,16 @@
         - '{{ rsyslog_output_elasticsearch }}'
       when: item.key_src | d()
 
+    - name: Check certs - use_cert is true, but triplets are not given
+      fail:
+        msg: "Error: you specified use_cert: true; you must specify all 3 of ca_cert, cert, key, or all 3 of ca_cert_src, cert_src, key_src, or set use_cert: false in the elasticsearch output named {{ item.name }}"
+      with_items:
+        - '{{ rsyslog_output_elasticsearch }}'
+      when:
+        - item.use_cert | d(true)
+        - not ((item.ca_cert | d() and item.cert | d() and item.key | d()) or
+               (item.ca_cert_src | d() and item.cert_src | d() and item.key_src | d()))
+
     - name: Update ca_cert path to /etc/rsyslog.d directory in Rsyslog
       set_fact:
         __rsyslog_elasticsearch_temp: >-
@@ -37,6 +47,16 @@
       with_items:
         - '{{ rsyslog_output_elasticsearch }}'
       when: item.ca_cert_src | d() or item.cert_src | d() or item.key_src | d()
+
+    - name: Check certs - key/certs data are provided, but use_cert is false
+      fail:
+        msg: "Error: you specified use_cert: false and also specified one or more cert/key files in the elasticsearch output named {{ item.name }}"
+      with_items:
+        - '{{ rsyslog_output_elasticsearch }}'
+      when:
+        - not (item.use_cert | d(true))
+        - item.ca_cert | d() or item.cert | d() or item.key | d()
+        - item.ca_cert_src | d() or item.cert_src | d() or item.key_src | d()
 
     - name: Set updated rsyslog_output_elasticsearch
       set_fact:
--- a/roles/rsyslog/tasks/outputs/elasticsearch/main.yml
+++ b/roles/rsyslog/tasks/outputs/elasticsearch/main.yml
@@ -4,53 +4,47 @@
     - name: Copy local Elasticsearch ca_certs to /etc/rsyslog.d directory in Rsyslog
       copy:
         src: '{{ item.ca_cert_src }}'
-        dest: '{{ rsyslog_config_dir }}'
+        dest: '{{ item.ca_cert | d(rsyslog_config_dir) }}'
       with_items:
         - '{{ rsyslog_output_elasticsearch }}'
-      when:
-        - item.ca_cert_src is defined
-        - item.ca_cert_src | length > 0
+      when: item.ca_cert_src | d()
 
     - name: Copy local Elasticsearch certs to /etc/rsyslog.d directory in Rsyslog
       copy:
         src: '{{ item.cert_src }}'
-        dest: '{{ rsyslog_config_dir }}'
+        dest: '{{ item.cert | d(rsyslog_config_dir) }}'
       with_items:
         - '{{ rsyslog_output_elasticsearch }}'
-      when:
-        - item.cert_src is defined
-        - item.cert_src | length > 0
+      when: item.cert_src | d()
 
     - name: Copy local Elasticsearch keys to /etc/rsyslog.d directory in Rsyslog
       copy:
         src: '{{ item.key_src }}'
-        dest: '{{ rsyslog_config_dir }}'
+        dest: '{{ item.key | d(rsyslog_config_dir) }}'
       with_items:
         - '{{ rsyslog_output_elasticsearch }}'
-      when:
-        - item.key_src is defined
-        - item.key_src | length > 0
+      when: item.key_src | d()
 
     - name: Update ca_cert path to /etc/rsyslog.d directory in Rsyslog
       set_fact:
         __rsyslog_elasticsearch_temp: >-
           {{ __rsyslog_elasticsearch_temp | d([]) }} +
-          {{ [ item | combine( { 'ca_cert': item.ca_cert_src | basename ,
-          'cert': item.cert_src | basename, 'key': item.key_src | basename } ) ] }}
+          {{ [ item |
+               combine( { 'ca_cert': item.ca_cert | d(rsyslog_config_dir + '/' + item.ca_cert_src | basename),
+                          'cert': item.cert | d(rsyslog_config_dir + '/' + item.cert_src | basename),
+                          'key': item.key | d(rsyslog_config_dir + '/' + item.key_src | basename) } )
+          ] }}
       with_items:
         - '{{ rsyslog_output_elasticsearch }}'
-      when:
-        - item.ca_cert_src is defined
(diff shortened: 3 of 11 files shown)


pcahyna:
I don't really understand the details, so my only comment is, please think about integration with the certificate role, if applicable. https://github.com/seocam/certificate/blob/master/README.md

README.md
```diff
@@ -213,9 +213,13 @@ Variables in vars.yml
       - `index_prefix`: Elasticsearch index prefix the particular log will be indexed to.
       - `input_type`: Specifying the input type. Type `ovirt` and `viaq` are supported. Default to `ovirt`.
       - `retryfailures`: Specifying whether retries or not in case of failure. on or off.  Default to on.
+      - `use_cert`: If true, the either of the following 2 triplets should be configured. Default to true.
       - `ca_cert`: Path to CA cert for ElasticSearch.  Default to '/etc/rsyslog.d/es-ca.crt'
       - `cert`: Path to cert for ElasticSearch.  Default to '/etc/rsyslog.d/es-cert.pem'
       - `key`: Path to key for ElasticSearch.  Default to "/etc/rsyslog.d/es-key.pem"
+      - `ca_cert_src`: Path to the local CA cert file to deploy for ElasticSearch. Effective when use_local_omelasticsearch_cert is true.
```
richm, line 220: Elasticsearch - no uppercase `S`

richm:
> I don't really understand the details, so my only comment is, please think about integration with the certificate role, if applicable. https://github.com/seocam/certificate/blob/master/README.md

The use case is that the user has already generated the necessary certs/keys.

Yes, this and other roles should be integrated with the certificate role, when it is available.

nhosoi:
> I don't really understand the details, so my only comment is, please think about integration with the certificate role, if applicable. https://github.com/seocam/certificate/blob/master/README.md

Could it be ok to make a plan for the next quarter? I don't think I can make it in this one... What I did in this PR is to introduce use_cert to allow elasticsearch configuration not to set up key/certs. Previously, they were always required.

richm:
> > I don't really understand the details, so my only comment is, please think about integration with the certificate role, if applicable. https://github.com/seocam/certificate/blob/master/README.md
> 
> Could it be ok to make a plan for the next quarter?

Yes.

> I don't think I can make it in this one... What I did in this PR is to introduce use_cert to allow elasticsearch configuration not to set up key/certs. Previously, they were always required.

nhosoi:
> > > I don't really understand the details, so my only comment is, please think about integration with the certificate role, if applicable. https://github.com/seocam/certificate/blob/master/README.md

> > Could it be ok to make a plan for the next quarter?

> Yes.

RHELPLAN-43084

seocam:
Regarding the integration with the certificate role:

I can foresee 3 scenarios:

1. The logging role will use exiting certificate
2. The logging role will upload new certificates (this PR)
3. The logging role will generate and manage certificates

The integration with the certificate role could happen in scenarios 1 and 3.
In scenario 3 the role would delegate the issuance and management of certs to the certificate role.
In scenario 1 the logging role would need to be included after the certificate role is included to ensure the files already exists.

I suppose it all depends to how tight and seamless we want this integration to happen.

nhosoi:
> Regarding the integration with the certificate role:
> 
> I can foresee 3 scenarios:
> 
>     1. The logging role will use exiting certificate
> 
>     2. The logging role will upload new certificates (this PR)
> 
>     3. The logging role will generate and manage certificates
> 
> 
> The integration with the certificate role could happen in scenarios 1 and 3.
> In scenario 3 the role would delegate the issuance and management of certs to the certificate role.
> In scenario 1 the logging role would need to be included after the certificate role is included to ensure the files already exists.
> 
> I suppose it all depends to how tight and seamless we want this integration to happen.

Let me add some more details to each scenario.

> 1. The logging role will use exiting certificate

This is an rsyslog upgrade scenario, I'd guess. On the target host, rsyslog is already configured with the certificate and runing the logging role makes some changes on the rsyslog config but keeps using the existing certificate.

>     2. The logging role will upload new certificates (this PR)

So far, the logging role does not do anything on the elasticsearch installation/configuration. It expects it's already up and running including the certificate setup. For rsyslog (the elasticsearch plugin) to talk to the elasticsearch, the certificate needs to be prepared on the host where ansible-playbook is executed and will be uploaded to the target host in executing the playbook. That's what's happening now.

>   3. The logging role will generate and manage certificates

Rsyslog works as a client as well as a server. In between it could use udp, plain tcp and tls enabled tcp. We need the certificate management for the tls enabled connection. Please note that this is independent from the elasticsearch case (this PR). The story would be something like using the logging role, we are deploying 10 rsyslog clients and 1 rsyslog server, whch are sharing the CA cert. It'd be great if we could generate the certificate in running the role and properly deployed. (sorry, not sure if we need client certs for this yet...)

roles/rsyslog/tasks/outputs/elasticsearch/main.yml
```diff
@@ -7,41 +7,34 @@
         dest: '{{ rsyslog_config_dir }}'
       with_items:
         - '{{ rsyslog_output_elasticsearch }}'
-      when:
-        - item.ca_cert_src is defined
-        - item.ca_cert_src | length > 0
+      when: item.ca_cert_crc | d()
```
richm, line 10: ca_cert_src

nhosoi: >_< Thank you, @richm. Pushed the fix...

roles/rsyslog/README.md
```diff
@@ -213,19 +213,24 @@ Common sub-variables
 
 Elasticsearch, Files and Forwards outputs sub-variables
 ----------------------------------
-- `elasticsearch`: array of dictionary to specify the parameters to forward the log messages to elasticsearch.
+- `use_local_omelasticsearch_cert` : Global variable. If set to 'true', local files ca_cert_src, cert_src and key_src in the elasticsearch type logging_outputs will be deployed to the remote host.
```
richm, line 216: `logging_use_local_elasticsearch_cert` - variable names must have a `logging_` prefix, and I don't think you need the `om` here.

pcahyna: What is a "sub-variable" anyway? I don't see it defined anywhere.
If a key in some dictionary is meant, then it does not need a `logging_` prefix (dictionary keys do not need to be namespaced, that would lead to very long and unreadable names.) But "Global variable" indicates that it is not the case.

roles/rsyslog/README.md
```diff
@@ -213,19 +213,24 @@ Common sub-variables
...
+- For the elasticsearch type logging_outputs, a set of following variables are to specify output elasticsearch configurations. It could be an array if multiple elasticsearch clusters to be configured.
```
pcahyna, line 217: Are those variables? They look more like dictionary keys (options).

tests/tests_files_elasticsearch_use_local_cert.yml
```diff
@@ -0,0 +1,86 @@
...
+            type: files
+            rsyslog_input_log_path: "{{ __test_inputfiles_dir }}/*.log"
+        logging_flows:
+          - name: flow_0
+            inputs: [files_input]
+            outputs: [elasticsearch_output, elasticsearch_output_ops]
+      include_role:
+        name: linux-system-roles.logging
+
+    - include: set_rsyslog_variables.yml
```
pcahyna, line 49: See the notes in https://docs.ansible.com/ansible/latest/modules/include_module.html#notes regarding nonintuitive behavior and possible future deprecation.

roles/rsyslog/README.md
nhosoi: > `logging_use_local_elasticsearch_cert` - variable names must have a `logging_` prefix, and I don't think you need the `om` here.

Ahhh, sorry, @richm. I'm planning to eliminate use_local_omelasticsearch_cert. Instead, using a cert or not is specified in each elasticsearch setting by `use_cert` (see the following example). @pcahyna, what'd be the good terminology to describe `use_cert` or that level of keys?
```
        logging_outputs:
          - name: elasticsearch_output
            type: elasticsearch
            server_host: logging-es
            server_port: 9200
            index_prefix: project.
            input_type: ovirt
            retryfailures: off
            use_cert: false
          - name: elasticsearch_output_ops
            type: elasticsearch
            server_host: logging-es-ops
            server_port: 9200
            index_prefix: .operations.
            input_type: viaq
            retryfailures: off
            use_cert: false
```

richm:
Ideally, the user would not have to specify `use_local_omelasticsearch_cert` - if the user specifies `ca_cert_src`, `cert_src`, and `key_src`, it is implied.

input: The user specifies none of the cert parameters
output: "Error: you must specify all of the cert/key files, or set `use_cert: false` in the elasticsearch output named xxx"

input: The user specifies none of the cert parameters and `use_cert: false`
output: success

input: The user specifies one or more of `ca_cert`, `cert`, `key`, `ca_cert_src`, `cert_src`, and `key_src`, and `use_cert: false`
output: "Error: you specified  `use_cert: false` and also specified one or more cert/key files."

input: The user specifies not enough cert/key files
output: "Error: you must specify all 3 of `ca_cert`, `cert`, `key`, or all 6 of `ca_cert`, `cert`, `key`, `ca_cert_src`, `cert_src`, `key_src`"

input: The user specifies all 3 of `ca_cert`, `cert`, `key` and nothing else
output: success - `use_local_omelasticsearch_cert` is implicitly set to `false`

input: The user specifies all 6 of `ca_cert`, `cert`, `key`, `ca_cert_src`, `cert_src`, `key_src`
output: success - `use_local_omelasticsearch_cert` is implicitly set to `true`

nhosoi:
@richm, @pcahyna, I've pushed https://github.com/linux-system-roles/logging/pull/117/commits/f32af0b7cffb6a82383c6ab44a2c71b61f3a62e7, in which these changes are made.
- Eliminating use_local_omelasticsearch_cert variable, which was a variable to indicate key/certificates on the local host to be copied to the target host. When local files are copied, the location of the key and certificates are specified in key_src, cert_src, and ca_cert_src. That should be good enough for the indication.
- An ambiguous word "sub-variable" is replaced with "dictionary key" in the docs.
- Replacing "include:" with "include_tasks" in the test playbooks.

roles/rsyslog/tasks/outputs/elasticsearch/main.yml
```diff
@@ -4,53 +4,77 @@
...
+          {{ [ item |
+               combine( { 'ca_cert': item.ca_cert | d(rsyslog_config_dir + '/' + item.ca_cert_src | basename),
+                          'cert': item.cert | d(rsyslog_config_dir + '/' + item.cert_src | basename),
+                          'key': item.key | d(rsyslog_config_dir + '/' + item.key_src | basename) } )
+          ] }}
+      with_items:
+        - '{{ rsyslog_output_elasticsearch }}'
+      when: item.ca_cert_src | d() or item.cert_src | d() or item.key_src | d()
+
+    - name: Check certs - use_cert is true, but no key/certs data are provided
```
richm, line 51: What is the difference between this check task and the check task `name: Check certs - triplets are not given` ?

nhosoi: You are right. There's no difference... :p

`triplets are not given` used to be for this checking.
> input: The user specifies not enough cert/key files
> output: "Error: you must specify all 3 of ca_cert, cert, key, or all 6 of ca_cert, cert, key, ca_cert_src, cert_src, key_src"

But I had to modify it to the current condition to support the following use case. Then, the 2 became identical... I'll delete the second one.
NOTE: allowing 3 of ca_cert_src, cert_src, key_src are specified, but not other 3.
In this case, the key/certs from the local host are copied to rsyslog_config_dir/{{ca_cert_drc|basename}}, etc.

roles/rsyslog/tasks/outputs/elasticsearch/main.yml
```diff
@@ -25,15 +25,15 @@
...
-        msg: "Error: you must specify all 3 of ca_cert, cert, key, or all 3 of ca_cert_src, cert_src, key_src in the elasticsearch output named {{ item.name }}"
+        msg: "Error: you must specify all 3 of ca_cert, cert, key, or all 3 of ca_cert_src, cert_src, key_src, or set use_cert: false in the elasticsearch output named {{ item.name }}"
       with_items:
         - '{{ rsyslog_output_elasticsearch }}'
       when:
         - item.use_cert | d(true)
-        - not ((item.ca_cert | d() and item.cert | d() and item.key | d()) or
-               (item.ca_cert_src | d() and item.cert_src | d() and item.key_src | d()))
+        - not (item.ca_cert | d() or item.cert | d() or item.key | d())
+        - not (item.ca_cert_src | d() or item.cert_src | d() or item.key_src | d())
```
richm, line 36: I don't think this new logic is correct.  I think the old logic is correct - fail if user does not supply all 3 of ca_cert/cert/key, or if the user does not supply all of ca_cert_src/cert_src/key_src

nhosoi:
Rebased against the master branch.

roles/rsyslog/tasks/outputs/elasticsearch/main.yml
nhosoi: I think this original code [1] works without ca_cert, cert and key. They local files ca_cert_src, cert_src and key_src are copied to {{ rsyslog_config_dir }} on the target host. And the paths are set to the rsyslog config params tls.cacert, tls.mycert, and tls.myprivkey, respectively, (I noticed there was an issue in the original code, e.g., if the file names are not es-ca.crt, es-cert.pem, or es-key.pem, rsyslog or if {{ rsyslog_config_dir }} is different from /etc/rsyslog.d, rsyslogd should issue the certs not found error, but it's not a point here :) The point is setting ca_cert_src, cert_src and key_src without ca_cert, cert and key was supported prior to this PR. Do you think it's not useful?
[1] - https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/tasks/outputs/elasticsearch/main.yml#L2-L53

richm: > The point is setting ca_cert_src, cert_src and key_src without ca_cert, cert and key was supported prior to this PR. Do you think it's not useful?

It is useful.  But I don't think that is related to the problem I see with the current code.  This is the current code:
```yaml
     when:
        - not (item.ca_cert | d() or item.cert | d() or item.key | d())
        - not (item.ca_cert_src | d() or item.cert_src | d() or item.key_src | d())
```
iiuc, this means that the user can set `item.ca_cert`, but not set `item.cert` or `item.key`, and the user can set `item.cert_src`, but not set `item.ca_cert_src` or `item.key_src`, and the `when` will report `true`.  Is this intended?

nhosoi: Thanks, @richm. My mistake. The fix was pushed.

tests/tests_files_elasticsearch_certs_incomplete.yml
```diff
@@ -0,0 +1,52 @@
...
+          logging_flows:
+            - name: flow_0
+              inputs: [files_input]
+              outputs: [elasticsearch_output, elasticsearch_output_ops]
+        include_role:
+          name: linux-system-roles.logging
+        register: result
+      rescue:
+        - debug:
+            msg: "Caught an expected error - {{ result }}"
```
richm, line 52: Does this mask the error?  That is, will the test report an error here, or do you need a separate `assert` or `fail` task?
https://docs.ansible.com/ansible/latest/user_guide/playbooks_blocks.html#blocks-error-handling
I'm not sure if `result` set in the `include_role` task in the `block` will be available in the `rescue` section.  You can use `ansible_failed_result` to get the result of the failed task.  Note that the page refers to `ansible_failed_task` but there is currently a bug and you cannot use this variable.

nhosoi: Thank you, @richm. That's exactly what I was looking for.
Also, rebased.

richm:
lgtm

roles/rsyslog/README.md
pcahyna: @nhosoi  Sorry for the late reply. In the network role, we call them "options" https://github.com/linux-system-roles/network/blob/master/README.md#options and I think this is the best option - also used in networking modules, see here: https://docs.ansible.com/ansible/latest/modules/junos_l3_interfaces_module.html#parameter-config. But the terminology can be cleaned up later.

roles/rsyslog/README.md
```diff
@@ -5,22 +5,6 @@ linux-system-roles-rsyslog
...
-   * [Deploy Rsyslog Configuration Files](#deploy)
-      * [Inventory File](#inventory-file)
-      * [vars.yml](#vars)
-      * [playbook.yml](#playbook)
-   * [Variables in vars.yml](#variables)
-      * [Common sub-variables](#common-variables)
-      * [Viaq sub-variables](#viaw)
-   * [Contents of Roles](#roles)
-<!--te-->
-
```
pcahyna, line 23: Was it intended to drop the table of contents?

nhosoi: Yeah, it the table of contents are obsolete... I think I should update the READMEs more readable and useful for the endusers...

nhosoi:
> > Eliminating use_local_omelasticsearch_cert variable
> > An ambiguous word "sub-variable" is replaced with "dictionary key" in the docs.
> > Replacing "include:" with "include_tasks" in the test playbooks.
> 
> Those three changes seem totally unrelated and should be split into separate commits. The third one should be limited to a fixup of [2b8b12d](https://github.com/nhosoi/logging/commit/2b8b12d6ef2fbfab41484678f4b708f844e02631) (to fix [linux-system-roles@2b8b12d#r426908015](https://github.com/linux-system-roles/logging/commit/2b8b12d6ef2fbfab41484678f4b708f844e02631#r426908015) ) and the fixes to the other similar places should are better done in a separate PR.

Do you think making another PR for the docs would be good enough? That'd be quick and very few chances for me to introduce mistakes.

pcahyna:
> > > Eliminating use_local_omelasticsearch_cert variable
> > > An ambiguous word "sub-variable" is replaced with "dictionary key" in the docs.
> > > Replacing "include:" with "include_tasks" in the test playbooks.
> > 
> > 
> > Those three changes seem totally unrelated and should be split into separate commits. The third one should be limited to a fixup of [2b8b12d](https://github.com/nhosoi/logging/commit/2b8b12d6ef2fbfab41484678f4b708f844e02631) (to fix [linux-system-roles@2b8b12d#r426908015](https://github.com/linux-system-roles/logging/commit/2b8b12d6ef2fbfab41484678f4b708f844e02631#r426908015) ) and the fixes to the other similar places should are better done in a separate PR.
> 
> Do you think making another PR for the docs would be good enough? That'd be quick and very few chances for me to introduce mistakes.

At least this:
https://github.com/nhosoi/logging/commit/f7de1cb22d1530e58c90d8e3036b884b810cd8d3#diff-1c7bee374b17a63912b35d50adb51f95L277-L279
and 
https://github.com/nhosoi/logging/commit/f7de1cb22d1530e58c90d8e3036b884b810cd8d3#diff-1c7bee374b17a63912b35d50adb51f95L263
and
https://github.com/nhosoi/logging/commit/f7de1cb22d1530e58c90d8e3036b884b810cd8d3#diff-04c6e90faac2675aa89e2176d2eec7d8
should be included with the commit "Eliminating use_local_omelasticsearch_cert variable" because they are related, or at least in the same PR.

The rest seems unrelated and can be done in a separate PR.

nhosoi:
Thank you for the clarification, @pcahyna !!

roles/rsyslog/README.md
nhosoi: @pcahyna, Thank you for the right word. I agree I should change the terminology later with more correction candidates. Also, I should restructure the docs better. I'm opening a RHELPLAN for that. Thanks! > @richm

nhosoi:
> > > > Eliminating use_local_omelasticsearch_cert variable
> > > > An ambiguous word "sub-variable" is replaced with "dictionary key" in the docs.
> > > > Replacing "include:" with "include_tasks" in the test playbooks.
> > > 
> > > 
> > > Those three changes seem totally unrelated and should be split into separate commits. The third one should be limited to a fixup of [2b8b12d](https://github.com/nhosoi/logging/commit/2b8b12d6ef2fbfab41484678f4b708f844e02631) (to fix [linux-system-roles@2b8b12d#r426908015](https://github.com/linux-system-roles/logging/commit/2b8b12d6ef2fbfab41484678f4b708f844e02631#r426908015) ) and the fixes to the other similar places should are better done in a separate PR.
> > 
> > 
> > Do you think making another PR for the docs would be good enough? That'd be quick and very few chances for me to introduce mistakes.
> 
> At least this:
> [nhosoi@f7de1cb#diff-1c7bee374b17a63912b35d50adb51f95L277-L279](https://github.com/nhosoi/logging/commit/f7de1cb22d1530e58c90d8e3036b884b810cd8d3#diff-1c7bee374b17a63912b35d50adb51f95L277-L279)
> and
> [nhosoi@f7de1cb#diff-1c7bee374b17a63912b35d50adb51f95L263](https://github.com/nhosoi/logging/commit/f7de1cb22d1530e58c90d8e3036b884b810cd8d3#diff-1c7bee374b17a63912b35d50adb51f95L263)
> and
> [nhosoi@f7de1cb#diff-04c6e90faac2675aa89e2176d2eec7d8](https://github.com/nhosoi/logging/commit/f7de1cb22d1530e58c90d8e3036b884b810cd8d3#diff-04c6e90faac2675aa89e2176d2eec7d8)
> should be included with the commit "Eliminating use_local_omelasticsearch_cert variable" because they are related, or at least in the same PR.
> 
> The rest seems unrelated and can be done in a separate PR.

Thanks, @pcahyna. Done!

tests/tests_files_elasticsearch_certs_incomplete.yml
```diff
@@ -46,7 +46,9 @@
...
         include_role:
           name: linux-system-roles.logging
-        register: result
       rescue:
         - debug:
-            msg: "Caught an expected error - {{ result }}"
+            msg: "Caught an expected error - {{ ansible_failed_result.results }}"
+        - name: assert...
+          assert:
+            that: "'{{ ansible_failed_result.results.0.msg }}' is match('Error: you specified use_cert: true; you must specify all 3 of ca_cert, cert, key, or all 3 of ca_cert_src, cert_src, key_src, or set use_cert: false in the elasticsearch output named elasticsearch_output')"
```
pcahyna, line 54: This way you check that if the role has failed, it failed with the expected message. But you don't seem to check whether the role has actually failed at all (`rescue` would not be executed).

pcahyna:
@nhosoi, commit https://github.com/linux-system-roles/logging/pull/117/commits/f830444daa14eea6af17a286e521dbb2d197b63a still has three independent changes, will you please split it and leave the `include_tasks:` parts for another commit or PR? Before you merge this PR, please also squash the last two fixups (reverting doc changes and fixing a rebase error) into commits where they belong.

nhosoi:
> @nhosoi, commit [f830444](https://github.com/linux-system-roles/logging/commit/f830444daa14eea6af17a286e521dbb2d197b63a) still has three independent changes, will you please split it and leave the `include_tasks:` parts for another commit or PR? Before you merge this PR, please also squash the last two fixups (reverting doc changes and fixing a rebase error) into commits where they belong.

https://github.com/linux-system-roles/logging/pull/122

tests/tests_files_elasticsearch_certs_incomplete.yml
pcahyna: Here's how to check this: https://github.com/pcahyna/storage/blob/787f00cba863a8f87883be6830c1e26488dadd48/tests/tests_disk_errors.yml#L29 .

nhosoi: Thanks a lot, @pcahyna. I added your UNREACH check to the test playbooks.843259

nhosoi:
@richm, @pcahyna, thank you for reviewing and approving this pr. Merging.

PR #118, opened 2020-05-18
Getting rid of cleanup.yml

nhosoi:
There is no use case for cleanup. Besides, it's implemented only for
the input tasks, not for the output tasks.


nhosoi:
Thank you, @richm. Merging.

pcahyna:
Where do you handle `state: absent` ?

richm:
> Where do you handle `state: absent` ?

It isn't handled - let's do this in a separate PR

nhosoi:
> > Where do you handle `state: absent` ?
> 
> It isn't handled - let's do this in a separate PR

Right. These cleanup.yml tasks were blindly deleting input files regardless of the flow relationships. We need to delete files assuring the healthy configuration... Opening a new RHELPLAN/PR.

pcahyna:
As it was an interface change (the `absent` value for `state` ceased to be supported), it would have been better to mention it more prominently in the PR description.

nhosoi:
> As it was an interface change (the `absent` value for `state` ceased to be supported), it would have been better to mention it more prominently in the PR description.

Sorry. This will be replaced with a new code. At that time, I'll emphasize it.

PR #119, opened 2020-05-21
RHELPLAN-33869 - Eliminating the variable rsyslog_default

nhosoi:
Removing rsyslog_default variable and stop copying the original rsyslog.conf.

Also, the following changes are made.
- moving the location of rsyslog.conf.j2 and rules.conf.j2 to roles/rsyslog/templates.
- changing the way to check the generated rsyslog.conf size from counting the lines
  to using the stat.size.
- changing the rsyslog version check from the version line in the config file to
  the internal variable __rsyslog_version.


design_docs/rsyslog_inputs_outputs_flows.md
```diff
@@ -13,12 +13,20 @@ This proposal will discuss about these items:
 - how to replace the `rsyslog_default` variable with the logging role inputs/outputs configuration.
 - how to support more generic way to connect inputs and outputs.
 
-Please note that this github commit starting this doc implements the first cut of the proposal including:
-- inputs - basics input role (`imjournal`, `imuxsock`, `imudp`, `imptcp`) and files input role (`imfile`) [0], and
-- outputs - files output role (`omfile`), forwards output role (`omfwd`) and elasticsearch role (`omelasticsearch`).
-Leftover tasks, other roles to be changed and eliminating `rsyslog_default`, are discussed in [To Do's](#to-dos).
+## Post Flos Control Implementation
```
richm, line 16: Flow

roles/rsyslog/templates/rsyslog.conf.j2
```diff
@@ -0,0 +1,6 @@
+# {{ ansible_managed }}
+# rsyslog version is {{ __rsyslog_version }}.
```
richm, line 2: I don't think we want to have `__rsyslog_version` here - having it here will mean the file will be updated unnecessarily every time rsyslog is updated to a new version

nhosoi: It was controversial and never liked by @pcahyna. :)
With the change in this PR, CI tests stop referring it. So, if it is ok to check the version using the internal __rsyslog_version in the CI test playbook, I'm happy to delete it.

richm: Can you show me how you are using `__rsyslog_version` ?

nhosoi: In the test playbook, __rsyslog_version is set to "N/A" first.
https://github.com/nhosoi/logging/blob/remove_rsyslog_default/tests/tests_enabled.yml#L4-L5
Then,after Logging role was run, it checks the value is no longer "N/A" or "0.0.0" that indicates something is wrong happened on the rsyslog package.
https://github.com/nhosoi/logging/blob/remove_rsyslog_default/tests/tests_enabled.yml#L14-L16

richm: ok - then yes, please remove from templates/rsyslog.conf.j2

nhosoi:
Fixing the conflicts and squashing...

richm:
lgtm

nhosoi:
[citest bad]

nhosoi:
There were conflict resolution errors in tests_remote_default_remote.yml and tests_remote_remote.yml. Fixed them and pushed.

richm:
lgtm

nhosoi:
Thank you for reviewing the PR, @richm.
Merging.

tests/tests_default.yml
```diff
@@ -1,8 +1,14 @@
...
+    __rsyslog_version: "N/A"
 
   tasks:
     - name: default run (NOOP)
       include_role:
         name: linux-system-roles.logging
+
+    - name: Check rsyslog version
+      assert:
+        that: __rsyslog_version != "N/A" and __rsyslog_version != "0.0.0"
```
pcahyna, line 14: Sorry for being late to the party, but does this assert belong here? For all other roles the default test just ensures that the role does not crash when no parameters are supplied. Is there a reason for the logging role to be different? Also, is it a good idea to test for an internal variable? I think it would be better to check the actual effect of the role on the system in tests rather than checking that the role has set some variable which the user will not use anyway - such a test is rather artificial.

nhosoi: > Sorry for being late to the party, but does this assert belong here? For all other roles the default test just ensures that the role does not crash when no parameters are supplied. Is there a reason for the logging role to be different? Also, is it a good idea to test for an internal variable? I think it would be better to check the actual effect of the role on the system in tests rather than checking that the role has set some variable which the user will not use anyway - such a test is rather artificial.

Thanks for your comments, @pcahyna. They are valuable at any timing!
1) I will remove the check from tests_defaults.yml, which is not necessary.
2) You might remember that in the beginning, I was writing the version number at the top of rsyslog.conf anc checking the value in the test playbooks, which you recommended to replace with some other way. This is the second attempt. All I want to check is the version is set to the internal variable __rsyslog_version and the value is valid... If I have to, the third attempt might be using the `fail` module in the task (not in the test code) - if the value of __rsyslog_version looks invalid, we could fail there. Do we want to make the change?

pcahyna: @nhosoi I don't understand the importance of checking, in the tests, that the value of `__rsyslog_version` is valid. The variable is not going to be used by the users, why is it important that it is valid?
If you want to make sure it is being set in a sensible way, you can add a separate unit test that checks for it (the current test are more integration tests, but adding a unit test is valid, too. I just don't want to mix those two test types).

pcahyna: I think though that it would be good to review first the need for this variable - can't it be simply eliminated?

nhosoi: > I think though that it would be good to review first the need for this variable - can't it be simply eliminated?

It is likely needed in this effort.
RHELPLAN-32352 - Add performance tuning parameters
Q. how do we handle rhel-7 and rhel-8 version difference?

nhosoi: Opened a PR https://github.com/linux-system-roles/logging/pull/125.
Executing "Check rsyslog version" test just in tests_version.yml and eliminating it from the rest of the test cases.

richm: >  Q. how do we handle rhel-7 and rhel-8 version difference?

https://github.com/oasis-roles/meta_standards/pull/16/files#diff-04c6e90faac2675aa89e2176d2eec7d8R184

so you would have `vars/RedHat_7.yml` and `vars/RedHat_8.yml` where you would set platform/version specific variables.  You could do the same with tasks e.g. have `tasks/setup_RedHat_7.yml` etc.  If CentOS works exactly the same way, then you could just symlink `vars/CentOS_7.yml` to `vars/RedHat_7.yml` etc.

I suppose if there are differences between RHEL 7.8 and 7.9, or RHEL 8.2 and RHEL 8.3, we could use `ansible_distribution_minor_version` too

pcahyna: If there are differences between minor versions, it may be in fact preferable to check the version of rsyslog rather than the distribution minor version. (What if the user updates just some of packages?) Is it really the case though? If not, it would be better to discuss the problem only when it really occurs.

PR #120, opened 2020-05-21
.travis.yml - Upgrading fedora versions to test from {30,31} to {31,32}.

nhosoi:


.travis.yml
```diff
@@ -7,8 +7,8 @@ env:
   matrix:
     - MOLECULE_DISTRO="centos:7"
     - MOLECULE_DISTRO="centos:8"
-    - MOLECULE_DISTRO="fedora:30"
```
richm, line 10: Is there some problem with fedora 30?

pcahyna: I guess the idea is that it is unsupported, or soon will be. We are removing it form the test-harness CI as well.
For now lgtm, but in a longer term it would be preferable to use the template and do those changes centrally in the template.

nhosoi: > Is there some problem with fedora 30?

There is... Somehow, python3 is used just for fedora 30??? (I thought it was because it was no longer supported... Maybe, my guess was wrong...)
https://travis-ci.org/github/linux-system-roles/logging/builds/689452565

richm:
lgtm - we'll do integration with https://github.com/linux-system-roles/template in a later commit

nhosoi:
> lgtm - we'll do integration with https://github.com/linux-system-roles/template in a later commit

Thanks, @richm. Opened RHELPLAN-44234.

.travis.yml
pcahyna: > > Is there some problem with fedora 30?
> 
> There is... Somehow, python3 is used just for fedora 30??? (I thought it was because it was no longer supported... Maybe, my guess was wrong...)
> https://travis-ci.org/github/linux-system-roles/logging/builds/689452565

It is not - https://travis-ci.org/github/linux-system-roles/logging/jobs/689452569 show that the same version of Python is used for fedora31. Why do you think the problem is in python3? 

> Hmmm, it might have been a temporary(?) failure. This run was successful... _puzzled_
> https://travis-ci.org/github/linux-system-roles/logging/builds/689530970

The actual error message is "ConnectionResetError: [Errno 104] Connection reset by peer" in ssl.py. This looks very much like a temporary network failure indeed, instead of being related to python3.

Could you please tell us why you need to do some change in the PR description instead of letting us guess?

PR #123, opened 2020-05-28
Fixing a bug - eliminating an rsyslog private variable $.syslogtag, w…

nhosoi:
…hich is no longer used.


nhosoi:
Thank you for your approval, @richm. Merging.

PR #124, opened 2020-05-28
Making "state: absent" work.

nhosoi:
When an item in logging_inputs and outputs has an absent state,
(assuming the to-be-generated config file name is XX-filename.conf)
case 1) XX-filename.conf does not exist.
logging role does not generate XX-filename.conf
case 2) XX-filename.conf exists.
logging role deletes XX-filename.conf

Questionable case) XX-filename.conf exists; instead of having the
item with the absent state, delete the item from logging_inputs or
outputs. Then, the existing XX-filename.conf remains untouched.

Adding a test playbook tests_combination_absent.yml

### DIFF
--- a/tests/tests_combination_absent.yml
+++ b/tests/tests_combination_absent.yml
@@ -1,0 +1,174 @@
+- name: Running logging role twice; in the second run, some logging_outputs are absent or eliminated.
+  hosts: all
+  become: true
+  vars:
+    __rsyslog_version: "N/A"
+    __test_tag: files_input
+    __test_inputfiles_conf: /etc/rsyslog.d/90-input-files-{{ __test_tag }}.conf
+    __test_inputfiles_dir: /var/log/inputdirectory
+    __test_basics_conf: 90-input-basics-basic_input.conf
+
+  tasks:
+    - name: Generate an input log dir {{ __test_inputfiles_dir }}
+      file:
+        path: "{{ __test_inputfiles_dir }}"
+        state: directory
+        mode: 0700
+
+    - name: deploy config to output into local files
+      vars:
+        logging_outputs:
+          - name: files_test0
+            type: files
+            severity: info
+            exclude:
+              - authpriv.none
+              - auth.none
+              - cron.none
+              - mail.none
+            path: /var/log/messages
+          - name: files_test1
+            type: files
+            facility: authpriv,auth
+            path: /var/log/secure
+          - name: forwards_severity_and_facility
+            type: forwards
+            facility: local1
+            severity: info
+            protocol: tcp
+            target: host.domain
+            port: 1514
+          - name: forwards_facility_only
+            type: forwards
+            facility: local2
+            protocol: tcp
+            target: host.domain
+            port: 2514
+        logging_inputs:
+          - name: basic_input
+            type: basics
+            rsyslog_imjournal_ratelimit_burst: 33333
+          - name: "{{ __test_tag }}"
+            type: files
+            rsyslog_input_log_path: "{{ __test_inputfiles_dir }}/*.log"
+        logging_flows:
+          - name: flow_0
+            inputs: [basic_input]
+            outputs: [files_test0, files_test1, forwards_severity_and_facility, forwards_facility_only]
+          - name: flow_1
+            inputs: ["{{ __test_tag }}"]
+            outputs: [files_test0, files_test1]
+      include_role:
+        name: linux-system-roles.logging
+
+    - include_tasks: set_rsyslog_variables.yml
+
+    # notify restart rsyslogd is executed at the end of this test task.
+    # thus we have to force to invoke handlers
+    - name: Force all notified handlers to run at this point, not waiting for normal sync points
+      meta: flush_handlers
+
+    - name: Check rsyslog version
+      assert:
+        that: __rsyslog_version != "N/A" and __rsyslog_version != "0.0.0"
+
+    - name: Check rsyslog.conf size
+      assert:
+        that: rsyslog_conf_stat.stat.size < 1000
+
+    - name: Check file counts in rsyslog.d
+      assert:
+        that: rsyslog_d_file_count.matched >= 11
+
+    - name: generate test config file
+      copy:
+        dest: "/etc/rsyslog.d/{{ __test_basics_conf }}"
+        content: |-
+          # GENERATED BASICS CONFIG FILE
+          module(load="imuxsock"    # provides support for local system logging (e.g. via logger command)
+                 SysSock.Use="off") # Turn off message reception via local log socket.
+          module(load="imjournal"
+                 StateFile="/var/lib/rsyslog/imjournal.state"
+                 RateLimit.Burst="20000"
+                 RateLimit.Interval="600"
+                 PersistStateInterval="10")
+          if
+            ($inputname == "imjournal")
+            then {
+              call files_test0
+          }
+
+    # Checking 'error' in stdout from systemctl status is for detecting the case in which rsyslog is running,
+    # but some functionality is disabled due to some error, e.g., error: 'tls.cacert' file couldn't be accessed.
+    - name: Check rsyslog errors
+      command: systemctl status rsyslog
+      register: systemctl_result
+      failed_when: "'error' in systemctl_result.stdout or systemctl_result.rc != 0"
+
+    - name: deploy reduced config to output into local files
+      vars:
+        logging_outputs:
+          - name: files_test0
+            type: files
+            severity: info
+            exclude:
+              - authpriv.none
+              - auth.none
+              - cron.none
+              - mail.none
+            path: /var/log/messages
+          - name: files_test1
+            type: files
+            facility: authpriv,auth
+            path: /var/log/secure
+            state: absent
+          - name: forwards_severity_and_facility
+            type: forwards
+            facility: local1
+            severity: info
+            protocol: tcp
+            target: host.domain
+            port: 1514
+            state: absent
+          - name: forwards_facility_only
+            type: forwards
+            facility: local2
+            protocol: tcp
+            target: host.domain
+            port: 2514
+            state: absent
+        logging_inputs:
+          - name: basic_input
+            type: basics
+            rsyslog_imjournal_ratelimit_burst: 33333
+          - name: "{{ __test_tag }}"
+            type: files
+            rsyslog_input_log_path: "{{ __test_inputfiles_dir }}/*.log"
+            state: absent
+        logging_flows:
+          - name: flow_0
+            inputs: [basic_input]
+            outputs: [files_test0]
+      include_role:
+        name: linux-system-roles.logging
+
+    - include_tasks: set_rsyslog_variables.yml
+
+    # notify restart rsyslogd is executed at the end of this test task.
+    # thus we have to force to invoke handlers
+    - name: Force all notified handlers to run at this point, not waiting for normal sync points
+      meta: flush_handlers
+
+    - name: Check file counts in rsyslog.d are less than no-absent case
+      assert:
+        that: rsyslog_d_file_count.matched < 11
+
+    - name: Check {{ __test_basics_conf }} was updated
+      command: '/bin/grep "# GENERATED BASICS CONFIG FILE" /etc/rsyslog.d/{{ __test_basics_conf }}'
+      register: result
+      failed_when: result.rc == 0
+
+    - name: Check rsyslog errors
+      command: systemctl status rsyslog
+      register: systemctl_result
+      failed_when: "'error' in systemctl_result.stdout or systemctl_result.rc != 0"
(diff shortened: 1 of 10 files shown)


tests/tests_combination_absent.yml
```diff
@@ -0,0 +1,155 @@
...
+          # GENERATED BASICS CONFIG FILE
+          if ($inputname == "imjournal") then {
+              call files_test0
+          }
+      delegate_to: localhost
+
+    - name: Update it to the target
+      copy:
+        src: "/tmp/{{ __test_basics_conf }}"
+        dest: "/etc/rsyslog.d/{{ __test_basics_conf }}"
```
richm, line 96: Do you need to create a local file first, before you copy it?  Can you remove the `name: generate test config file` task, and just do
```yaml

    - name: generate test config file
      copy:
        dest: "/etc/rsyslog.d/{{ __test_basics_conf }}"
        content: |-
          # GENERATED BASICS CONFIG FILE
          if ($inputname == "imjournal") then {
              call files_test0
          }
```
?

nhosoi: Thank you, @richm. Fixing it.
Strangely, CI tests are failing on RHEL/CentOS-7 (not on 8 or Fedoras...) Invetigating it.

nhosoi: @richm, this might be really problematic.
```
Questionable case) XX-filename.conf exists; instead of having the
item with the absent state, delete the item from logging_inputs or
outputs. Then, the existing XX-filename.conf remains untouched.
```
In tests_combination_absent.yml, it does a full deployment in the first run; in the second run's config, some dict items are set to `absent` and some dict items are removed. The second run successfully removes the config files to be generated from the `absent` items, but it leaves config files from the dict items removed in the second run. It causes this error.
`May 28 22:30:44 ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com rsyslogd[14038]: error during config processing: ruleset 'files_test1' cannot be found  [v8.24.0-52.el7 try http://www.rsyslog.com/e/2207 ]`

Actually, there is another issue in RHEL/CentOS-7. systemd package puts listen.conf in /etc/rsyslog.d.
```
# rpm -qf /etc/rsyslog.d/listen.conf 
systemd-219-62.el7_6.9.x86_64
```
It expects imuxsock, but we set it off.
```
# cat /etc/rsyslog.d/listen.conf 
$SystemLogSocketName /run/systemd/journal/syslog
```
It makes rsyslog issue this error (although rsyslog is up and running.)
`May 28 22:30:44 ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com rsyslogd[14038]: command 'SystemLogSocketName' is currently not permitted - did you already set it via a RainerScript command (v6+ config)? [v8.24.0-52.el7 try http://www.rsyslog.com/e/2222 ]`

The easiest way to solve the above 2 issues is to set `logging_purge_confs: true` in the second run, which cleans up all existing files in /etc/rsyslog.d and starts from scratch. But it may cause some other issue such as there's another config file `21-cloudinit.conf` placed by `cloud-init-18.2-1.el7.centos.2.x86_64` which we may not want to delete...

For instance, is it acceptable to have a black list for the task handling logging_purge_confs? Can we put listen.conf and config files starting with ##-input- or ##-output- which are ours? Is it too ugly?
```
# cat 21-cloudinit.conf 
# Log cloudinit generated log messages to file
:syslogtag, isequal, "[CLOUDINIT]" /var/log/cloud-init.log
```

nhosoi: Please note that now all the CI tests are failing since I added the systemctl status error check to tests_combination_absent.yml.

richm: > The second run successfully removes the config files to be generated from the absent items, but it leaves config files from the dict items removed in the second run.

I'm not sure what you mean.

> listen.conf
```
# cat /etc/rsyslog.d/listen.conf 
$SystemLogSocketName /run/systemd/journal/syslog
```
so systemd adds a file that assumes rsyslog is loading the `imuxsock` module?

one way to solve it would be to always load the `imuxsock` module.  I guess that's part of the rsyslog "public API" that we can't easily change, especially for EL7.  Do we have the same problem on EL8?

richm: > Please note that now all the CI tests are failing since I added the systemctl status error check to tests_combination_absent.yml.

Then I guess the systemctl status error check is working  e.g. https://fedorapeople.org/groups/linuxsystemroles/logs/linux-system-roles-logging-pull-linux-system-roles_logging-124-3f46aca-fedora-32-20200529-000038/artifacts/ansible.log
```
TASK [Check rsyslog errors] ****************************************************
task path: /tmp/tmpgf02lvy_/tests/tests_combination_absent.yml:160
fatal: [/cache/Fedora-Cloud-Base-32-1.6.x86_64.qcow2]: FAILED! => {"changed": true, "cmd": ["systemctl", "status", "rsyslog"], ... systemd[1]: Starting System Logging Service...\nMay 29 00:10:22 ibm-p8-kvm-03-guest-02.virt.pnr.lab.eng.rdu2.redhat.com rsyslogd[16994]: error during parsing file /etc/rsyslog.conf, on or before line 6: ruleset 'files_test1' cannot be found 
```
that looks like a legitimate error message.

nhosoi:
> > The second run successfully removes the config files to be generated from the absent items, but it leaves config files from the dict items removed in the second run.
>
> I'm not sure what you mean.

Assuming we have existing A.conf and B.conf, both were generated by logging role. In the next logging role run with the updated playbook, if the dict to generate A.conf is set "state: absent", A.conf is removed. On the other hand, if the dict that generated B.conf is just removed from the playbook, B.conf remains intact which could cause some unexpected results. Is it good enough for us to write this fact in the Doc to suggest to set "state: absent" to remove a config file in the update?

> > listen.conf
>
> ```
> # cat /etc/rsyslog.d/listen.conf
> $SystemLogSocketName /run/systemd/journal/syslog
> ```
>
> so systemd adds a file that assumes rsyslog is loading the `imuxsock` module?

Yes, we do in the basics conf, but with SysSock.Use="off"

module(load="imuxsock"    # provides support for local system logging (e.g. via logger command)
       SysSock.Use="off") # Turn off message reception via local log socket.

> one way to solve it would be to always load the `imuxsock` module.

Just loading does not help, but we need to set SysSock.Use to "on". Then, the message `"command 'SystemLogSocketName' is currently not permitted - did you already set it via a RainerScript command"` disappears. But, that means, instead of imjournal, we use imuxsock. Do we want to enable imuxsock for rhel-7 and imjournal for rhel-8?

It looks this is an old issue.
https://github.com/rsyslog/rsyslog/issues/2150

But I'd rather like to remove listen.conf and use imjournal in rhel-7, as well. (Or do we live with the message `"command 'SystemLogSocketName' is currently not permitted - did you already set it via a RainerScript command"`? Although the message appears in the rsyslog output, rsyslog is working with imjournal.)

> I guess that's part of the rsyslog "public API" that we can't easily change, especially for EL7. Do we have the same problem on EL8?

No, EL8 systemd does not add listen.conf.

richm:
> On the other hand, if the dict that generated B.conf is just removed from the playbook, B.conf remains intact which could cause some unexpected results. Is it good enough for us to write this fact in the Doc to suggest to set "state: absent" to remove a config file in the update?

Yes.  It is the responsibility of the user to determine if it is safe to remove A.conf while leaving B.conf intact.  We just need to document this fact to the user.

> But, that means, instead of imjournal, we use imuxsock. Do we want to enable imuxsock for rhel-7 and imjournal for rhel-8?

What happens by default, on rhel7, if you install rsyslog and systemd, without doing any configuration?  Is rsyslog enabled to listen to `/run/systemd/journal/syslog` using `imuxsock`?  Does rsyslog emit a warning or error message, but continues to work?

nhosoi:
> > But, that means, instead of imjournal, we use imuxsock. Do we want to enable imuxsock for rhel-7 and imjournal for rhel-8?
> 
> What happens by default, on rhel7, if you install rsyslog and systemd, without doing any configuration? Is rsyslog enabled to listen to `/run/systemd/journal/syslog` using `imuxsock`? Does rsyslog emit a warning or error message, but continues to work?

Found this comment by Rainer!
https://github.com/rsyslog/rsyslog/issues/2150#issuecomment-348896583
> it looks like going back to old-style for imuxsock load seems the only solution currently for the way the RH journal packages works.

The default rsyslog.conf in RHEL-7 uses old rsyslog config syntax.
```
$ModLoad imuxsock # provides support for local system logging (e.g. via logger command)
$ModLoad imjournal # provides access to the systemd journal
# Turn off message reception via local log socket;
# local messages are retrieved through imjournal now.
$OmitLocalLogging on
```
Using this old-style, this additional socket path from listen.conf is accepted and used.
`$SystemLogSocketName /run/systemd/journal/syslog`
But according to the comment, it does not work with the new syntax.
```
module(
    load="imuxsock"
    sysSock.use="off"
)
```
To make it work, we have to add this option to the module call. The message `"command 'SystemLogSocketName' is currently not permitted - did you already set it via a RainerScript command"` indicates it.
`    sysSock.name="/run/systemd/journal/syslog"`

We allow enabling imuxsock instead of imjournal. I guess we could set the socket path to "/run/systemd/journal/syslog" for rhel-7, only? 
https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/templates/input_basics.j2#L7

richm:
It seems to me that in order to support `imuxsock`, we'll have to use the old-style config on el7.  I don't think we should try to do anything with files we do not own such as `/etc/rsyslog.d/listen.conf`.

In fact, do we need to support `imuxsock` in 8.3?  Are there any requirements to support it?

nhosoi:
> It seems to me that in order to support `imuxsock`, we'll have to use the old-style config on el7. I don't think we should try to do anything with files we do not own such as `/etc/rsyslog.d/listen.conf`.
> 
> In fact, do we need to support `imuxsock` in 8.3? Are there any requirements to support it?

@jvymazal may have the answer...
At least, the rsyslog.conf file in the rpm package has this module call with the socket turned off, which is basically the same configuration as in rhel-7. Just the style difference.
```
module(load="imuxsock"    # provides support for local system logging (e.g. via logger command)
       SysSock.Use="off") # Turn off message reception via local log socket; 
                          # local messages are retrieved through imjournal now.
```
And if I set SysSock.Use to "on", then it works as expected. That is, RHEL-8 uses the default socket path `/dev/log` instead of `/run/systemd/journal/syslog`, listen.conf is no longer needed to alternate the path.

I think I'm going to try the old style ModLoad imuxsock just for rhel-7 (in a separate pr ;).

nhosoi:
Updated the test tests_combination_absent.yml https://github.com/linux-system-roles/logging/pull/124/commits/d8d1ce25d6f61aeba302bf12679dffb20e9b5285.

Rebasing against PR https://github.com/linux-system-roles/logging/pull/127, it passes the CI test.

jvymazal:
> > It seems to me that in order to support `imuxsock`, we'll have to use the old-style config on el7. I don't think we should try to do anything with files we do not own such as `/etc/rsyslog.d/listen.conf`.
> > In fact, do we need to support `imuxsock` in 8.3? Are there any requirements to support it?
> 
> @jvymazal may have the answer...
> At least, the rsyslog.conf file in the rpm package has this module call with the socket turned off, which is basically the same configuration as in rhel-7. Just the style difference.
> 
> ```
> module(load="imuxsock"    # provides support for local system logging (e.g. via logger command)
>        SysSock.Use="off") # Turn off message reception via local log socket; 
>                           # local messages are retrieved through imjournal now.
> ```
> 
> And if I set SysSock.Use to "on", then it works as expected. That is, RHEL-8 uses the default socket path `/dev/log` instead of `/run/systemd/journal/syslog`, listen.conf is no longer needed to alternate the path.
> 
> I think I'm going to try the old style ModLoad imuxsock just for rhel-7 (in a separate pr ;).
I do not think we have to have the imuxsock supported for the initial release, it can be always added later on and it is meant mainly for performance critical use-cases where imjournal is simply too slow. Otherwise there is not so much need for it. 

The unfortunate thing is the mess with other packages like systemd having their old-style snippets, which are not really compatible with the modern config syntax. Another unfortunate thing is that messing with those files is probably going to cause some problems in the long run where that may be replaced/brought back by update of that package, so one way is to accept and document this, or making workarounds so those files do not issue the errors, which complicates things for us.

nhosoi:
> The unfortunate thing is the mess with other packages like systemd having their old-style snippets, which are not really compatible with the modern config syntax. Another unfortunate thing is that messing with those files is probably going to cause some problems in the long run where that may be replaced/brought back by update of that package, so one way is to accept and document this, or making workarounds so those files do not issue the errors, which complicates things for us.

Thanks for your responds, @jvymazal ! I think another PR https://github.com/linux-system-roles/logging/pull/127 does what you recommend. Could you please take a look and comment on it? Thanks in advance.

nhosoi:
[citest bad]

nhosoi:
@richm, @pcahyna, could you please review this PR one more time?

nhosoi:
Thank you for the approval, @richm. Squashed the 3 commits into one. Once CI-tests have passed. I'm going to merge this pr.

PR #125, opened 2020-05-28
Executing "Check rsyslog version" test just in tests_version.yml

nhosoi:
and eliminating it from the rest of the test cases.


nhosoi:
Thank you for your reviews and approvals, @richm and @pcahyna. Merging.

PR #126, opened 2020-05-28
RHELPLAN-43536 - Support tls for the remote input and forwards output

nhosoi:
- New variables for configuring tls
  - logging_pki - Specifying rsyslog encryption library
                  one of the implementations is allowed none, ptcp, tls, gtls, gnutls
                  Note: none->ptcp, tls,gtls,gnutls->gtls
                  Default to ptcp (== no tls)
                  Note: renaming logging_encryption from the previous PR121.
  - logging_pki_files - list of pki files dict
    logging_pki_files:
      - type: ca_cert | cert | key
        src:  location of the file on the local host;
              if given, the file is deployed to dest value path.
        dest: path to be deployed on the target host;
              if given, the path is set to the config file.
              Default to /etc/pki/tls/certs/ca.crt for ca_cert
                         /etc/pki/tls/certs/cert.pem for cert
                         /etc/pki/tls/private/key.pem for key
    Note: the paths are adjusted to the certificate system.

- Variables renamed from starting with rsyslog_ to logging_
  - logging_pki_authmode - Specify the default network driver authentication mode.
    `x509/name` or `anon` are available: Default to "x509/name".
  - logging_domain - The default DNS domain used to accept remote incoming logs from
    remote hosts. Default to {{ ansible_domain if ansible_domain else ansible_hostname }}
  - logging_permitted_peers - List of hostnames, IP addresses or wildcard DNS domains
    which will be allowed by the `logging` server to connect and send logs over TLS.
    Default to ['*.{{ logging_domain }}']
  - logging_send_permitted_peers - List of hostnames, IP addresses or wildcard DNS
    domains which will be verified by the `logging` client and will allow to send logs
    to the remote server over TLS. Default to '{{ logging_permitted_peers }}'.

- Implementation details
  - Moving __rsyslog_conf_global_options from the rsyslog defaults main.yml to global.j2.
  - Introduced __rsyslog_pki_map to specify gnutls or ptcp.

- Test playbook
  - tests/tests_forwards_cert.yml

- Doc - adding the new and updated variables to roles/rsyslog/README.md

### DIFF
--- a/roles/rsyslog/defaults/main.yml
+++ b/roles/rsyslog/defaults/main.yml
@@ -58,7 +58,7 @@ __rsyslog_default_netstream_driver: '{{ __rsyslog_pki_map[logging_pki] | d("ptcp
 rsyslog_send_over_tls: |-
   $ActionSendStreamDriver {{ __rsyslog_default_netstream_driver }}
   $ActionSendStreamDriverAuthMode {{ logging_pki_authmode }}
-  {% if logging_pki_authmode != "anon" %}
+  {% if logging_pki_authmode == "x509/name" %}
   {%   if logging_send_permitted_peers is string %}
   $ActionSendStreamDriverPermittedPeer {{ logging_send_permitted_peers }}
   {%   else %}
--- a/roles/rsyslog/defaults/main.yml
+++ b/roles/rsyslog/defaults/main.yml
@@ -46,91 +46,28 @@ rsyslog_custom_config_files: []
 # By setting false, it'd change 2020-03-27T14:16:47.139796+00:00)
 rsyslog_basics_use_traditional_timestamp_format: true
 
-# Encrypted communication
+# communication over tls
 # ------------------------
 
-# X.509 certificates related variables
-
-# rsyslog_pki_path
-#
-# Path to the directory with PKI realms.
-rsyslog_pki_path: "/etc/pki"
-
-# rsyslog_pki_realm
-#
-# Name of the PKi realm to use with ``rsyslogd``.
-rsyslog_pki_realm: "domain"
-
-# rsyslog_pki_ca
-#
-# Name of the root CA certificate used by the ``rsyslog`` role.
-rsyslog_pki_ca: "CA.crt"
-
-# rsyslog_pki_crt
-#
-# Name of the client certificate file used by the ``rsyslog`` role.
-rsyslog_pki_crt: "default.crt"
-
-# rsyslog_pki_key
-#
-# Name of the private key file used by the ``rsyslog`` role.
-rsyslog_pki_key: "default.key"
-
-# Lookup rsyslog network stream driver
-__rsyslog_encryption_map:
-  none: ptcp
-  ptcp: ptcp
-  tls: gtls
-  gtls: gtls
-  gnutls: gtls
-  openssl: openssl
-
-# rsyslog_default_netstream_driver
-#
-# Specify the default netstream driver used by the ``imtcp`` module.
-rsyslog_default_netstream_driver: '{{ __rsyslog_encryption_map[logging_encryption] | d("ptcp") }}'
-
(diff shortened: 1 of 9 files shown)


richm:
lgtm, but it looks like you'll need to rebase based on recent PRs that are either proposed or have merged

nhosoi:
> lgtm, but it looks like you'll need to rebase based on recent PRs that are either proposed or have merged

Thank you, @richm. I think I fixed the conflicts.
```
$ git rebase master
Successfully rebased and updated refs/heads/cert_path.
```
@pcahyna, is it ok to merge or I'd better wait for more reviews?

roles/rsyslog/defaults/main.yml
```diff
@@ -46,92 +46,29 @@ rsyslog_custom_config_files: []
...
 # enabled. It's used to configure TLS options.
 rsyslog_send_over_tls: |-
-  $ActionSendStreamDriver {{ rsyslog_default_netstream_driver }}
-  $ActionSendStreamDriverAuthMode {{ rsyslog_default_driver_authmode }}
-  {% if rsyslog_default_driver_authmode != "anon" %}
-  {%   if rsyslog_send_permitted_peers is string %}
-  $ActionSendStreamDriverPermittedPeer {{ rsyslog_send_permitted_peers }}
+  $ActionSendStreamDriver {{ __rsyslog_default_netstream_driver }}
+  $ActionSendStreamDriverAuthMode {{ logging_pki_authmode }}
+  {% if logging_pki_authmode != "anon" %}
```
richm, line 61: I think this would be better to check explicitly for auth modes that use permitted peers e.g.
```
{% if logging_pki_authmode == "x509/name" %}
```
that way we won't be surprised if we add another authmode in the future that doesn't support permitted peers.

roles/rsyslog/vars/main.yml
```diff
@@ -51,3 +76,12 @@ rsyslog_weight_map:
   'rulesets': '50'
   'input': '90'
   'inputs': '90'
+
+# X.509 certificates related variables
+# Lookup rsyslog netwrok stream driver
```
richm, line 81: network

roles/rsyslog/vars/main.yml
```diff
@@ -51,3 +76,12 @@ rsyslog_weight_map:
...
+__rsyslog_pki_map:
+  none: ptcp
+  ptcp: ptcp
+  tls: gtls
+  gtls: gtls
+  gnutls: gtls
```
richm, line 87: Was there an `openssl` in this list, or was that another list?

nhosoi: First I put openssl, but I noticed we don't enable openssl in building rsyslog package for rhel's. So, I removed it (but I forgot to clean up all of them... Sorry, I'm going to do it. ):
```
# cat /etc/redhat-release 
CentOS Linux release 8.1.1911 (Core) 
# yum search rsyslog | egrep ssl
# yum search rsyslog | egrep gnu
rsyslog-gnutls.x86_64 : TLS protocol support for rsyslog
```

richm: > First I put openssl, but I noticed we don't enable openssl in building rsyslog package for rhel's

What about Fedora?  Is there openssl support there?

nhosoi: > > First I put openssl, but I noticed we don't enable openssl in building rsyslog package for rhel's
> 
> What about Fedora? Is there openssl support there?

No, it's not supported in Fedora 31 and 32...

richm:
lgtm

nhosoi:
> lgtm

Thank you, @richm.
@pcahyna, is it ok to merge this PR?

nhosoi:
I'm refactoring the doc and I'd like to include this change in it. So, let me merge this PR for now. If there are some issues found later, let me work in another pr.

PR #127, opened 2020-05-30
Support imuxsock in the basics input.

nhosoi:
RHEL-7 systemd contains listen.conf in the package to be placed in /etc/rsyslog.d,
which expects the imuxsock configuration in the old style. (See [1] for the details.)
This PR let RHEL-7 use the old rsyslog syntax style to handle imuxsock and the rest
platforms do the new style.

[1] - https://github.com/linux-system-roles/logging/pull/124#issuecomment-636193575

CI-tests
- adding tests_imuxsock_files.yml
- checking whether "a RainerScript command" string is in the systemctl status output since
  when the conflict caused by listen.conf occurs, a message "command 'SystemLogSocketName'
  is currently not permitted - did you already set it via a RainerScript command" is issued
  by rsyslogd.


roles/rsyslog/tasks/main.yml
```diff
@@ -1,3 +1,11 @@
+- name: Set version specific variables
+  include_vars: "{{ item }}"
+  with_first_found:
+    - "{{ ansible_distribution }}-{{ ansible_distribution_version }}.yml"
+    - "{{ ansible_distribution }}-{{ ansible_distribution_major_version }}.yml"
+    - "{{ ansible_distribution }}.yml"
+    - "{{ ansible_os_family }}.yml"
```
richm, line 7: you should have a `default.yml` here to catch the case where the `ansible_os_family` and `ansible_distribution` is unknown.

roles/rsyslog/vars/RedHat.yml
```diff
@@ -0,0 +1 @@
+__rsyslog_input_basics: input_basics.j2
```
richm, line 1: this should be in `roles/rsyslog/vars/default.yml`

nhosoi: Ok, I'm going to add default.yml. Had we better remove RedHat.yml (and its symlink CentOS.yml) for now?

richm: > Ok, I'm going to add default.yml. Had we better remove RedHat.yml (and its symlink CentOS.yml) for now?

Yes, please

nhosoi: Thank you, @richm. Pushed the fix.

nhosoi:
Thank you for your reviews and approval, @richm. Merging.

PR #128, opened 2020-06-01
Fixing an ansible-lint warning "wrong indentation: expected 8 but fou…

nhosoi:
…nd 6  (indentation)"

### DIFF
--- a/tests/tests_files_elasticsearch_certs_incomplete.yml
+++ b/tests/tests_files_elasticsearch_certs_incomplete.yml
@@ -22,34 +22,34 @@
         - "{{ __test_key }}"
 
     - block:
-      - name: deploy config to send to elasticsearch
-        vars:
-          logging_outputs:
-            - name: elasticsearch_output
-              type: elasticsearch
-              server_host: logging-es
-              server_port: 9200
-              index_prefix: project.
-              input_type: ovirt
-              retryfailures: off
-              ca_cert: /etc/rsyslog.d/ca_cert.crt
-              key: /etc/rsyslog.d/key.pem
-              cert_src: "{{ __test_cert }}"
-              key_src: "{{ __test_key }}"
-          logging_inputs:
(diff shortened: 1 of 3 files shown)


tests/tests_files_elasticsearch_certs_incomplete.yml
```diff
@@ -21,8 +21,8 @@
         - "{{ __test_cert }}"
         - "{{ __test_key }}"
 
-    - block:
-      - name: deploy config to send to elasticsearch
+    - name: deploy config to send to elasticsearch
+      block:
```
pcahyna, line 25: This does not look correct. You essentially deleted the list item with the `include_role` below.

tests/tests_files_elasticsearch_certs_incomplete.yml
```diff
@@ -47,7 +47,6 @@
         include_role:
           name: linux-system-roles.logging
 
-      - name: unreachable task
```
pcahyna, line 50: Why are you deleting this? There has to be a list of tasks under the block and you are removing the `- ` list item marker.

nhosoi: > Why are you deleting this? There has to be a list of tasks under the block and you are removing the `- ` list item marker.

Yeah, obviously this patch is broken :( Sorry for pushing it too early.

nhosoi: > > Why are you deleting this? There has to be a list of tasks under the block and you are removing the `- ` list item marker.
> 
> Yeah, obviously this patch is broken :( Sorry for pushing it too early.

@pcahyna, I've updated the patch. All I had to do was adding 2 column indentation... :p

nhosoi:
Thank you for the reviews and approval, @pcahyna and @richm. Merging.

PR #130, opened 2020-06-03
Unifying the inner loop var name, e.g., basics_item, files_item, etc,…

nhosoi:
… to inner_item.

Note: if we use the default name item, the following WARNING is issued.
The loop variable 'item' is already in use. You should set the `loop_var` value in the
`loop_control` option for the task to something else to avoid variable collisions and
 unexpected behavior."

### DIFF
--- a/roles/rsyslog/tasks/deploy.yml
+++ b/roles/rsyslog/tasks/deploy.yml
@@ -11,29 +11,33 @@
   template:
     src: 'rules.conf.j2'
     dest: >-
-      {{ item.path | d(rsyslog_config_dir) }}/{{ item.filename | d((item.weight
-      if item.weight | d() else rsyslog_weight_map[item.type | d("rules")]) +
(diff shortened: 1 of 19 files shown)


richm:
> The loop variable 'item' is already in use. You should set the loop_var value in the
loop_control option for the task to something else to avoid variable collisions and
unexpected behavior.

I don't understand - where is `item` used?

nhosoi:
> > The loop variable 'item' is already in use. You should set the loop_var value in the
> > loop_control option for the task to something else to avoid variable collisions and
> > unexpected behavior.
> 
> I don't understand - where is `item` used?

It might be clearer in the second commit https://github.com/linux-system-roles/logging/pull/130/commits/6f49ef5c38315974146faced8379ae119fde47a3.
These tasks are actually inner loops in the original commit's loop.
https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/tasks/deploy.yml#L10-L39

nhosoi:
Thank you, @richm. Squashed.

PR #131, opened 2020-06-03
Renaming option key names to more generic ones.

nhosoi:
- files input: rsyslog_input_log_path -> input_log_path
- basic input: rsyslog_imjournal_... -> journal_...
- rsyslog_config_dir -> logging_config_dir

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,9 +242,9 @@ Variables in vars.yml
       - `ca_cert`: Path to CA cert for Elasticsearch.  Default to '/etc/rsyslog.d/es-ca.crt'
       - `cert`: Path to cert for Elasticsearch.  Default to '/etc/rsyslog.d/es-cert.pem'
       - `key`: Path to key for Elasticsearch.  Default to "/etc/rsyslog.d/es-key.pem"
-      - `ca_cert_src`: Path to the CA cert file on the local host to copy to the target host. If `ca_cert` is specified, copied to the location. Otherwise, to rsyslog_config_dir.
-      - `cert_src`: Path to the cert file on the local host to copy to the target host. If `cert` is specified, copied to the location. Otherwise, to rsyslog_config_dir.
-      - `key_src`: Path to the key file on the local host to copy to the target host. If `key` is specified, copied to the location. Otherwise, to rsyslog_config_dir.
+      - `ca_cert_src`: Path to the CA cert file on the local host to copy to the target host. If `ca_cert` is specified, copied to the location. Otherwise, to logging_config_dir.
+      - `cert_src`: Path to the cert file on the local host to copy to the target host. If `cert` is specified, copied to the location. Otherwise, to logging_config_dir.
+      - `key_src`: Path to the key file on the local host to copy to the target host. If `key` is specified, copied to the location. Otherwise, to logging_config_dir.
    -  ** `type: files`**
       - `facility`: facility; default to `*`
       - `severity`: severity; default to `*`
@@ -278,11 +278,11 @@ Variables in vars.yml
    -  ** `type: basics`**
       - `kernel_message`: load `imklog`. Default to `false`.
       - `use_imuxsock`: use `imuxsock` instead of `imjournal`. Default to `false`.
-      - `rsyslog_imjournal_ratelimit_burst`: set to imjournal RateLimit.Burst. Default to 20000.
-      - `rsyslog_imjournal_ratelimit_interval`: set to imjournal RateLimit.Interval. Default to 600.
-      - `rsyslog_imjournal_persist_state_interval`: set to imjournal PersistStateInterval. Default to 10.
+      - `journal_ratelimit_burst`: set to imjournal RateLimit.Burst. Default to 20000.
+      - `journal_ratelimit_interval`: set to imjournal RateLimit.Interval. Default to 600.
+      - `journal_persist_state_interval`: set to imjournal PersistStateInterval. Default to 10.
    -  ** `type: files`**
-      - `rsyslog_input_log_path`: File name to be read by the imfile plugin. The value should be full path. Wildcard '*' is allowed in the path.  Default to `/var/log/containers/*.log`.
+      - `input_log_path`: File name to be read by the imfile plugin. The value should be full path. Wildcard '*' is allowed in the path.  Default to `/var/log/containers/*.log`.
    -  ** `type: remote_files`**
       - `udp_port`: if the port number is given, rsyslog is configured to listen at the udp port number. Default to 514.
       - `tcp_port`: if the port number is given, rsyslog is configured to listen at the tcp port number. Default to 514.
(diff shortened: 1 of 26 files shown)


pcahyna:
Only indirectly related remark, but is there any reason for having `logging_config_dir`/`rsyslog_config_dir` configurable? Does one ever need to change it? Other roles usually have the path to their configuration files hardoded, because that's where the software they manage expects to find it.

nhosoi:
> Only indirectly related remark, but is there any reason for having `logging_config_dir`/`rsyslog_config_dir` configurable? Does one ever need to change it? Other roles usually have the path to their configuration files hardoded, because that's where the software they manage expects to find it.

A good point. Rsyslog main config file /etc/rsyslog.conf path is hardcoded. Sub-config files included from rsyslog.conf can be theoretically placed in any directory instead of /etc/rsyslog.d. But considering the SELinux policy, we may not want to allow such extra flexibility. Shall I remove logging_config_dir?

pcahyna:
My reasoning was that we should not provide too many knobs that virtually nobody would actually use. I have not even considered selinux, but it is a good point, which further supports removing this option (if tbere was no request for it). Since you are already renaming the variable, please prefix it with double underscore in this PR and move from defaults to vars.

nhosoi:
> My reasoning was that we should not provide too many knobs that virtually nobody would actually use. I have not even considered selinux, but it is a good point, which further supports removing this option (if there was no request for it). Since you are already renaming the variable, please prefix it with double underscore in this PR and move from defaults to vars.

Done. Sorry, I guess the second commit https://github.com/linux-system-roles/logging/pull/131/commits/87ddf8247f8918740f913b270b87976c75379588 should be treated separately from the original one. If both commits look ok, I won't squash them, but merge them as they are as 2 separate commits.

README.md
```diff
@@ -278,11 +278,11 @@ Variables in vars.yml
...
       - `use_imuxsock`: use `imuxsock` instead of `imjournal`. Default to `false`.
-      - `rsyslog_imjournal_ratelimit_burst`: set to imjournal RateLimit.Burst. Default to 20000.
-      - `rsyslog_imjournal_ratelimit_interval`: set to imjournal RateLimit.Interval. Default to 600.
-      - `rsyslog_imjournal_persist_state_interval`: set to imjournal PersistStateInterval. Default to 10.
+      - `journal_ratelimit_burst`: set to imjournal RateLimit.Burst. Default to 20000.
+      - `journal_ratelimit_interval`: set to imjournal RateLimit.Interval. Default to 600.
+      - `journal_persist_state_interval`: set to imjournal PersistStateInterval. Default to 10.
    -  ** `type: files`**
-      - `rsyslog_input_log_path`: File name to be read by the imfile plugin. The value should be full path. Wildcard '*' is allowed in the path.  Default to `/var/log/containers/*.log`.
+      - `input_log_path`: File name to be read by the imfile plugin. The value should be full path. Wildcard '*' is allowed in the path.  Default to `/var/log/containers/*.log`.
```
richm, line 285: Is the default still `/var/log/containers/*.log`?

roles/rsyslog/README.md
```diff
@@ -386,20 +385,20 @@ __rsyslog_conf_yourname:
...
 
-If the deploy destination is other than {{ rsyslog_config_dir }}, the path is to be set to path.
+If the deploy destination is other than `/etc/rsyslog.d`, the path is to be set to path.
 
 By default, the generated configuration file starts with a comment "# Ansible managed".  It could break some type of configurations.  For instance, "version=2" must be the first line in a rulebase file.  To avoid having "# Ansible managed", set true to nocomment.
 
 Some full path configuration may be referred from other configuration file, e.g., 20-viaq_formatting.conf refers parse_json.rulebase as follows.
 ```
-action(type="mmnormalize" ruleBase="{{ rsyslog_config_dir }}/parse_json.rulebase" variable="$!MESSAGE")
+action(type="mmnormalize" ruleBase="{{ __rsyslog_config_dir }}/parse_json.rulebase" variable="$!MESSAGE")
```
richm, line 394: this is possibly different than `/etc/rsyslog.d`?

README.md
nhosoi: Yeah, I haven't changed it. Do you have any recommendation?

roles/rsyslog/README.md
nhosoi: Hmmm, do we want to introduce an extra var like `logging_aux_dir` (or `logging_rulebase_dir` more specifically)? I think we don't want to allow users change `rsyslog_config_dir` (based on the discussion https://github.com/linux-system-roles/logging/pull/131#issuecomment-638100420)?

pcahyna: sorry, I don't know what rule base dir is, but does anybody need to set it to anything else than `/etc/rsyslog.d`? @richm is your comment implying that this may be needed?

pcahyna: If not, you should replace the template expansion by its value, since this is a documentation file file and who wants to see templates referring to internal variables in user documentation?

README.md
richm: > Yeah, I haven't changed it. Do you have any recommendation?

I don't think there is a default.  I think this is a required parameter?  It doesn't make sense for the user to specify a file input with no file to read from . . .

roles/rsyslog/README.md
richm: I thought that one of the purposes of this PR was to address this issue:

> Only indirectly related remark, but is there any reason for having logging_config_dir/rsyslog_config_dir configurable? Does one ever need to change it? Other roles usually have the path to their configuration files hardoded, because that's where the software they manage expects to find it.

I was confused because some of the changes change `rsyslog_config_dir` to `/etc/rsyslog.d`, and some keep `rsyslog_config_dir` or use `__rsyslog_config_dir` instead.

jvymazal:
I would just add that from practical point of view, while config files could be included from anywhere and be valid for rsyslog, I 1) do not remember anybody using some other location (from sosreports, customer cases etc.) and 2) see no reason why we need such support for the system role. As I mentioned in other places someone doing really advanced config will probably not want to use the role anyway, so from my perspective it is okay to have minor limitations like this..

roles/rsyslog/README.md
nhosoi: Ah, sorry. I think I should have used __rsyslog_config_dir in all the places. Taking a look and fixing them.

nhosoi:
> I would just add that from practical point of view, while config files could be included from anywhere and be valid for rsyslog, I 1) do not remember anybody using some other location (from sosreports, customer cases etc.) and 2) see no reason why we need such support for the system role. As I mentioned in other places someone doing really advanced config will probably not want to use the role anyway, so from my perspective it is okay to have minor limitations like this..

Thank you for your comments, @jvymazal. I assume your experience includes the location of rulebase files. If it is true, I'm keeping the rsyslog config path internal, i.e., users are not allowed to change it.

README.md
nhosoi: > > Yeah, I haven't changed it. Do you have any recommendation?
> 
> I don't think there is a default. I think this is a required parameter? It doesn't make sense for the user to specify a file input with no file to read from . . .

Done in https://github.com/linux-system-roles/logging/pull/131/commits/9b8ebe69a1fcc24274cd1138bb44e5bde5e5c40e. (I think this commit also should not be squashed, but be a separate commit.)

richm:
lgtm

pcahyna:
> see no reason why we need such support for the system role

Exactly, system roles should use opinionated good defaults and do not need to support changing them, if there is no actual need for this.

nhosoi:
Thank you for your reviews and comments, @richm, @pcahyna, and @jvymazal.
Merging.

PR #134, opened 2020-06-10
Updating the LICENSE text

nhosoi:
Replacing Apache license with GPL v3; Adding GPL v3 License file COPYING.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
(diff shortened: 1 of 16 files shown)


richm:
You should have a file called `COPYING` in the root directory which has the full text of the GPLv3 - https://www.gnu.org/licenses/gpl-3.0.txt

nhosoi:
> You should have a file called `COPYING` in the root directory which has the full text of the GPLv3 - https://www.gnu.org/licenses/gpl-3.0.txt

Yes. This pr contains `COPYING` and I verified the file is identical to gpl-3.0.txt.
https://github.com/linux-system-roles/logging/pull/134/files#diff-7116ef0705885343c9e1b2171a06be0e

nhosoi:
Thanks for the reviews, @richm!

PR #136, opened 2020-06-10
Updating meta file to adjust the galaxy_tags and the platform versions.

nhosoi:


nhosoi:
Thanks for your review, @richm.

PR #139, opened 2020-06-14
Avoiding 'on' and 'off' for the boolean values.

nhosoi:
Rsyslog parameter retryfailures (elasticsearch output) async_writing (remote_files output)
expect on | off. The logging role inherited on | off for the allowed values for the params.
But on | off are not the valid bool in ansible. Thus, replacing the pair with true | false
in docs and tests.


nhosoi:
Thank you for your review, @richm.

PR #143, opened 2020-06-16
RHELPLAN-46757 - bug with current implementation of platform/version …

nhosoi:
…specific vars/tasks file include/import

Adding 'paths: [ "{{ role_path }}/vars" ]' to the version specific var files.


nhosoi:
Thank you, @richm. Squashed.

PR #145, opened 2020-06-18
Fixing performance issues.

nhosoi:
- Avoid calling the package module with empty list.
- Eliminating unnecessary loop in the input template.

These changes makes the duration of the typical test (tests_basics_files2) shorter by ~25%.
before
```
real	5m0.877s
user	0m58.480s
sys	0m10.091s
```
after
```
real	3m42.993s
user	0m46.090s
sys	0m8.019s
```


nhosoi:
Thank you for your review, @richm. Merging.

PR #149, opened 2020-07-05
Fixing combination test playbooks.

nhosoi:
- Instead of using copy, use shell to create a test input log file for imfile.
  It is needed to have the correct selinux type when the input log file is put
  in the directory inotify is watching.
- Making the check code for the processed log more strict.


nhosoi:
[citest bad]

nhosoi:
[citest bad]

nhosoi:
[citest bad]

PR #150, opened 2020-07-07
2 Bug fixes in the basics input

nhosoi:
- There should not be multiple basics input. The basics input code is implemented to
  skip the second and the rest, if any. There was a mistake in the skipping code.
  Instead of using ansible_loop.first, change to sort the loop variables by the type
  and check ansible_loop.previtem.
- There were wrong loop variables in templates/input_basics.j2. Replacing "input"
  with "item".

Adding the test case to tests_combination.yml.


nhosoi:
[citest bad]

nhosoi:
Thank you for your review, @richm. Merging.

PR #155, opened 2020-07-22
Bug fix: ca cert w/o cert/key is not deployed

nhosoi:
When only ca cert is to be deployed, the condition to copy the ca cert was not satisfied by a typo.


nhosoi:
@richm, @pcahyna, is it ok to merge?

PR #156, opened 2020-07-22
Bug fix: items in logging_flows.input needs to be checked.

nhosoi:
If logging_flows.input contains an input which matches none of the logging_input names,
the logging_flows.input is silently ignored. It could cause a misconfiguration.

To prevent it, adding a check if logging_flows.inputs includes an item which is not part
of logging_inputs.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -37,6 +37,21 @@
         - item.type | d('') == 'custom'
         - logging_enabled | d(true)
 
+    - name: Prepare checking logging_inputs item in logging_flows.inputs
+      set_fact:
+        __logging_input_names: "{{ __logging_input_names }} + [ '{{ item.name }}' ]"
+      loop: "{{ logging_inputs | flatten(1) }}"
+      when:
+        - logging_enabled | d(true)
+
+    - name: Check logging_inputs item in logging_flows.inputs
+      fail:
+        msg: "Error: {{ item.inputs }} includes undefined logging_inputs item."
+      loop: "{{ logging_flows | flatten(1) }}"
+      when:
+        - logging_enabled | d(true)
+        - item.inputs | intersect(__logging_input_names) | length != item.inputs | length
+
     - block:
         - name: Re-read facts after adding custom fact
           setup:
@@ -77,6 +92,8 @@
       when:
         - logging_debug | d(false)
 
+  vars:
+    __logging_input_names: []
   when: logging_provider == 'rsyslog'
 
 - name: Include Rsyslog role
--- a/tests/tests_basics_files2_missing_flows.yml
+++ b/tests/tests_basics_files2_missing_flows.yml
@@ -1,0 +1,74 @@
+# Test the configuration, basics input and 4 configured files output
+#
+# [Configuration]
+# basics input (imjournal) -> 4 files output (omfile)
+# logging_purge_confs: true
+#
+# [Test scenario]
+# 0. Run logging role.
+# 1. Check rsyslog.conf file size.
+#    If logging role is executed, the file size is about 100 bytes.
+#    Thus, assert the size is less than 1000.
+# 2. Check file count in /etc/rsyslog.d.
+#    If logging role is executed, 8 config files are generated.
+#    By setting logging_purge_confs, pre-existing config files are deleted.
+#    Thus, assert the the count is equal to 8.
+# 3. Check systemctl status of rsyslog as well as error or specific message in the output.
+# 4. To verify the generated filename is correct, check the config file of files output exists.
+# 4.1 Check the config file contains the expected filter and the output file as configured.
+# 4.2 Run logger command and check the log is in the local output file {{ __default_system_log }}.
+#
+- name: Ensure that the role runs with parameters from imjournal to four omfile outputs
+  hosts: all
+  become: true
+  vars:
+    __test_files_conf: /etc/rsyslog.d/30-output-files-files_output1.conf
+    __default_system_log: /var/log/messages
+    __expected_error: "Error: ['bogus_basic_input'] includes undefined logging_inputs item."
+
+  tasks:
+    - block:
+        - name: deploy config to output into local files
+          vars:
+            logging_purge_confs: true
+            logging_outputs:
+              - name: files_output0
+                type: files
+                facility: authpriv,auth
+                path: /var/log/secure
+              - name: files_output1
+                type: files
+                severity: info
+                exclude:
+                  - authpriv.none
+                  - auth.none
+                  - cron.none
+                  - mail.none
+                path: "{{ __default_system_log }}"
+              - name: files_output2
+                type: files
+                severity: emerg
+                path: :omusrmsg:*
+              - name: files_output3
+                type: files
+                facility: local7
+                path: /var/log/boot.log
+            logging_inputs:
+              - name: basic_input
+                type: basics
+            logging_flows:
+              - name: flow_0
+                inputs: [bogus_basic_input]
+                outputs: [files_output0, files_output1, files_output2, files_output3]
+          include_role:
+            name: linux-system-roles.logging
+
+        - name: unreachable task
+          fail:
+            msg: UNREACH
+
+      rescue:
+        - debug:
+            msg: Caught an expected error - {{ ansible_failed_result }}
+        - assert:
+            that: ansible_failed_result.results.0.msg == __expected_error


tasks/main.yml
```diff
@@ -37,6 +37,21 @@
         - item.type | d('') == 'custom'
         - logging_enabled | d(true)
 
+    - name: Prepare checking logging_inputs item in logging_flows.inputs
+      set_fact:
```
pcahyna, line 41: Why are you setting a fact, which will persist in the namespace forever? What if the role is called multiple times? Please transform it into a local variable (on a block if necessary).

nhosoi: Note: If logging_flows contain a non-existing input name, ansible-playbook fails with this error.
`Error: ['bogus_basic_input'] includes undefined logging_inputs item.`

I'm trying to add a test playbook, but I cannot make this assertion pass, although the strings on the both sides `\"Error: ['bogus_basic_input'] includes undefined logging_inputs item.\"` look identical to me. @pcahyna, do you have any ideas why this assertion [0] fails?
```
TASK [assert...] ***************************************************************************************************
task path: /.../linux-system-roles/logging/tests/tests_basics_files2_missing_flows.yml:73
fatal: [/.../CentOS-Stream-GenericCloud-8-20200113.0.x86_64.qcow2]: FAILED! => {
    "assertion": "\"Error: ['bogus_basic_input'] includes undefined logging_inputs item.\" is match(\"Error: ['bogus_basic_input'] includes undefined logging_inputs item.\")",
    "changed": false,TASK [assert...] ***************************************************************************************************
task path: /.../linux-system-roles/logging/tests/tests_basics_files2_missing_flows.yml:73
fatal: [/.../CentOS-Stream-GenericCloud-8-20200113.0.x86_64.qcow2]: FAILED! => {
    "assertion": "\"Error: ['bogus_basic_input'] includes undefined logging_inputs item.\" is match(\"Error: [.*] includes undefined logging_inputs item.\")",
    "changed": false,
    "evaluated_to": false
}
```
[0] This is the assertion in the rescue.
```
      rescue:
        - debug:
            msg: "Caught an expected error - {{ ansible_failed_result.results }}"
        - name: assert...
          assert:
            that: '"{{ ansible_failed_result.results.0.msg }}" is match("{{ __expected_error }}")'
```
This is the __expected_error.
    `__expected_error: "Error: ['bogus_basic_input'] includes undefined logging_inputs item."`

richm: I tried a simple ansible playbook:
```yaml
---
- name: test
  hosts: localhost
  tasks:
    - block:
        - set_fact:
            __expected_error: "Error: ['bogus_basic_input'] includes undefined logging_inputs item."
            __results: "Error: ['bogus_basic_input'] includes undefined logging_inputs item."
        - fail:
            msg: "{{ __results }}"
      rescue:
        - debug:
            msg: Caught an expected error - {{ ansible_failed_result }}
        - assert:
            that: ansible_failed_result.msg == __expected_error
```
this is the output
```yaml
PLAY [test] ********************************************************************

TASK [Gathering Facts] *********************************************************
task path: /home/rmeggins/ansible_sandbox/block-rescue-test.yml:2
ok: [localhost]
META: ran handlers

TASK [set_fact] ****************************************************************
task path: /home/rmeggins/ansible_sandbox/block-rescue-test.yml:6
ok: [localhost] => {
    "ansible_facts": {
        "__expected_error": "Error: ['bogus_basic_input'] includes undefined logging_inputs item.",
        "__results": "Error: ['bogus_basic_input'] includes undefined logging_inputs item."
    },
    "changed": false
}

TASK [fail] ********************************************************************
task path: /home/rmeggins/ansible_sandbox/block-rescue-test.yml:9
fatal: [localhost]: FAILED! => {
    "changed": false
}

MSG:

Error: ['bogus_basic_input'] includes undefined logging_inputs item.

task path: /home/rmeggins/ansible_sandbox/block-rescue-test.yml:12
ok: [localhost] => {}

MSG:

Caught an expected error - {'failed': True, 'msg': "Error: ['bogus_basic_input'] includes undefined logging_inputs item.", '_ansible_no_log': False, 'changed': False}

TASK [assert] ******************************************************************
task path: /home/rmeggins/ansible_sandbox/block-rescue-test.yml:14
ok: [localhost] => {
    "changed": false
}

MSG:

All assertions passed
META: ran handlers
META: ran handlers

localhost                  : ok=4    changed=0    unreachable=0    failed=0    skipped=0    rescued=1    ignored=0   
```
couple of notes:
- afaict the failure message is in `ansible_failed_result.msg` not `ansible_failed_result.results.0.msg`
- if `__expected_error` is the exact string you can use `==` - only use `is match(thing)` if `thing` is a regexp
- no need for the single or double quoting in the `that`

nhosoi: Thank, @richm!
I had to use `ansible_failed_result.results.0.msg` since the value of `ansible_failed_result.msg` is `All items completed` in this case as follows. But `==` made the test work!
```
Caught an expected error - {'results': [{'failed': True, 'msg': "Error: ['bogus_basic_input'] includes undefined 
logging_inputs item.", '_ansible_no_log': False, 'changed': False, 'item': {'name': 'flow_0', 'inputs': 
['bogus_basic_input'], 'outputs': ['files_output0', 'files_output1', 'files_output2', 'files_output3']}, 
'ansible_loop_var': 'item', '_ansible_item_result': True, '_ansible_item_label': {'name': 'flow_0', 'inputs': 
['bogus_basic_input'], 'outputs': ['files_output0', 'files_output1', 'files_output2', 'files_output3']}}], 'failed': 
True, 'msg': 'All items completed', 'changed': False}.
```
To be honest, I still think `is match` should work, as well. But I'd better move on. :)

nhosoi:
@richm, @pcahyna, is it ok to squash 3 commits into one and merge?

tasks/main.yml
pcahyna: > Why are you setting a fact, which will persist in the namespace forever? What if the role is called multiple times? Please transform it into a local variable (on a block if necessary).

So, this comment was not addressed?

nhosoi: > > Why are you setting a fact, which will persist in the namespace forever? What if the role is called multiple times? Please transform it into a local variable (on a block if necessary).
> 
> So, this comment was not addressed?

I made __logging_input_names a "local" variable in the block here;
https://github.com/linux-system-roles/logging/pull/156/files#diff-2444ad0870f91f17ca6c2a5e96b26823R95-R96
By doing it, I could get rid of initializing __logging_input_name in line 42.

I thought that was what you suggested. If it was not, could you please give me some more details? Thanks.

pcahyna: No, I suggested to entirely eliminate `set_fact`. What you did introduces a local variable, but keeps the fact, so you have both. (A fact overrides a local variable according to https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable)

pcahyna: here's an example of using your method and suggested method (if you run it, you'll see that the results are the same).
```yaml
- hosts: localhost
  vars:
    logging_inputs:
      - name: foo
      - name: bar
      - name: baz
  tasks:
    - block:
        - name: Prepare checking logging_inputs item in logging_flows.inputs
          set_fact:
            __logging_input_names: "{{ __logging_input_names }} + [ '{{ item.name }}' ]"
          loop: "{{ logging_inputs | flatten(1) }}"
        - debug:
            var: __logging_input_names
        - debug:
            msg: "intersection = {{ __logging_input_names |  intersect(['baz', 'foo'])}}"        
      vars:
        __logging_input_names: []
  
    - block:
        - debug:
            var: __logging_input_names_bis
        - debug:
            msg: "intersection = {{ __logging_input_names_bis |  intersect(['baz', 'foo'])}}"
      vars:
        __logging_input_names_bis:  "{{ logging_inputs | flatten(1) | map(attribute='name')|list}}"
```
Also, what is `flatten(1)` doing there, are those lists expected to be nested? The README does say that `inputs`, `flows` are lists of dicts, not lists of lists.

nhosoi: Thank you sooooo much, @pcahyna. I've opened [pr/160](https://github.com/linux-system-roles/logging/pull/160) as suggested (hopefully, this time...)

pcahyna: thanks, and what about the `flatten()`s ?

nhosoi: Ah, sorry... You asked me if we could get rid of flatten from here by this question?
> doing there, are those lists expected to be nested?

Yes, as long as we could trust the users' inputs, we can remove it. 
Actually, we have 4 others in roles/rsyslog/tasks/main.yml. Do we want to remove them, as well?
https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/tasks/main.yml#L243
https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/tasks/main.yml#L255
https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/tasks/main.yml#L267
https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/tasks/main.yml#L279

pcahyna: Yes, please remove all of those in #160.

PR #161, opened 2020-08-03
Imjournal configuration for RHEL-7 was missing the following directives

nhosoi:
# File to store the position in the journal
 $IMJournalStateFile imjournal.state
 $imjournalRatelimitBurst "{{ item.ratelimit_burst | d(20000) }}"
 $imjournalRatelimitInterval "{{ item.ratelimit_interval | d(600) }}"
 $imjournalPersistStateInterval "{{ item.journal_persist_state_interval | d(10) }}")

### DIFF
--- a/roles/rsyslog/templates/input_basics_rhel7.j2
+++ b/roles/rsyslog/templates/input_basics_rhel7.j2
@@ -7,8 +7,15 @@ $ModLoad imuxsock # provides support for local system logging (e.g. via logger c
 $OmitLocalLogging off
 {% else %}
 $ModLoad imjournal # provides access to the systemd journal
+# File to store the position in the journal
+$IMJournalStateFile imjournal.state
+$imjournalRatelimitBurst {{ item.ratelimit_burst | d(20000) }}
+$imjournalRatelimitInterval {{ item.ratelimit_interval | d(600) }}
+$imjournalPersistStateInterval {{ item.journal_persist_state_interval | d(10) }}
+
 # Turn off message reception via local log socket;
 # local messages are retrieved through imjournal now.
 $OmitLocalLogging on
+
 {% endif %}
 {{ lookup('template', 'input_template.j2') }}


sopos:
I think we can simply copy complete RHEL-8 template as they should be fully compatible. I think it should be compatible even with Feroda.

sopos:
I can confirm that copying `input_basics.j2` to `input_basics_rhel7.j2` produces correct config.

nhosoi:
> I think we can simply copy complete RHEL-8 template as they should be fully compatible. I think it should be compatible even with Feroda.

Unfortunately, on rhel-7, other system/application installs rsyslog config file (e.g., listen.conf) in /etc/rsyslog.d, which uses the old style. It makes rsyslog fail to start. Please see this comment.

commit b3e87e6abb63cb04b4d7d8307970b5b57c4a51b1
Date:   Sat May 30 14:27:24 2020 -0700

    Support imuxsock in the basics input.
    
    RHEL-7 systemd contains listen.conf in the package to be placed in /etc/rsyslog.d,
    which expects the imuxsock configuration in the old style. (See [1] for the details.)
    This PR let RHEL-7 use the old rsyslog syntax style to handle imuxsock and the rest
    platforms do the new style.
    
    [1] - https://github.com/linux-system-roles/logging/pull/124#issuecomment-636193575

roles/rsyslog/templates/input_basics_rhel7.j2
```diff
@@ -7,8 +7,15 @@ $ModLoad imuxsock # provides support for local system logging (e.g. via logger c
 $OmitLocalLogging off
 {% else %}
 $ModLoad imjournal # provides access to the systemd journal
+# File to store the position in the journal
+$IMJournalStateFile imjournal.state
+$imjournalRatelimitBurst "{{ item.ratelimit_burst | d(20000) }}"
+$imjournalRatelimitInterval "{{ item.ratelimit_interval | d(600) }}"
+$imjournalPersistStateInterval "{{ item.journal_persist_state_interval | d(10) }}")
```
sopos, line 14:
```suggestion
$imjournalPersistStateInterval "{{ item.journal_persist_state_interval | d(10) }}"
```

sopos, line 14:
```suggestion
$imjournalRatelimitBurst {{ item.ratelimit_burst | d(20000) }}
$imjournalRatelimitInterval {{ item.ratelimit_interval | d(600) }}
$imjournalPersistStateInterval {{ item.journal_persist_state_interval | d(10) }}
```

nhosoi: Thanks! Done.

nhosoi:
Thanks, @sopos, for the discussion on the rsyslog mailing list. I think we are able to support the use case(s).

Two questions on the example in the discussion.
1) we don't allow stream driver names other than gtls and ptcp. Do we want to add `ossl`?
2) MaxListners and MaxSessions are useful parameters to add to the server?
```
  ## TCP syslog reception:
  module (load="imtcp"
    MaxListeners = "30"
    MaxSessions = "1024"
    StreamDriver.Name="ossl"
  # StreamDriver.Name="gtls"
    StreamDriver.Mode="1"
    StreamDriver.AuthMode="x509/name"
  # StreamDriver.AuthMode="anon"
    PermittedPeer=.....
  )
```

nhosoi:
@sopos, do you think this PR is ok to merge?

sopos:
According to my testing it looks sane.

nhosoi:
Thanks, @sopos!!! Merging.

PR #163, opened 2020-08-10
Fix verbatim formatting in markdown

pcahyna:
Use backticks. Otherwise asciidoc conversion aborts on the unquoted asterisk.


nhosoi:
[citest bad]

pcahyna:
[citest bad]

PR #167, opened 2020-08-13
Changing the key-value pair 'pki: tls|none' to 'tls: true|false'

nhosoi:
Changing the key-value pair
  pki: tls|none
to
  tls: true|false
in the remote input (server) and the forwards output (client).

### DIFF
--- a/roles/rsyslog/tasks/inputs/remote/main.yml
+++ b/roles/rsyslog/tasks/inputs/remote/main.yml
@@ -4,19 +4,23 @@
     - name: Ensure Remote inputs contain no conflict connection type
       fail:
         msg: "Error: {{ item.0.name }} and {{ item.1.name }} conflict."
-      loop: "{{ [__logging_remote_udp, __logging_remote_tcp, __logging_remote_tls] }}"
+      loop: "{{ [__logging_remote_udp, __logging_remote_ptcp, __logging_remote_tlstcp] }}"
       when:
         - item | length > 1
   vars:
-    __logging_remote_udp: "{{ logging_inputs | selectattr('type', '==', 'remote') |
+    __logging_remote: "{{ logging_inputs | selectattr('type', '==', 'remote') | list }}"
+    __logging_remote_udp: "{{ __logging_remote |
                               selectattr('udp_ports', 'defined') | list }}"
-    __logging_remote_tcp: "{{ logging_inputs | selectattr('type', '==', 'remote') |
-                              selectattr('tcp_ports', 'defined') |
-                              selectattr('pki', 'undefined') | list }}"
-    __logging_remote_tls: "{{ logging_inputs | selectattr('type', '==', 'remote') |
-                              selectattr('tcp_ports', 'defined') |
-                              selectattr('pki', 'defined') |
-                              selectattr('pki', '==', 'tls') | list }}"
+    __logging_remote_tcp: "{{ __logging_remote |
+                              selectattr('tcp_ports', 'defined') | list }}"
+    __logging_remote_tls: "{{ __logging_remote_tcp |
+                              selectattr('tls', 'defined') | list }}"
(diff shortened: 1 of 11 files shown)


roles/rsyslog/tasks/inputs/remote/main.yml
```diff
@@ -4,19 +4,23 @@
...
-                              selectattr('pki', 'defined') |
-                              selectattr('pki', '==', 'tls') | list }}"
+    __logging_remote_tcp: "{{ __logging_remote |
+                              selectattr('tcp_ports', 'defined') | list }}"
+    __logging_remote_tls: "{{ __logging_remote_tcp |
+                              selectattr('tls', 'defined') | list }}"
+    __logging_remote_ptcp: "{{ __logging_remote_tcp |
+                               selectattr('tls', 'undefined') | list }} +
+                            {{ __logging_remote_tls |
+                               selectattr('tls', 'equalto', False) | list }}"
```
richm, line 21: use `selectattr('tls', 'false')` - see https://jinja.palletsprojects.com/en/2.10.x/templates/#false

roles/rsyslog/tasks/inputs/remote/main.yml
```diff
@@ -4,19 +4,23 @@
...
+    __logging_remote_tlstcp: "{{ __logging_remote_tls |
+                                 selectattr('tls', 'equalto', True) | list }}"
```
richm, line 23: use selectattr('tls', 'true') - see https://jinja.palletsprojects.com/en/2.10.x/templates/#true

roles/rsyslog/tasks/main.yml
```diff
@@ -209,12 +209,12 @@
 
       vars:
         __logging_forwards_tls: "{{ logging_outputs | selectattr('type', '==', 'forwards') |
-                                  selectattr('pki', 'defined') |
-                                  selectattr('pki', '==', 'tls') | list }}"
+                                  selectattr('tls', 'defined') |
+                                  selectattr('tls', 'equalto', True) | list }}"
```
richm, line 213: use `selectattr('tls', 'true')` - see https://jinja.palletsprojects.com/en/2.10.x/templates/#true

roles/rsyslog/tasks/main.yml
```diff
@@ -209,12 +209,12 @@
...
         __logging_remote_tls: "{{ logging_inputs | selectattr('type', '==', 'remote') |
                                   selectattr('tcp_ports', 'defined') |
-                                  selectattr('pki', 'defined') |
-                                  selectattr('pki', '==', 'tls') | list }}"
+                                  selectattr('tls', 'defined') |
+                                  selectattr('tls', 'equalto', True) | list }}"
```
richm, line 217: use `selectattr('tls', 'true')` - see https://jinja.palletsprojects.com/en/2.10.x/templates/#true

nhosoi:
Thank you, @richm! Updated the pr.

nhosoi:
Thanks @richm. Merged.

roles/rsyslog/tasks/main.yml
pcahyna: @richm "New in version 2.11" - I don't think we want to use it!

nhosoi: @pcahyna, @richm, let me replace `true` & `false` with 'equalto', True & False...

nhosoi: Opened a pr - https://github.com/linux-system-roles/logging/pull/169

richm: sorry I didn't notice this - I guess I assumed since it was on the 2.10 doc page it was supported in 2.10 :-(
but if we are worried about 2.7 support, then this problem goes beyond just `true` and `false` tests being supported

PR #170, opened 2020-08-20
Fixing the file counting code in the tests.

nhosoi:
The logging test playbooks checks the generated rsyslog config file
count first to capture the failure at the early stage. The count
could be influenced by the other applications which put its config
file in /etc/rsyslog.d. The extra count was considered, but it
was not strict enough to reduce the error.

Since the config file name generated by the logging role starts
with '[0-9][0-9]-', thus adding it to the find patterns to make
the generated file count more accurate.
Note: Most additional config file does not start with '[0-9][0-9]-',
but there is an exception 21-cloudinit.conf.

### DIFF
--- a/tests/set_rsyslog_variables.yml
+++ b/tests/set_rsyslog_variables.yml
@@ -6,5 +6,5 @@
 - name: Get file counts in /etc/rsyslog.d
   find:
     paths: /etc/rsyslog.d
-    patterns: '*.conf,*.template,*.remote,*.rulebase'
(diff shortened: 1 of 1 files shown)


nhosoi:
Sadly, it requires some more fixes... (;_;)

nhosoi:
It was my pure mistake. Only *.conf files start with `[0-9][0-9]-`, but not the others...

nhosoi:
Thank you, @richm. Merged.

PR #171, opened 2020-08-25
Test playbooks - Adding 1 second pause between the logger and its result checking.

nhosoi:
In the code to check the log message is successfully logged or not in the /var/log/messages file, adding "until: __result.rc == 0" and waiting up to 5 seconds.

### DIFF
--- a/tests/tests_basics_files.yml
+++ b/tests/tests_basics_files.yml
@@ -74,4 +74,8 @@
 
     - name: Check the test log message in {{ __default_system_log }}
       command: /bin/grep testMessage0000 {{ __default_system_log }}
+      register: __result
+      until: __result is success
+      retries: 5
+      delay: 1
       changed_when: false
--- a/tests/tests_basics_files2.yml
+++ b/tests/tests_basics_files2.yml
@@ -99,4 +99,8 @@
 
     - name: Check the test log message in {{ __default_system_log }}
       command: /bin/grep testMessage0000 "{{ __default_system_log }}"
+      register: __result
+      until: __result is success
+      retries: 5
+      delay: 1
       changed_when: false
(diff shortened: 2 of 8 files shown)


tests/tests_basics_files.yml
```diff
@@ -72,6 +72,10 @@
       command: /bin/logger -i -p local6.info -t testTag0 testMessage0000
       changed_when: false
 
+    - name: Pause 1 sec before checking "{{ __default_system_log }}"
+      pause:
+        seconds: 1
+
     - name: Check the test log message in {{ __default_system_log }}
       command: /bin/grep testMessage0000 {{ __default_system_log }}
       changed_when: false
```
richm, line 81: I think it would be better to use `until` on the Check task - https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#retrying-a-task-until-a-condition-is-met - that is - the check task would retry N times with a M second pause between each retry

nhosoi: Thanks a lot, @richm. Updated the commit.

nhosoi: Hi @xjezda00, ^^^^^

tests/tests_basics_files.yml
```diff
@@ -74,4 +74,8 @@
 
     - name: Check the test log message in {{ __default_system_log }}
       command: /bin/grep testMessage0000 {{ __default_system_log }}
+      register: __result
+      until: __result.rc == 0
```
richm, line 78: Use `until: __result is success` instead of comparing `rc` directly - see https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#task-results

nhosoi: Thanks, @richm. Done.

PR #172, opened 2020-08-31
RHELPLAN-32454 - Add imrelp support to inputs.

nhosoi:
RHELPLAN-32349 - Add omrelp support to outputs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@
     * [Standalone configuration](#standalone-configuration)
     * [Client configuration](#client-configuration)
     * [Server configuration](#server-configuration)
+    * [Client configuration with Relp](#client-configuration-with-relp)
+    * [Server configuration with Relp](#server-configuration-with-relp)
   * [Providers](#providers)
   * [Tests](#tests)
   * [Implementation Details](#implementation-details)
@@ -111,12 +113,25 @@ This is a schematic logging configuration to show log messages from input_nameA 
 - `ovirt` type - `ovirt` input supports oVirt specific inputs.<br>
    For the details, visit [oVirt Support](../../design_docs/rsyslog_ovirt_support.md).
 
-- `remote` type - `remote` input supports receiving logs from the remote logging system over the network. This input type makes rsyslog a server.<br>
+- `relp` type - `relp` input supports receiving logs from the remote logging system over the network using relp.<br>
+  **available options**
+  - `port`: Port number Relp is listening to. Default to `20514`.
+  - `tls`: If true, encrypt the connection with TLS. You must provide key/certificates and triplets {`ca_cert`, `cert`, `private_key`} and/or {`ca_cert_src`, `cert_src`, `private_key_src`}. Default to `true`.
+  - `ca_cert`: Path to CA cert to configure Relp with tls. Default to `logging_config_dir/basename of ca_cert_src`.
+  - `cert`: Path to cert to configure Relp with tls.  Default to `logging_config_dir/basename of cert_src`.
+  - `private_key`: Path to key to configure Relp with tls.  Default to `logging_config_dir/basename of private_key_src`.
+  - `ca_cert_src`: Local CA cert file path which is copied to the target host. If `ca_cert` is specified, it is copied to the location. Otherwise, to logging_config_dir.
+  - `cert_src`: Local cert file path which is copied to the target host. If `cert` is specified, it is copied to the location. Otherwise, to logging_config_dir.
+  - `private_key_src`: Local key file path which is copied to the target host. If `private_key` is specified, it is copied to the location. Otherwise, to logging_config_dir.
+  - `pki_authmode`: Specifying the authentication mode. `name` or `fingerprint` is accepted. Default to `name`.
+  - `permitted_clients`: List of hostnames, IP addresses, fingerprints(sha1), and wildcard DNS domains which will be allowed by the `logging` server to connect and send logs over TLS. Default to `['*.{{ logging_domain }}']`
+
+- `remote` type - `remote` input supports receiving logs from the remote logging system over the network.<br>
   **available options**
   - `udp_ports`: List of UDP port numbers to listen. If set, the `remote` input listens on the UDP ports. No defaults. If both `udp_ports` and `tcp_ports` are set in a `remote` input item, `udp_ports` is used and `tcp_ports` is dropped.
   - `tcp_ports`: List of TCP port numbers to listen. If set, the `remote` input listens on the TCP ports. Default to `[514]`. If both `udp_ports` and `tcp_ports` are set in a `remote` input item, `udp_ports` is used and `tcp_ports` is dropped. If both `udp_ports` and `tcp_ports` are not set in a `remote` input item, `tcp_ports: [514]` is added to the item.
   - `tls`: Set to `true` to encrypt the connection using the default TLS implementation used by the provider. Default to `false`.
-  - `pki_authmode`: Specifying the default network driver authentication mode. `x509/name`, `x509/fingerprint`, `anon` is accepted. Default to `x509/name`.
+  - `pki_authmode`: Specifying the default network driver authentication mode. `x509/name`, `x509/fingerprint`, or `anon` is accepted. Default to `x509/name`.
   - `permitted_clients`: List of hostnames, IP addresses, fingerprints(sha1), and wildcard DNS domains which will be allowed by the `logging` server to connect and send logs over TLS. Default to `['*.{{ logging_domain }}']`
 
   There are 3 types of items in the remote type - udp, plain tcp and tls tcp. The udp type configured using `udp_ports`; the plain tcp type is configured using `tcp_ports` without `tls` or with `tls: false`; the tls tcp type is configured using `tcp_ports` with `tls: true` at the same time. Please note there might be only one instance of each of the three types. E.g., if there are 2 `udp` type items, it fails to deploy.
@@ -169,10 +184,11 @@ This is a schematic logging configuration to show log messages from input_nameA 
   - `index_prefix`: Elasticsearch index prefix the particular log will be indexed to. **Required**.
   - `input_type`: Specifying the input type. Currently only type `ovirt` is supported. Default to `ovirt`.
   - `retryfailures`: Specifying whether retries or not in case of failure. Allowed value is `true` or `false`.  Default to `true`.
-  - `use_cert`: If true, key/certificates are used to access Elasticsearch. Triplets {`ca_cert`, `cert`, `private_key`} and/or {`ca_cert_src`, `cert_src`, `private_key_src`} should be configured. Default to `true`.
-  - `ca_cert`: Path to CA cert for Elasticsearch.  Default to `/etc/rsyslog.d/es-ca.crt`
-  - `cert`: Path to cert to connect to Elasticsearch.  Default to `/etc/rsyslog.d/es-cert.pem`.
-  - `private_key`: Path to key to connect to Elasticsearch.  Default to `/etc/rsyslog.d/es-key.pem`.
+  - `tls`: If true, encrypt the connection with TLS. You must provide key/certificates and triplets {`ca_cert`, `cert`, `private_key`} and/or {`ca_cert_src`, `cert_src`, `private_key_src`}. Default to `true`.
+  - `use_cert`: [DEPRECATED] If true, encrypt the connection with TLS. You must provide key/certificates and triplets {`ca_cert`, `cert`, `private_key`} and/or {`ca_cert_src`, `cert_src`, `private_key_src`}. Default to `true`. Option `use_cert` is deprecated in favor of `tls` and `use_cert` will be removed in the next minor release.
+  - `ca_cert`: Path to CA cert for Elasticsearch. Default to `logging_config_dir/basename of ca_cert_src`.
+  - `cert`: Path to cert to connect to Elasticsearch.  Default to `logging_config_dir/basename of cert_src`.
+  - `private_key`: Path to key to connect to Elasticsearch.  Default to `logging_config_dir/basename of private_key_src`.
   - `ca_cert_src`: Local CA cert file path which is copied to the target host. If `ca_cert` is specified, it is copied to the location. Otherwise, to logging_config_dir.
   - `cert_src`: Local cert file path which is copied to the target host. If `cert` is specified, it is copied to the location. Otherwise, to logging_config_dir.
   - `private_key_src`: Local key file path which is copied to the target host. If `private_key` is specified, it is copied to the location. Otherwise, to logging_config_dir.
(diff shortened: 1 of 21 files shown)


nhosoi:
@sopos, could you please take a look whenever you have time (note: this is not urgent)? Thanks!

README.md
```diff
@@ -111,12 +113,25 @@ This is a schematic logging configuration to show log messages from input_nameA
 - `ovirt` type - `ovirt` input supports oVirt specific inputs.<br>
    For the details, visit [oVirt Support](../../design_docs/rsyslog_ovirt_support.md).
 
+- `relp` type - `relp` input supports receiving logs from the remote logging system over the network using relp. This input type makes rsyslog a server.<br>
```
richm, line 116: Do we need the part about "This input type makes rsyslog a server."?

README.md
```diff
@@ -204,9 +219,24 @@ This is a schematic logging configuration to show log messages from input_nameA
   - `udp_port`: UDP port number. Default to `514`.
   - `tcp_port`: TCP port number. Default to `514`.
   - `tls`: Set to `true` to encrypt the connection using the default TLS implementation used by the provider. Default to `false`.
-  - `pki_authmode`: Specifying the default network driver authentication mode. `x509/name`, `x509/fingerprint`, `anon` is accepted. Default to `x509/name`.
+  - `pki_authmode`: Specifying the default network driver authentication mode. `x509/name`, `x509/fingerprint`, or `anon` is accepted. Default to `x509/name`.
   - `permitted_server`: Hostname, IP address, fingerprint(sha1) or wildcard DNS domain of the server which this client will be allowed to connect and send logs over TLS. Default to `*.{{ logging_domain }}`
 
+- `relp` type - `relp` output sends logs to the remote logging system over the network using relp. This output type makes rsyslog a client.<br>
```
richm, line 225: do we need this part "This output type makes rsyslog a client." ?

roles/rsyslog/templates/input_relp.j2
```diff
@@ -0,0 +1,59 @@
...
+      type="imrelp"
+      port="{{ item.port | d(20514) | int }}"
+      maxDataSize="{{ item.max_data_size | d(logging_max_message_size) | int }}"
+{% if item.use_cert | default(true) %}
+{%   if item.ca_cert is defined %}
+{%     set __cacert = item.ca_cert %}
+{%   elif item.ca_cert_src is defined %}
+{%     set __cacert = __rsyslog_config_dir + '/' + item.ca_cert_src | basename %}
+{%   else %}
+{%     set __cacert = __rsyslog_config_dir + '/es-ca.crt' %}
```
richm, line 11: input relp uses `es-ca.crt`?

nhosoi: Sorry, copy&paste error... :p (Actually, there's no chance to go there since the logging role had failed prior to this template lookup if both ca_cert and ca_cert_src are not configured.)

README.md
nhosoi: Let me remove the sentence. Remote input also has it. Shall I remove it, as well?
`129 - `remote` type - `remote` input supports receiving logs from the remote logging system o    ver the network. This input type makes rsyslog a server.<br>`

nhosoi:
BTW, I see lots of failures in the new CI tests. An obvious cause is from jinja2 namespace... Do we want to avoid using it?

richm:
> BTW, I see lots of failures in the new CI tests. An obvious cause is from jinja2 namespace... Do we want to avoid using it?

jinja 2.7 namespace, and probably the use of `selectattr` - so yes, we will need to change it, but it isn't urgent.  We should do that in a separate task which is part of a feature to "support system roles on el7" - there are likely other roles which will need to be changed to support jinja 2.7

README.md
richm: yes - I think it might be confusing to say "makes rsyslog a server" if someone can misinterpret that to ask "Does that mean it doesn't act as a client anymore?"

roles/rsyslog/tasks/outputs/relp/main.yml
```diff
@@ -0,0 +1,28 @@
...
+    __rsyslog_rules:
+      - name: "output-relp-{{ item.name }}"
+        type: "output"
+        weight: "31"
+        state: "{{ item.state | d('present') }}"
+        sections:
+          - options: "{{ lookup('template', 'output_relp.j2') }}"
+  include_tasks:
+    file: "{{ role_path }}/tasks/deploy.yml"
+  loop: '{{ rsyslog_output_relp }}'
```
richm, line 28: should be consistent with use of single quotes vs. double quotes - I think you should use double quotes except where you have to use single quotes within a double quote

tests/ca.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN CERTIFICATE-----
```
richm, line 1: This will expire Jul 29 23:09:16 2022 GMT after which point the tests will fail

tests/test-cert.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN CERTIFICATE-----
```
richm, line 1: This will expire Jul 29 23:09:16 2022 GMT after which point the tests will fail

richm:
I don't think it is a good idea to have cert/key files used for testing in the repo.  I think it is much better to generate them as part of the test.  There are several modules built-in to ansible that can be used like https://docs.ansible.com/ansible/2.8/modules/openssl_certificate_module.html#openssl-certificate-module https://docs.ansible.com/ansible/2.8/modules/openssl_privatekey_module.html#openssl-privatekey-module etc.

I suppose you would have to 
- generate a CA key
- generate a self signed CA cert using the CA key
- generate a server/client key
- generate a server/client cert using the key, and signed by the self signed CA

If this is too much trouble, then you could use the `certificate` role, but I don't like having additional role/module dependencies being required for testing - better to use what is already included with ansible

tests/tests_relp_server.yml
```diff
@@ -12,14 +12,61 @@
...
+        common_name: logging.fedora.org
+
+    - name: Generate a server/client cert using the key, and signed by the self signed CA
+      openssl_certificate:
+        path: "{{ __test_cert }}"
+        privatekey_path: "{{ __test_key }}"
+        csr_path: "{{ __test_cert_csr }}"
+        ownca_path: "{{ __test_ca_cert }}"
+        ownca_privatekey_path: "{{ __test_ca_key }}"
+        provider: ownca
```
richm, line 68: Nice.
This should be refactored into something like `tests/tasks/create_tests_certs.yml` for reuse.

README.md
```diff
@@ -111,12 +113,25 @@ This is a schematic logging configuration to show log messages from input_nameA
 - `ovirt` type - `ovirt` input supports oVirt specific inputs.<br>
    For the details, visit [oVirt Support](../../design_docs/rsyslog_ovirt_support.md).
 
-- `remote` type - `remote` input supports receiving logs from the remote logging system over the network. This input type makes rsyslog a server.<br>
+- `relp` type - `relp` input supports receiving logs from the remote logging system over the network using relp.<br>
+  **available options**
+  - `port`: Port number Relp is listening to. Default to `20514`.
+  - `use_cert`: If true, key/certificates are used. Triplets {`ca_cert`, `cert`, `private_key`} and/or {`ca_cert_src`, `cert_src`, `private_key_src`} should be configured. Default to `true`.
```
richm, line 119: Is it possible to use RELP without TLS?  Is it possible to use RELP with TLS but without certs?

If you can use RELP without TLS, then maybe use `tls: true|false` instead of `use_cert`?  Unless it is somehow possible to use RELP with TLS but with no certs, but that would be strange.  I'm assuming RELP is using TLS for encryption, especially since it uses certs/keys.

README.md
```diff
@@ -196,17 +211,32 @@ This is a schematic logging configuration to show log messages from input_nameA
...
-  - `pki_authmode`: Specifying the default network driver authentication mode. `x509/name`, `x509/fingerprint`, `anon` is accepted. Default to `x509/name`.
+  - `pki_authmode`: Specifying the default network driver authentication mode. `x509/name`, `x509/fingerprint`, or `anon` is accepted. Default to `x509/name`.
   - `permitted_server`: Hostname, IP address, fingerprint(sha1) or wildcard DNS domain of the server which this client will be allowed to connect and send logs over TLS. Default to `*.{{ logging_domain }}`
 
+- `relp` type - `relp` output sends logs to the remote logging system over the network using relp.<br>
+  **available options**
+  - `port`: Port number Relp is listening to. Default to `20514`.
+  - `server_host`: Host name the remote logging system is running on. **Required**.
+  - `server_port`: Port number the remote logging system is listening to. Default to `20514`.
+  - `use_cert`: If true, key/certificates are used. Triplets {`ca_cert`, `cert`, `private_key`} and/or {`ca_cert_src`, `cert_src`, `private_key_src`} should be configured. Default to `true`.
```
richm, line 230: see above - `tls`

README.md
nhosoi: > Is it possible to use RELP without TLS? Is it possible to use RELP with TLS but without certs?
> 
> If you can use RELP without TLS, then maybe use `tls: true|false` instead of `use_cert`? Unless it is somehow possible to use RELP with TLS but with no certs, but that would be strange. I'm assuming RELP is using TLS for encryption, especially since it uses certs/keys.

Yes, we can use RELP without TLS. And `tls: true|false` should be better than `use_cert`. The thing is I found the set of RELP keys look similar to the one in the elasticsearch output and am hoping to have them in common, where use_cert is defined and used... Do you think it's too late to replace `use_cert` with `tls` in the elasticsearch output? I'd believe no one has even touched the output including the ovirt team. And the elasticsearch output is now available with the ovirt input only.

richm: I think add support for `tls: true|false` to the elasticsearch output AND keep support for `use_cert` - add documentation to the elasticsearch output saying that `use_cert` is deprecated in favor of `tls` and `use_cert` will be removed in the next minor release (e.g. 1.1.0)

sopos: RELP is on the same level as TCP or UDP. The original idea of using protocol variable would fit here as well but that would make other this more complicated. While RELP is going to be a separate type is should look very similar to TCP, though.

README.md
```diff
@@ -116,7 +116,7 @@ This is a schematic logging configuration to show log messages from input_nameA
 - `relp` type - `relp` input supports receiving logs from the remote logging system over the network using relp.<br>
   **available options**
   - `port`: Port number Relp is listening to. Default to `20514`.
-  - `use_cert`: If true, key/certificates are used. Triplets {`ca_cert`, `cert`, `private_key`} and/or {`ca_cert_src`, `cert_src`, `private_key_src`} should be configured. Default to `true`.
+  - `tls`: If true, key/certificates are used. Triplets {`ca_cert`, `cert`, `private_key`} and/or {`ca_cert_src`, `cert_src`, `private_key_src`} should be configured. Default to `true`.
```
richm, line 119: "If true, encrypt the connection with TLS.  You must provide key/certificates, and triplets ....."

README.md
```diff
@@ -184,7 +184,8 @@ This is a schematic logging configuration to show log messages from input_nameA
   - `index_prefix`: Elasticsearch index prefix the particular log will be indexed to. **Required**.
   - `input_type`: Specifying the input type. Currently only type `ovirt` is supported. Default to `ovirt`.
   - `retryfailures`: Specifying whether retries or not in case of failure. Allowed value is `true` or `false`.  Default to `true`.
-  - `use_cert`: If true, key/certificates are used to access Elasticsearch. Triplets {`ca_cert`, `cert`, `private_key`} and/or {`ca_cert_src`, `cert_src`, `private_key_src`} should be configured. Default to `true`.
+  - `tls`: If true, key/certificates are used to access Elasticsearch. Triplets {`ca_cert`, `cert`, `private_key`} and/or {`ca_cert_src`, `cert_src`, `private_key_src`} should be configured. Default to `true`.
```
richm, line 187: see above

nhosoi:
Thank you, @richm. I updated the README file and squashed the commits into one.

roles/rsyslog/templates/output_elasticsearch.j2
```diff
@@ -56,10 +56,33 @@ ruleset(name="{{ item.name }}") {
             retryfailures="off"
 {% endif %}
             usehttps="{{ item.usehttps | default("on") }}"
-{% if item.use_cert | default(true) %}
-            tls.cacert="{{ item.ca_cert | default('/etc/rsyslog.d/es-ca.crt') }}"
-            tls.mycert="{{ item.cert | default('/etc/rsyslog.d/es-cert.pem') }}"
-            tls.myprivkey="{{ item.private_key | default('/etc/rsyslog.d/es-key.pem') }}"
+
+
+{% if (item.tls is defined and item.tls) or (item.use_cert is defined and item.use_cert) or (item.tls is not defined and item.use_cert | d(true)) %}
```
richm, line 61: I think this is a good case for using `ternary`:
```
if (item.tls is defined) | ternary(item.tls, item.use_cert | d(true))
```

richm (approved):
lgtm

nhosoi:
Thank you, @richm! Merged.

sopos:
When setting an non-standard port, a SELinux policy should be updated. This applies to UDP and TCP as well. Do we have this covered? Or what's our stand here? May be we could use selinux system role for that. Or may be we need to update the docs to instruct user to use selinux role to label ports accordingly.

richm:
> When setting an non-standard port, a SELinux policy should be updated. This applies to UDP and TCP as well. Do we have this covered? Or what's our stand here? May be we could use selinux system role for that. Or may be we need to update the docs to instruct user to use selinux role to label ports accordingly.

Yes.  We generally require the user to do the selinux labeling outside of the role if they are using non-standard ports, paths, etc.  We have the same issue in certificate role, and other roles.

nhosoi:
> When setting an non-standard port, a SELinux policy should be updated. This applies to UDP and TCP as well. Do we have this covered? Or what's our stand here? May be we could use selinux system role for that. Or may be we need to update the docs to instruct user to use selinux role to label ports accordingly.

@sopos, a good point. We should be able to recommend to run selinux role prior to the logging role. I'd think just mentioning this section works?
https://github.com/linux-system-roles/selinux/
```
Set SELinux ports
selinux_ports:
  - { ports: '22100', proto: 'tcp', setype: 'ssh_port_t', state: 'present' }
```

sopos:
> @sopos, a good point. We should be able to recommend to run selinux role prior to the logging role. I'd think just mentioning this section works?
> https://github.com/linux-system-roles/selinux/
> 
> ```
> Set SELinux ports
> selinux_ports:
>   - { ports: '22100', proto: 'tcp', setype: 'ssh_port_t', state: 'present' }
> ```

Yes, that could be enough together with a note that the port label should be `syslogd_port_t` or `syslog_tls_port_t`.

nhosoi:
@sopos, @richm, please review https://github.com/linux-system-roles/logging/pull/174.
Thanks!

PR #175, opened 2020-09-11
Fixing a logic bug in elasticsearch output template.

nhosoi:
When evaluated, the retryfailures value was denied by "not", which
should not have been. Removing the "not" and adding a test case to
tests_files_elasticsearch_use_local_cert.yml.


nhosoi:
Hi @sradco,
This is the bug I mentioned in the email. I also added a test case for the `on|off` value of retryfailures, which you are concerned about. I'd appreciate if you could review this pr. Thanks!

richm (approved):
lgtm

nhosoi:
Thank you for your reviewing the pr, @richm.

@sradco, if you want to configure `retryfailure = "on"` for omelasticsearch in the ovirt logging config test using 8.3 rpm, please do not set retryfailures in the elasticsearch logging_outputs. Then, it'll pick up the default value, which is "on". Sorry for the inconvenience. This will be fixed in 8.3.1.

PR #176, opened 2020-09-24
Instead of having four symlinks in tests/roles/linux-system-roles.log…

nhosoi:
…ging,

changing tests/roles/linux-system-roles.logging itself a symlink as most
system roles do.


richm (approved):
lgtm

nhosoi:
Thank you, @richm. Merging.

PR #180, opened 2020-10-14
lock ansible-lint version at 4.3.5; suppress role name lint warning

richm:
lock the ansible-lint version to use the latest version 4.3.5 which
is compatible with the latest ansible
This means we need to suppress the error about the role name, and
the error about missing `mode` in `file` and `template` usage.


nhosoi (approved):
lgtm

PR #193, opened 2020-12-23
use tox for ansible-lint instead of molecule

richm:
Deprecate the use of molecule for linting.  Use tox for ansible-lint.
The next step will be to remove the ansible-lint dependency from
the molecule environments.


nhosoi (approved):
lgtm

PR #200, opened 2021-01-25
Make the var load test compatible with old Jinja2 (2.7)

pcahyna:
Remove the case_sensitive=true parameter to the unique filter, not supported in
older Jinja2. Case sensitivity is not necessary here.

See linux-system-roles/storage#156, linux-system-roles/template#44


richm (approved):
lgtm

PR #201, opened 2021-01-27
support jinja 2.7

richm:
Do not use the case_sensitive argument to the unique filter
For doing selectattr with string comparison, use `match` with
an argument like `^STRING$` instead of `==` or `equalto` with
`STRING`.
For doing selectattr with boolean values, you don't need to
provide a test - `selectattr('boolvar')` is the same as
`selectattr('boolvar', '==', true)`.  For `false` comparison,
use `rejectattr('boolvar')` which is the same as
selectattr('boolvar', '==', false)`
Do not use `namespace` as it is only needed with loops, not `if`

### DIFF
--- a/roles/rsyslog/tasks/inputs/remote/main.yml
+++ b/roles/rsyslog/tasks/inputs/remote/main.yml
@@ -8,7 +8,7 @@
       when:
         - item | length > 1
   vars:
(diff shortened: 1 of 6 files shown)


nhosoi (approved):
lgtm

PR #206, opened 2021-02-03
Bug fixes in relp_input and relp_output

nhosoi:
This PR contains 2 commits:
1) Bug fix in relp_input - an unwanted loop index was added as suffix
    
    There was a bug in roles/rsyslog/templates/input_relp.j2, in which
    $inputname was compared with a name with extra _{{ loop.index }} to
    determine the output. It made the relp input never delivered to the
    output.
    
    Test case is added to TEST CASE 0 in tests_relp.yml.

2) Bug fix in relp_output - there was inconsistencies in the parameters.
    
    To specify the server host and server port, relp required "server_host"
    and "server_port", while the forwards output does "target" and "port".
    To eliminate the inconsistency, adjusting the relp parameters to the
    forwards output's parameters.
    
    Test case is added to TEST CASE 1 in tests_relp.yml.

### DIFF
--- a/roles/rsyslog/templates/input_relp.j2
+++ b/roles/rsyslog/templates/input_relp.j2
@@ -49,7 +49,7 @@ if
 {%         if not loop.first %}
   or
 {%         endif %}
-  ($inputname == "{{ item.name }}_{{ loop.index }}" )
+  ($inputname == "{{ item.name }}" )
 {%       endif %}
 {%     endfor %}
   then {
--- a/README.md
+++ b/README.md
@@ -252,8 +252,8 @@ This is a schematic logging configuration to show log messages from input_nameA 
 - `relp` type - `relp` output sends logs to the remote logging system over the network using relp.<br>
   **available options**
   - `port`: Port number Relp is listening to. Default to `20514`.
-  - `server_host`: Host name the remote logging system is running on. **Required**.
-  - `server_port`: Port number the remote logging system is listening to. Default to `20514`.
+  - `target`: Host name the remote logging system is running on. **Required**.
+  - `port`: Port number the remote logging system is listening to. Default to `20514`.
   - `tls`: If true, encrypt the connection with TLS. You must provide key/certificates and triplets {`ca_cert`, `cert`, `private_key`} and/or {`ca_cert_src`, `cert_src`, `private_key_src`}. Default to `true`.
   - `ca_cert`: Path to CA cert to configure Relp with tls. Default to `logging_config_dir/basename of ca_cert_src`.
   - `cert`: Path to cert to configure Relp with tls.  Default to `logging_config_dir/basename of cert_src`.
@@ -698,7 +698,7 @@ Deploying `basics input` reading logs from systemd journal and `relp output` to 
     logging_outputs:
       - name: relp_client
         type: relp
-        server_host: logging.server.com
+        target: logging.server.com
         port: 20514
         tls: true
         ca_cert_src: /path/to/ca.pem
--- a/roles/rsyslog/templates/output_relp.j2
+++ b/roles/rsyslog/templates/output_relp.j2
@@ -1,8 +1,8 @@
 ruleset(name="{{ item.name }}") {
     action(name="{{ item.name }}"
            type="omrelp"
-           target="{{ item.server_host }}"
-           port="{{ item.server_port | d(20514) | int }}"
+           target="{{ item.target }}"
+           port="{{ item.port | d(20514) | int }}"
 {% if item.tls | default(true) %}
 {%   if item.ca_cert is defined %}
 {%     set __cacert = item.ca_cert %}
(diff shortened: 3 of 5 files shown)


roles/rsyslog/templates/input_relp.j2
```diff
@@ -49,7 +49,7 @@ if
 {%         if not loop.first %}
   or
 {%         endif %}
-  ($inputname == "{{ item.name }}_{{ loop.index }}" )
+  ( $inputname == "{{ item.name }}" )
```
richm, line 52: The rsyslog templates are not consistent with whitespace - some templates use `($inputname == something)` - some use `($inputname == something )` and some use `( $inputname == something)` etc.

I would prefer not to change the whitespace here at this time - keep it as `($inputname == something )` - we might revisit this issue in the future to make all of the templates consistent

richm:
This is a major change to the api - it introduces breaking changes - I think it is worth it since the logging role relp/forward feature is new and this corrects a problem with the api - but the next version of logging will have to bump the major release - we also must change this in the docs, release notes, etc. - we need to make sure we communicate this change to users

roles/rsyslog/templates/input_relp.j2
nhosoi: Yeah, I also noticed the whitespace inconistencies in multiple templates... (I was hoping to fix all of them together, but I registed. :)

I thought this particular line is ok since the change is covered by the real bug fix... But if you prefer to keep it, I'll revert the whitespace change.

nhosoi:
> This is a major change to the api - it introduces breaking changes - I think it is worth it since the logging role relp/forward feature is new and this corrects a problem with the api - but the next version of logging will have to bump the major release - we also must change this in the docs, release notes, etc. - we need to make sure we communicate this change to users

I thought relp was not public yet, but you are right, @richm. It's in ansible-galaxy... (tag 1.1.0)

Actually, there was a bug in the example in README and tests_relp.yml, where I mistakenly used `port` for `server_port` there. It was found in the downstream tests run by @sopos. Instead of just fixing them in README and tests_relp.yml, he proposed to make `server_host` and `server_port` in `relp` match the coresponding parameters in `forwards`. (On the other hand, in the `elasticsearch` output's parameter, they are `server_host` and `server_port`. But probably, `relp` is closer to `forwards` compared to `elasticsearch`.)

I have to ask the doc team if the relp section is already finished or not. If it is, I'll ask them to update. Not sure about the release note. Do we have a release note in ansible galaxy?

roles/rsyslog/templates/input_relp.j2
richm: > But if you prefer to keep it, I'll revert the whitespace change.

Let's first agree on a whitespace convention to use for generating rsyslog configs, then we can go back and make them all consistent.

richm:
> But probably, relp is closer to forwards compared to elasticsearch.

Yes - so I think this change is the right thing to do - but we have to make sure it is communicated to users

> Not sure about the release note. Do we have a release note in ansible galaxy?

No, just downstream.  The major version change should be a clue to users that something major has changed.

nhosoi:
@richm,
> > Not sure about the release note. Do we have a release note in ansible galaxy?
> 
> No, just downstream. The major version change should be a clue to users that something major has changed.

Ok, if that's just for downstream, since RELP is a new feature, there's no users yet. I guess we are good, then?

richm:
> Ok, if that's just for downstream, since RELP is a new feature, there's no users yet. I guess we are good, then?

Ah, right.  It wasn't in rhel 8.3 or 8.3.1 - so yes, we're good

> And we are going to tag this version as 2.0.0, rather than 1.3.0?

Yes

roles/rsyslog/templates/input_relp.j2
nhosoi: Reverted the whitespace change.

Regarding "a whitespace convention to use for generating rsyslog configs", do we have some guidance in the rsyslog team, @sopos? (I'm not planning to solve the problem in this pr, but I'd like to have some inputs for the preparation.)

Looking into the examples in the rsyslog docs, they do not care much about whiltespaces... E.g., in the first example in https://www.rsyslog.com/doc/master/rainerscript/control_structures.html, the first `if` does not have an extra whitespace after `(` and before `)`, but the second `if` does. But at least, it's consistent, of course. If it has one after `(`, then there is another before `)`.
```
if ($msg contains "important") then {
   if ( $.foo != "" ) then set $.foo = $.bar & $.baz;
   action(type="omfile" file="/var/log/important.log" template="outfmt")
}
```

richm (approved):
lgtm

nhosoi:
Thank you, @richm! Merging.

roles/rsyslog/templates/input_relp.j2
sopos: no guidance here, no strong opinion neither

nhosoi: > no guidance here, no strong opinion neither

Thanks, @sopos.

@richm, is it ok to have one whitespace like this? And open a pr?
`( $inputname == something )`

richm: > @richm, is it ok to have one whitespace like this for the control code such as in if and foreach expression? And open a pr?

yes

Issue #208, opened 2021-02-04
Default remote_log_path should use FROMHOST instead of HOSTNAME


richm:
When rsyslog receives a log from a remote host using one of the remote input modules like imfwd, imrelp, etc. the fields that identify the remote host are `FROMHOST` and `FROMHOST-IP`.  The current docs/code use `HOSTNAME` which is the hostname of the server host.  The code should use e.g.
```
roles/rsyslog/templates/output_remote_files.j2:  string="{{ __remote_log_path }}/msg/%FROMHOST%/%PROGRAMNAME:::secpath-replace%.log"
```
This will have to be changed in the README, in roles/rsyslog/templates/output_remote_files.j2, and in tests/tests_remote.yml

PR #209, opened 2021-02-04
use FROMHOST instead of HOSTNAME for remote files log file name

richm:
resolves https://github.com/linux-system-roles/logging/issues/208


nhosoi (approved):
Thanks for the issue/fix, @richm!
lgtm

PR #210, opened 2021-02-05
Fixing bugs in permitted_servers and permitted_clients parameter in relp.

nhosoi:
There were logic bugs in input_relp.j2 and output_relp.j2 to set default values.
The bugs are fixed and the test cases are added to tests_relp.yml.

### DIFF
--- a/tests/tests_relp.yml
+++ b/tests/tests_relp.yml
@@ -47,9 +47,6 @@
             cert: "{{ __test_cert }}"
             private_key: "{{ __test_key }}"
             pki_authmode: name
-            permitted_clients:
-              - '*.client.com'
-              - '*.example.com'
         logging_outputs:
           - name: files_output
             type: files
@@ -63,6 +60,7 @@
             outputs: [files_output, forwards_output]
       include_role:
         name: linux-system-roles.logging
+        public: yes
 
     # notify restart rsyslogd is executed at the end of this test task.
     # thus we have to force to invoke handlers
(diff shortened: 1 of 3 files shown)


tests/tests_relp.yml
```diff
@@ -47,9 +47,6 @@
             cert: "{{ __test_cert }}"
             private_key: "{{ __test_key }}"
             pki_authmode: name
-            permitted_clients:
-              - '*.client.com'
-              - '*.example.com'
```
richm, line 52: If these are deleted from here, where are they set?

nhosoi: It is set here:  https://github.com/linux-system-roles/logging/pull/210/files#diff-1a6cdc6949743db21fb21f528511be7e51a1ebc1117b0b90568de7bbdde43552R35

And the value would be `[ "*.$( hostname -d )" ]`.

richm: hmm - ok - then not sure why the relp tests are failing in CI

nhosoi: > hmm - ok - then not sure why the relp tests are failing in CI

Me, too... Temporarily pushed a debugging commit...

nhosoi:
[citest bad]

nhosoi:
[citest bad]

tests/tests_relp.yml
nhosoi: @richm, could you please review one more time? The failure on fedora 33 host was there was no domain configured.
```
Note: if the target node has no domain configured [0], ansible_domain is set
to localdomain, although `hostname -d` returns none. It must be considered
in the test cases.
[0]
  # grep domain /etc/resolv.conf
  #
```

nhosoi:
[citest pending]

tests/tests_relp.yml
```diff
@@ -113,6 +110,21 @@
...
+      shell: >-
+        grep 'tls.permittedpeer=\["\*.client.com","\*.example.com"\]'
+        "{{ __test_relp_server0 }}" | wc -l
+      register: __result
+      failed_when: __result.stdout != "1"
+
+    - name: Check tls.permittedpeer in relp_server1
+      shell: >-
+        grep "tls.permittedpeer=\[\"\*.$( _domain=$(hostname -d);
+        _domain=${_domain:-"localdomain"}; echo $_domain )\"\]"
```
richm, line 123: Can you use the value of `logging_domain` here?  If not - is the problem that the `logging_domain` value is not exported from the role, for use in testing?  If so, you can use the `public: true` option of `include_role` to export the vars and defaults - https://docs.ansible.com/ansible/2.9/modules/include_role_module.html#include-role-module

nhosoi: Thank you, @richm. It worked!

nhosoi:
[citest]

nhosoi:
[citest pending]

richm (approved):
lgtm

nhosoi:
Thank you, @richm!

Issue #205, opened 2021-02-03
RELP: cert and key files should be deployed in the /etc/pki/tls folder


sopos:
Currently, the cert and key files are stored in _/etc/rsyslog.d_.
This is not a functional issue but the files does not get the usual selinux labels. It gets _syslog_conf_t_ which grants rsyslogd the read permissions. It is just not a good practice.

nhosoi:
Since https://github.com/linux-system-roles/logging/pull/211 is merged, closing this issue.

PR #211, opened 2021-02-05
Issue: cert and key files should be deployed in the /etc/pki/tls folder

nhosoi:
When ca_cert_src, cert_src, and private_key_src are specified without the
corresponding target locations ca_cert, cert, and private_key, the files
are copied to /etc/rsyslog.d for the elasticsearch output and the relp
input/output. On the other hand, logging_pki_files copies to /etc/pki/tls.

The inconsistent behaviour should confuse the users. Since relp is new,
and we are releasing the next version as a major update (with major_version++),
it is a good timing to adjust it.

README and test cases are updated.

https://github.com/linux-system-roles/logging/issues/205


richm (approved):
lgtm

nhosoi:
Sorry, some tests were not updated. Fixing them.

nhosoi:
[citest pending]

richm:
Don't worry about the `(staging)` statuses - all of the required tests have passed

nhosoi:
Thank you for your reviewing, @richm.

PR #217, opened 2021-02-19
RHELPLAN-68108 - Collections - Logging - fixing ansible-test errors

nhosoi:
- Adding roles/rsyslog/meta/main.yml for the private sub-role rsyslog
- Removing obsolete molecule/default/{INSTALL.rst,playbook.yml,yaml-lint.yml}
- Fixing an ansible-lint error on the double-quotes
- Fixing a shellcheck error


tests/setup_module_utils.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -eu
 # SPDX-License-Identifier: MIT
 
 set -euo pipefail
```
richm, line 4: Then I guess you should change this to `set -o pipefail` since the `eu` are already in the bash invocation

richm (approved):
lgtm

nhosoi:
Thank you for your reviews, @richm.

PR #218, opened 2021-02-22
A bug fix brought in in commit da399b95287ea1623ca844c91e3368372b2e60…

nhosoi:
…c9 (tls)

Issue: cert and key files should be deployed in the /etc/pki/tls folder
https://github.com/linux-system-roles/logging/issues/205

The default private key path should be /etc/pki/tls/private, but it was
set to /etc/pki/tls/certs.

### DIFF
--- a/roles/rsyslog/tasks/set_certs.yml
+++ b/roles/rsyslog/tasks/set_certs.yml
@@ -20,7 +20,7 @@
     - name: Copy key on the control host to the specified path on the target host
       copy:
         src: '{{ item.private_key_src }}'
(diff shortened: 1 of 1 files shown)


sopos:
looks correct to me

sopos (approved):

richm (approved):
lgtm

nhosoi:
Thank you for your reviews, @sopos and @richm.

sopos:
I also applied the fix manually and ran my test - it is ok

nhosoi:
> I also applied the fix manually and ran my test - it is ok

Thank you for your pre-testing, @sopos. What a relief...

PR #222, opened 2021-03-02
use tox-lsr 2.3.0 and enable ansible-test

nhosoi:
- Upgrade to tox-lsr 2.3.0 which fixes several issues with running ansible-test sanity from tox.
- Add ansible-test to the python 3.6 test list.


richm (approved):
lgtm

PR #223, opened 2021-03-23
Cleaning up yamllint line-length errors.

nhosoi:


roles/rsyslog/tasks/inputs/files/main.yml
```diff
@@ -15,9 +16,10 @@
         type: input
         state: "{{ item.state | d('present') }}"
         sections:
-          - options: |-
-              input(type="imfile" file="{{ item.input_log_path }}" tag="{{ item.name }}")
-              {{ lookup('template', 'input_template.j2') }}
+          - options: 'input(type="imfile"
```
richm, line 19: I think `|-` is ok here

roles/rsyslog/tasks/inputs/files/main.yml
```diff
@@ -15,9 +16,10 @@
...
+            file="{{ item.input_log_path }}" tag="{{ item.name }}")
+
```
richm, line 21: blank line here?

roles/rsyslog/vars/inputs/ovirt/main.yml
```diff
@@ -70,19 +77,27 @@ rsyslog_rulebase_ovirt_engine:
     filename: 'ovirt_engine.rulebase'
     nocomment: true
     sections:
-      - options: |-
-          version=2
-          rule=:%date:date-iso% %time:time-24hr%,%msec:number%%tz:word% %level:word%%-:whitespace%%class:string{"quoting.char.begin":"[","quoting.char.end":"]"}% %thread:string{"quoting.char.begin":"(","quoting.char.end":")"}% %correlationid:string{"quoting.char.begin":"[","quoting.char.end":"]"}% %message:rest%
+      - options: 'version=2
+
```
richm, line 81: blank line?

roles/rsyslog/vars/inputs/ovirt/main.yml
```diff
@@ -70,19 +77,27 @@ rsyslog_rulebase_ovirt_engine:
...
+        rule=:%date:date-iso% %time:time-24hr%,%msec:number%%tz:word%
+        %level:word%%-:whitespace%%class:string{"quoting.char.begin":\
+        "[","quoting.char.end":"]"}% %thread:string{"quoting.char.begin":\
+        "(","quoting.char.end":")"}%
```
richm, line 85: do you need a `\` at end of line here?

tests/tasks/test_logger.yml
```diff
@@ -3,7 +3,8 @@
   vars:
     __logging_tag: "testTag{{ __logging_index }}"
     __logging_message: "testMessage{{ __logging_index }}"
-  command: /bin/logger -i -p local6.info -t "{{ __logging_tag }}" "{{ __logging_message }}"
+  command: "/bin/logger -i -p local6.info -t {{ __logging_tag }}
```
richm, line 6: with command it is better to use `>-` which will generate a single line string with each newline replaced with a single space

tests/tests_basics_files.yml
```diff
@@ -47,7 +51,9 @@
       include_tasks: tasks/check_packages.yml
 
     - name: Grep output to messages line
-      command: /bin/grep '\*.info;mail.none;authpriv.none;cron.none.*{{ __default_system_log }}' '{{ __test_default_files_conf }}'
+      command: "/bin/grep
```
richm, line 54: with `command` it is better to use `>-` which will generate a single line string with each newline replaced with a single space

tests/tests_basics_files.yml
```diff
@@ -119,8 +128,12 @@
         __logging_file: "{{ __logging_system_log_dir }}/messages"
       include_tasks: tasks/test_logger.yml
 
-    - name: Check the files output config that the path is {{ __logging_system_log_dir }}/messages
-      command: /bin/grep '\*.info;mail.none;authpriv.none;cron.none.*{{ __logging_system_log_dir }}/messages' {{ __test_files_conf0 }}
+    - name: "Check the files output config that the path is
+      {{ __logging_system_log_dir }}/messages"
+      command: "/bin/grep
```
richm, line 133: with `command` it is better to use `>-` which will generate a single line string with each newline replaced with a single space

tests/tests_basics_files.yml
```diff
@@ -305,7 +333,10 @@
       include_tasks: tasks/check_packages.yml
 
     - name: Check the filter
-      command: /bin/grep '\*.info;authpriv.none;auth.none;cron.none;mail.none.*{{ __default_system_log }}' '{{ __test_files_conf0 }}'
+      command: "/bin/grep
```
richm, line 336: with `command` it is better to use `>-` which will generate a single line string with each newline replaced with a single space

tests/tests_basics_forwards.yml
```diff
@@ -571,7 +601,8 @@
             msg: "Caught an expected error - {{ ansible_failed_result }}"
         - name: assert...
           assert:
-            that: "'{{ ansible_failed_result.results.0.msg }}' is match('{{ __expected_error }}')"
+            that: "'{{ ansible_failed_result.results.0.msg }}'
```
richm, line 604: `that: ansible_failed_result.results.0.msg is match(__expected_error)`

richm: ansible parses the value of `when` and `that` as jinja expressions directly

tests/tests_combination.yml
```diff
@@ -96,7 +104,10 @@
       failed_when: not __result.stat.exists
 
     - name: Check the filter and action in the files output config
-      command: /bin/grep '\*.info;authpriv.none;auth.none;cron.none;mail.none.*{{ __default_system_log }}' '{{ __test_files_conf }}'
+      command: "/bin/grep
```
richm, line 107: with `command` it is better to use `>-` which will generate a single line string with each newline replaced with a single space

tests/tests_combination.yml
```diff
@@ -130,15 +141,25 @@
       failed_when: not __result.stat.exists
 
     - name: Check the input call with tag={{ __test_tag }}
-      command: /bin/grep ' *input(type="imfile" file="\/var\/log\/inputdirectory\/\*.log" tag="{{ __test_tag }}"' '{{ __test_inputfiles_conf }}'
+      command: "/bin/grep
```
richm, line 144: with `command` it is better to use `>-` which will generate a single line string with each newline replaced with a single space

tests/tests_combination.yml
```diff
@@ -130,15 +141,25 @@
...
+        file=\"/var/log/inputdirectory/\\*.log\"
+        tag=\"{{ __test_tag }}\"'
+        {{ __test_inputfiles_conf }}"
       changed_when: false
 
-    - name: Create a test log file with a log message in {{ __test_inputfiles_dir }}
-      shell: echo "<167>Mar 20 01:00:00 11.22.33.44 tag msgnum:00000000:26:abcdefghijklmnopqrstuvwxyz" > "{{ __test_inputfiles_dir }}/test.log"
+    - name: "Create a test log file with a log message in
+      {{ __test_inputfiles_dir }}"
+      shell: "echo
```
richm, line 153: with `shell` use `|-` or `>-` - I prefer `|-` when using `if` or `for` or other constructs

tests/tests_combination.yml
```diff
@@ -130,15 +141,25 @@
...
+        '<167>Mar 20 01:00:00 11.22.33.44 tag \
+        msgnum:00000000:26:abcdefghijklmnopqrstuvwxyz'
+        > {{ __test_inputfiles_dir }}/test.log"
       changed_when: false
 
     - name: Check the fake test log message in {{ __default_system_log }}
-      command: /bin/grep '{{ __test_tag }} .*abcdefghijklmnopqrstuvwxyz$' '{{ __default_system_log }}'
+      command: "/bin/grep
```
richm, line 160: with `command` it is better to use `>-` which will generate a single line string with each newline replaced with a single space

tests/tests_combination.yml
```diff
@@ -275,7 +299,10 @@
       failed_when: not __result.stat.exists
 
     - name: Grep output to messages line
-      command: /bin/grep '*.info;authpriv.none;auth.none;cron.none;mail.none.*{{ __default_system_log }}' '{{ __test_files_conf }}'
+      command: "/bin/grep
```
richm, line 302: with `command` it is better to use `>-` which will generate a single line string with each newline replaced with a single space

tests/tests_combination.yml
```diff
@@ -315,15 +342,25 @@
       failed_when: not __result.stat.exists
 
     - name: Check the input call with tag={{ __test_tag }}
-      command: /bin/grep ' *input(type="imfile" file="\/var\/log\/inputdirectory\/\*.log" tag="{{ __test_tag }}"' '{{ __test_inputfiles_conf }}'
+      command: "/bin/grep
```
richm, line 345: with `command` it is better to use `>-` which will generate a single line string with each newline replaced with a single space

tests/tests_combination.yml
```diff
@@ -315,15 +342,25 @@
...
+      {{ __test_inputfiles_dir }}"
+      shell: "echo
+        '<167>Dec 26 01:00:00 11.22.33.44 tag
+        msgnum:00000001:26:ABCDEFGHIJKLMNOPQRSTUVWXYZ'
+        >> {{ __test_inputfiles_dir }}/test.log"
       changed_when: false
 
     - name: Check the fake test log message in {{ __default_system_log }}
-      command: /bin/grep '{{ __test_tag }} .*ABCDEFGHIJKLMNOPQRSTUVWXYZ$' '{{ __default_system_log }}'
+      command: "/bin/grep
```
richm, line 361: with `command` it is better to use `>-` which will generate a single line string with each newline replaced with a single space

tests/tests_files_files.yml
```diff
@@ -97,7 +103,8 @@
       failed_when: __result.stat.exists
 
     - name: Grep output to messages line
-      command: /bin/grep "{{ __default_system_log }}" {{ __test_outputfiles_conf }}
+      command: "/bin/grep
```
richm, line 106: with `command` it is better to use `>-` which will generate a single line string with each newline replaced with a single space

tests/tests_files_files.yml
```diff
@@ -113,11 +120,19 @@
       failed_when: not __result.stat.exists
 
     - name: Check the filter 0
-      command: /bin/grep 'input(type="imfile" file="\/var\/log\/inputdirectory0\/\*\.log" tag="files_input0")' '{{ __test_inputfiles_conf0 }}'
+      command: "/bin/grep
```
richm, line 123: with `command` it is better to use `>-` which will generate a single line string with each newline replaced with a single space

roles/rsyslog/tasks/inputs/files/main.yml
nhosoi: I thought otherwise the lines above and below are concatenated... (not tried, though)

tests/tests_files_files.yml
```diff
@@ -113,11 +120,19 @@
...
+        'input(type=\"imfile\"
+        file=\"/var/log/inputdirectory0/\\*.log\"
+        tag=\"files_input0\")'
+        {{ __test_inputfiles_conf0 }}"
       changed_when: false
 
     - name: Check the filter 1
-      command: /bin/grep 'input(type="imfile" file="\/var\/log\/inputdirectory1\/\*\.log" tag="files_input1")' '{{ __test_inputfiles_conf1 }}'
+      command: "/bin/grep
```
richm, line 131: with `command` it is better to use `>-` which will generate a single line string with each newline replaced with a single space

tests/tests_imuxsock_files.yml
```diff
@@ -60,7 +67,9 @@
       include_tasks: tasks/check_packages.yml
 
     - name: Check output to messages line
-      command: /bin/grep '\*.info;mail.none;authpriv.none;cron.none.*{{ __default_system_log }}' '{{ __test_files_conf }}'
+      command: "/bin/grep
```
richm, line 70: with `command` it is better to use `>-` which will generate a single line string with each newline replaced with a single space

roles/rsyslog/tasks/inputs/files/main.yml
nhosoi: ok, let me revert and retry...

roles/rsyslog/vars/inputs/ovirt/main.yml
nhosoi: the same reason...

roles/rsyslog/vars/inputs/ovirt/main.yml
nhosoi: Yes, otherwise a whitespace is added.

roles/rsyslog/tasks/inputs/files/main.yml
richm: it depends - what do you want the output to look like?
```
input(type="imfile" file="{{ item.input_log_path }}" tag="{{ item.name }}")

template begins here
...
```
or like this:
```
input(type="imfile" file="{{ item.input_log_path }}" tag="{{ item.name }}")
template begins here
...
```
or does it matter if the `input` is wrapped in the output?
```
input(type="imfile" file="{{ item.input_log_path }}"
         tag="{{ item.name }}")
template begins here
...
```
If it is the last one, then use `|-`:
```yaml
options: |-
  input(type="imfile" file="{{ item.input_log_path }}"
           tag="{{ item.name }}")
  {{ lookup("template", "input_template.j2") }}
```

roles/rsyslog/vars/inputs/ovirt/main.yml
richm: I think this is a case where we want to ignore the line length checking because the long line is more readable than the wrapped line

or, move this into a file under `files/`

roles/rsyslog/vars/inputs/ovirt/main.yml
richm: I think this is a case where we want to ignore the line length checking because the long line is more readable than the wrapped line

or, move this into a file under files/

tests/tasks/test_logger.yml
nhosoi: Not this command, but I have a requirement to convert to something like this.
```
command: my_command long_arg_012345678901234567890123456789012345678901234567890123456789012345678901234567890123456789 another_arg
```
I could not find a good way to separate the long argument > 80 into multiple lines and concatinate them back to one without white spaces in-between if I use `>` or `|`. Only a solution I found was to make it a string and put `\` at the end of the line where a white space is not needed.

roles/rsyslog/tasks/inputs/files/main.yml
nhosoi: I'd like to have the first or the second one. Not the last one.
So, I think this works.
```
options: >-
  input(type="imfile" file="{{ item.input_log_path }}"
           tag="{{ item.name }}")
  {{ lookup("template", "input_template.j2") }}
```

tests/tasks/test_logger.yml
richm: right - there is no way to do this without using `\`.  That's a good case for leaving the long yaml line length for readability.

roles/rsyslog/tasks/inputs/files/main.yml
```diff
@@ -16,10 +16,10 @@
         type: input
         state: "{{ item.state | d('present') }}"
         sections:
-          - options: 'input(type="imfile"
-            file="{{ item.input_log_path }}" tag="{{ item.name }}")
-
-            {{ lookup("template", "input_template.j2") }}'
+          - options: >-
```
richm, line 19: I think this would be better to use `|-` - what you have works because `>-` knows how to handle the indentation of the `tag` line correctly, and the template expansion happens after `>-` processes the string (or `>-` preserves the newlines in the template).  Using `|-` makes it explicit that you want the line breaks and the indentation preserved as-is.

tests/tests_relp.yml
```diff
@@ -111,22 +111,22 @@
       changed_when: false
 
     - name: Check the flows in relp_server0 - 0
-      shell: >-
-        grep '$inputname == "relp_server0"' "{{ __test_relp_server0 }}" |
+      shell: |-
+        grep '$inputname == "relp_server0"' "{{ __test_relp_server0 }}" | \
```
richm, line 115: it was fine with `>-` and without `\` - unless you ran into some problem?

tests/tests_relp.yml
```diff
@@ -111,22 +111,22 @@
...
-        grep "call files_output" | wc -l
+        grep -A 2 "$inputname == \"relp_server0\"" "{{ __test_relp_server0 }}" \
+        | grep "call files_output" | wc -l
       register: __result
       failed_when: __result.stdout != "1"
 
     - name: Check the ca cert in relp_server0 - 0
-      shell: >-
-        grep 'tls.cacert="/etc/pki/tls/certs/test-ca.crt"'
+      shell: |-
```
richm, line 128: same here - `>-` without `\` should work fine here

tests/tests_relp.yml
nhosoi: @richm, I'm cleaning up ansible-lint errors next, where `[306] Shells that use pipes should set the pipefail option` is one of them, and it requires to add `set -o pipefail`. Using `|-` makes it easier as follows?
```
      shell: |-
        set -o pipefail
        grep -A 2 "$inputname == \"relp_server0\"" "{{ __test_relp_server0 }}" | \
        grep "call files_output" | wc -l
```

richm: yes - using `|-` is recommended for multiline `shell` scripts

richm: in fact, we should use `set -euo pipefail` in all of our `shell` to fail fast and catch use of unset variables

nhosoi: Ok. I'm opening another pr for ansible-lint. I'm going to add `-eu` and/or `set -euo pipefail` to all the tasks using `shell`. Thanks!

jharuda:
[citest bad]

nhosoi:
Added [another commit for ansible-lint fixes](https://github.com/linux-system-roles/logging/pull/223/commits/71a12ae36ff516d0d8506f1f7de4c0edc25b24e2).
```
    Cleaning up ansible-lint errors except 301 and 303.
    - '301'  # Commands should not change things if nothing needs doing
    - '303'  # systemctl used in place of systemd module
```

roles/rsyslog/tasks/main.yml
```diff
@@ -55,14 +55,14 @@
       file:
         state: directory
         path: '{{ __rsyslog_config_dir }}'
-        mode: '0755'
+        mode: preserve
```
richm, line 58: I think this should be explicitly set to a mode - so `mode: '0755'` is ok

You should only use `mode: preserve` when you are absolutely sure that you want to use the file mode from the source for the destination.  In almost every case here, you want to provide an explicit mode.

The only reason we use `mode: preserve` in https://github.com/linux-system-roles/template/blob/master/tests/tests_include_vars_from_parent.yml#L13 is because it is a test file and the actual mode doesn't really matter.  For non-test code (i.e. production code), the mode really does matter.

nhosoi: I was also wondering about it. But `/etc/rsyslog.d`, `rsyslog.conf`, and `/var/lib/rsyslog` are in the rsyslog rpm package. And the logging role ensures the package is installed. So, I thought we are supposed to use `preserve` there. One question is on `/var/lib/rsyslog`, not like the other two, the path is configurable. If a customer happens to change the path and the mode is `preserve` and the directory did not exist, the mode will be the default one.

richm: ok - then I guess `preserve` is ok here.
re: `/var/lib/rsyslog`
- does the rpm create this by default?
- if the user changes this value, is the user responsible for creating the directory and configuring the ownership/permission?

nhosoi: Ahhh, sorry. Never mind. It's not programmable in the logging role level. I misunderstood by the var name not starting with `__` and it's defined in defaults/main.yml instead of vars/main.yml as the other "hardcoded" paths are...

If ok, may I move rename `rsyslog_work_dir` to `__rsyslog_work_dir` and move the definition to vars? Please note that `rsyslog_work_dir` does not appear any README files.

richm: > Ahhh, sorry. Never mind. It's not programmable in the logging role level. I misunderstood by the var name not starting with `__` and it's defined in defaults/main.yml instead of vars/main.yml as the other "hardcoded" paths are...
> 
> If ok, may I move rename `rsyslog_work_dir` to `__rsyslog_work_dir` and move the definition to vars? Please note that `rsyslog_work_dir` does not appear any README files.

yes.  Even though `rsyslog_work_dir` is defined in `roles/rsyslog/defaults/main.yml`, that isn't considered to be part of the public interface of the `logging` role.

nhosoi: Done!
(and *shame shame*, the mode key in the `file` module does not take `preserve`... So, put the value back to the ones defined in the rpm spec files.)

.ansible-lint
```diff
@@ -1,11 +1,4 @@
 ---
 skip_list:
-- '106'
-- '204'
-- '208'
 - '301'  # Commands should not change things if nothing needs doing
```
richm, line 3: are you planning to fix `301` issues?  This will likely be adding `changed_when: false` to tasks that use `command` and `shell` to get information from the system.

.ansible-lint
```diff
@@ -1,11 +1,4 @@
...
 - '303'  # systemctl used in place of systemd module
```
richm, line 4: I'm assuming this will remain - probably tests that use `systemctl` which is ok

roles/rsyslog/tasks/main.yml
```diff
@@ -77,20 +84,23 @@
...
         owner: 'root'
         group: 'root'
-        mode: '0700'
+        mode: '0755'
 
     - name: Generate main rsyslog configuration
       template:
         src: 'rsyslog.conf.j2'
         dest: '/etc/rsyslog.conf'
+        mode: preserve
```
richm, line 103: Note that `preserve` doesn't mean "preserve the existing mode on the existing file" - it means "the mode on the `dest` file should be exactly the same as the mode on the `src` file" which is probably not what you intended - see https://docs.ansible.com/ansible/2.8/modules/copy_module.html#copy-module
"`preserve` means that the file will be given the same permissions as the source file."

You could do this by omitting the `mode`, but then ansible-lint will complain.

I think the only way to keep the existing mode, while at the same time avoiding the ansible-lint warning, is to use the `stat` module to get the file mode, something like this:
```yaml
- name: Get mode of rsyslog.conf if it exists
  stat:
    path: /etc/rsyslog.conf
  register: __rsyslog_register_stat_conf

- name: Generate main rsyslog configuration
  template:
    src: 'rsyslog.conf.j2'
    dest: '/etc/rsyslog.conf'
    mode: "{{ __rsyslog_register_stat_conf.stat.mode | d('0700') }}"
```

roles/rsyslog/vars/inputs/ovirt/main.yml
```diff
@@ -79,10 +87,11 @@ rsyslog_rulebase_ovirt_vdsm:
   - name: 'ovirt_vdsm'
     filename: 'ovirt_vdsm.rulebase'
     nocomment: true
-    sections:
-      - options: |-
+    sections: |-
+      - options:
```
richm, line 91: I think `options: |-` is correct

.ansible-lint
nhosoi: Yeah, `systemd` module does not support `systemctl status` to get the full status message...

nhosoi:
If ok, squashing the multiple commits into one.

richm:
> If ok, squashing the multiple commits into one.

yes - lgtm

nhosoi:
Thank you, @richm. Merging...

PR #226, opened 2021-05-20
Fixing "logging README.html examples are rendered incorrectly"

nhosoi:
As suggested, removed numbering and itemizing.

RHELPLAN-80116


nhosoi:
logging_readme

richm (approved):
lgtm

nhosoi:
There's another issue in rendering. Sadly, no links work in the rendered page even the ones inside the same readme file... Should we remove the links such as table of contents?

richm:
> There's another issue in rendering. Sadly, no links work in the rendered page even the ones inside the same readme file... Should we remove the links such as table of contents?

yes, if there is no way to make them work

nhosoi:
> > There's another issue in rendering. Sadly, no links work in the rendered page even the ones inside the same readme file... Should we remove the links such as table of contents?
> 
> yes, if there is no way to make them work

I found other collections in AH do not use the local links (#tag), while we do in these roles. The local links in these readme files do not work on AH. I'd think we should modify them all together just in the collection artifact for AH as the links are useful in /usr/share/doc/rhel-system-roles/*.
```
certificate/README.md
cockpit/README.md
ha_cluster/README.md
logging/README.md
network/README.md
storage/README.md
vpn/README.md
```
So, merging this pr for now.

PR #227, opened 2021-05-22
Fixing "logging README.html is rendered incorrectly"

nhosoi:
Options are rendered incorrectly, as well

Instead of using the nested itemize, use headers as h5, h6 are supported by kramdoc.

No wording changes. Modified just markdown codes.

RHELPLAN-80116


richm (approved):
lgtm

nhosoi:
Thank you, @richm! Merging...

PR #230, opened 2021-07-16
Instead of the archive module, use "tar" command for backup.

nhosoi:
Note: having the module 'archive' makes the logging role fail with
an error "couldn't resolve module/action 'archive'." if executed
with ansible-navigator.

Stop using set_fact to set rsyslog_backup_dir in roles/rsyslog/tasks/main.yml
to fix a bug that the backup dir could have repeated
rsyslog.d-{{ ansible_date_time.iso8601_basic_short }}
sub-directories if the logging role is executed multiple times.

Move "logging_purge_confs: true" to TEST CASE 3 in tests_basic_files.yml.

### DIFF
--- a/roles/rsyslog/tasks/main.yml
+++ b/roles/rsyslog/tasks/main.yml
@@ -64,28 +64,31 @@
         path: '{{ __rsyslog_work_dir }}'
         mode: '0700'
 
-    # Creating a backup dir for rsyslog.d
-    - name: Set backup dir name
-      set_fact:
-        rsyslog_backup_dir: '{{ rsyslog_backup_dir |
-          d("/tmp") }}/rsyslog.d-{{ ansible_date_time.iso8601_basic_short }}'
-
-    - name: Create a backup dir
-      file:
-        state: directory
-        path: '{{ rsyslog_backup_dir }}'
-        mode: '0700'
-      changed_when: false
-
     # Back up the pre-existing rsyslog config files in the backup dir,
     # then removing the files/dirs in rsyslog.d.
-    - name: Archive the contents of {{ __rsyslog_config_dir }} to the backup dir
-      archive:
-        path: ["{{ __rsyslog_config_dir }}", /etc/rsyslog.conf]
-        dest: "{{ rsyslog_backup_dir }}/backup.tgz"
-        remove: '{{ true if __rsyslog_purge_original_conf | bool else false }}'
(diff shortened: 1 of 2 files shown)


richm:
I'm not sure what you mean by "is not used nor advertised.".  It is currently "used" in that a backup is made every time you run the logging role.  We don't know if customers are relying on this.  We don't know if removing this will break or otherwise cause problems for ovirt.

The `archive` module is provided in the `community.general` collection now - https://docs.ansible.com/ansible/latest/collections/community/general/archive_module.html#ansible-collections-community-general-archive-module - this is problematic because it isn't supported in Automation Hub, so it makes sense to look for an alternative.

nhosoi:
> I'm not sure what you mean by "is not used nor advertised.". It is currently "used" in that a backup is made every time you run the logging role. We don't know if customers are relying on this. We don't know if removing this will break or otherwise cause problems for ovirt.

We did not mention it in the doc (README) at all. I'll ask the oVirt team as well. But even if there's no customers relying on whom we are not aware of, do you think it's not good to remove it?

richm:
> > I'm not sure what you mean by "is not used nor advertised.". It is currently "used" in that a backup is made every time you run the logging role. We don't know if customers are relying on this. We don't know if removing this will break or otherwise cause problems for ovirt.
> 
> We did not mention it in the doc (README) at all. I'll ask the oVirt team as well. But even if there's no customers relying on whom we are not aware of, do you think it's not good to remove it?

It is generally not a good idea to remove functionality from the product that has a visible effect, however hidden and undocumented it is, because a general rule of software is that there is some customer, somewhere, that has discovered this feature and has incorporated it into her/his workflow, and removing it suddenly will cause a problem.  "Hey!  I was relying on the fact that my config was always backed up when I ran the logging role, and I have no backups, and I lost my old configs that I wanted to save!"  A reasonable response is to say "well, shame on you for relying on undocumented behavior" but we generally don't like to have that conversation with customers.

That being said - if ovirt doesn't care about this feature (and they should test this patch thoroughly to make sure it doesn't break anything), then I think we can remove it, as long as we document it in the release notes and in the product documentation.

richm (approved):
line-length warning, otherwise, lgtm

richm (approved):
lgtm

nhosoi:
Thank you for reviewing this pr, @richm. Merging...

PR #232, opened 2021-08-04
use correct python-cryptography package

richm:
Use `python-cryptography` on EL7 - use `python-cryptography3` everywhere
else
https://bugzilla.redhat.com/show_bug.cgi?id=1989962

PR #233, opened 2021-08-04
create_tests_certs: Ensure python3-cryptography exists in RedHat/CentOS-8 and newer

nhosoi:
Test cases that requires the TLS configuration failed due to the missing
the python3-cryptography package. The package is not installed by
default. This patch ensures it is installed.

bz1989962


nhosoi:
Dup of https://github.com/linux-system-roles/logging/pull/232. Closing this pr.

Issue #44, opened 2018-12-24
When using rsyslog_purge_original_conf true it does not delete the sub directories


sradco:
When using rsyslog_purge_original_conf true, the files are deleted from the sub-directories, but the directory itself is not deleted.

nhosoi:
Thanks, @sradco.  I could reproduce the problem.

https://github.com/linux-system-roles/logging/pull/48

sradco:
Due to fix  #50 , this issue is no longer relevant. Closing.

PR #234, opened 2021-08-05
[FIX] "Purge original conf" task

Kariton:
`file` module `state=absent` does not work with `path=/some/dir/*` (wildcard)
`file` module should support `state=empty` see: https://github.com/ansible/ansible/issues/18910
`state=empty` is needed here - and should be used later

### DIFF
--- a/roles/rsyslog/tasks/main.yml
+++ b/roles/rsyslog/tasks/main.yml
@@ -85,11 +85,6 @@
           shell: |-
             set -euo pipefail
             /bin/rm -rf {{ __rsyslog_config_dir }}/*
-        # 'file' module "state=absent" does not work with "path=/some/dir/*" (wildcard)
-        # 'file' module should support "state=empty" see: https://github.com/ansible/ansible/issues/18910
-        #  file:
-        #    state: empty
-        #    path: "{{ __rsyslog_config_dir }}"
           when: __rsyslog_purge_original_conf | bool | d(false)
       vars:
         __rsyslog_backup_dir: '{{ rsyslog_backup_dir |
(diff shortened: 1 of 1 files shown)


Kariton:
rel: https://github.com/linux-system-roles/logging/issues/44

Kariton:
i think the problem here is that `files` is literally looking for a file called `*`.

an other option could be a two tasks task:
```
  - name: Get files to Purge
    find:
      paths: "{{ __rsyslog_config_dir }}"
      patterns: "*.conf"
    register: files_to_delete

  - name: Purge original conf
    file:
      path: "{{ item.path }}"
      state: absent
    with_items: "{{ files_to_delete.files }}"
```
(NOT tested)

Kariton:
> Updated sub-roles default config directory to deploy all configuration files to /etc/rsyslog.d/ directory.
https://github.com/linux-system-roles/logging/pull/50#issue-242498937

my PR would work with sub-dirs.
the other way mentioned would not work - AFAIK. (add  `recurse=yes` to `find` *should* do the trick)

nhosoi:
Thank you for you pr, @Kariton.
I have 2 requests for this pr. Could you please take a look at https://github.com/Kariton/logging/pull/3?
Thanks!
--noriko

Kariton:
updated PR to include `set -euo pipefail` and `/bin/rm` in multiline.

richm:
@Kariton looks like my attempt to fix the linter errors has broken github - please rebase and squash your PR on top of the latest master branch

richm:
Another suggested fix - use `rm -rfv` - this way we can see and confirm that the purge is working

richm:
fixed - closed in favor of https://github.com/linux-system-roles/logging/pull/240

PR #235, opened 2021-08-05
[FIX] README false variable name

Kariton:
inside role rsyslog `logging_purge_original_conf` variable is named `logging_purge_confs`

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,7 +400,7 @@ These variables are set in the same level of the `logging_inputs`, `logging_outp
 - `logging_enabled`: When 'true', logging role will deploy specified configuration file set. Default to 'true'.
 - `logging_mark`: Mark message periodically by immark, if set to `true`. Default to `false`.
 - `logging_mark_interval`: Interval for `logging_mark` in seconds. Default to `3600`.
-- `logging_purge_original_conf`: `true` or `false`. If set to `true`, files in /etc/rsyslog.d are purged.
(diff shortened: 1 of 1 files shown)


spetrosi (approved):
Thanks for the PR, lgtm

nhosoi:
Well, I'd think we should discuss about this...
The parameter name is already public for a while and changing it might be a problem (although there was no effect by setting it.) If ok, I'd rather change the source code as proposed in this pr against your pr/234.
https://github.com/Kariton/logging/pull/3

@spetrosi, @richm, what do you think? The downstream doc does not mention `logging_purge_original_conf`, but README.md does. It's available in galaxy and automation hub...

richm:
> Well, I'd think we should discuss about this...
> The parameter name is already public for a while and changing it might be a problem (although there was no effect by setting it.) If ok, I'd rather change the source code as proposed in this pr against your pr/234.
> [Kariton#3](https://github.com/Kariton/logging/pull/3)
> 
> @spetrosi, @richm, what do you think? The downstream doc does not mention `logging_purge_original_conf`, but README.md does. It's available in galaxy and automation hub...

I think we should update the README.md to have the correct parameter name.  I do not think we should change the code to rename `logging_purge_confs` to be `logging_purge_original_conf` as that would be a breaking API change.  If a user figured out how to use `logging_purge_confs` despite the incorrect documentation, they will be unhappy that we have changed the parameter out from under them.

nhosoi (approved):
Thanks for your comments, @richm.
lgtm

PR #237, opened 2021-08-05
python2 renders server_host list incorrectly

richm:
On python2 (el7) when server_host is a list, this:
```
          server={{ item.server_host }}
```
gets rendered as this:
```
          server=[u'logging-es0', u'logging-es1']
```
I think we should not rely on the python/jinja2 list to string
conversion being exactly the same format as the rsyslog string list format.

### DIFF
--- a/roles/rsyslog/templates/output_elasticsearch.j2
+++ b/roles/rsyslog/templates/output_elasticsearch.j2
@@ -36,7 +36,7 @@ ruleset(name="{{ item.name }}") {
 {% if item.server_host is string %}
             server="{{ item.server_host }}"
 {% elif item.server_host is sequence %}
-            server={{ item.server_host }}
+            server=[{% for srv in item.server_host %}"{{ srv }}"{{ '' if loop.last else ', ' }}{% endfor %}]
 {% else %}
             server="logging-es"
 {% endif %}
--- a/tests/tests_ovirt_elasticsearch.yml
+++ b/tests/tests_ovirt_elasticsearch.yml
@@ -427,12 +427,13 @@
 
     - name: Check server param in "{{ __test_es_conf }}"
       command: >-
-        /bin/grep "server=\['logging-es0', 'logging-es1'\]" {{ __test_es_conf }}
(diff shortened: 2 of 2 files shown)


nhosoi:
Does this pr not cover the single quotes vs. double quotes issue https://bugzilla.redhat.com/show_bug.cgi?id=1986460#c10?

I expected single quotes are accepted as well in rsyslog as noted here:
https://www.rsyslog.com/doc/master/rainerscript/constant_strings.html
But it is just in rainerscript, not in the other places, like in the action parameters... :(

This is a snippet of `31-output-elasticsearch-elasticsearch_output_ops.conf` generated by this pr.
```
    if ($.input_type == "ovirt") and ($.index_prefix startswith ".operations.") then {
        action(
            type="omelasticsearch"
            server=['logging-es-ops0', 'logging-es-ops1']
            serverport="9200"
            <<snip>>
```

richm:
> Does this pr not cover the single quotes vs. double quotes issue https://bugzilla.redhat.com/show_bug.cgi?id=1986460#c10?

Yes, but the problem is that assumes that the rsyslog array-of-strings format is exactly the same as the JSON format - now and in the future - I don't want to make that assumption - I want us to explicitly use the rsyslog array-of-strings format.

> 
> I expected single quotes are accepted as well in rsyslog as noted here:
> https://www.rsyslog.com/doc/master/rainerscript/constant_strings.html
> But it is just in rainerscript, not in the other places, like in the action parameters... :(
> 
> This is a snippet of `31-output-elasticsearch-elasticsearch_output_ops.conf` generated by this pr.
> 
> ```
>     if ($.input_type == "ovirt") and ($.index_prefix startswith ".operations.") then {
>         action(
>             type="omelasticsearch"
>             server=['logging-es-ops0', 'logging-es-ops1']
>             serverport="9200"
>             <<snip>>
> ```

I can change my PR to use double quotes if you like - it's just that it was using single quotes before - but in rainierscript, string constants can use either single or double quotes.

nhosoi:
> I can change my PR to use double quotes if you like 

Could you please change to use double quotes? Otherwise, we won't be given the pre-verification...

PR #239, opened 2021-08-06
do not warn about unarchive or leading slashes

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1984182
Use the `args.warn: false` to omit the warning about using
`unarchive` instead of the `tar` command.
Use the `tar` `-P` flag to omit the warning about trimming
leading slashes `/` from the paths.


nhosoi (approved):
lgtm

PR #240, opened 2021-08-07
WIP: [FIX] "Purge original conf" task

richm:
`file` module `state=absent` does not work with `path=/some/dir/*` (wildcard)
`file` module should support `state=empty` see: https://github.com/ansible/ansible/issues/18910
`state=empty` is needed here - and should be used later
Use rm -rfv so that the files being removed will be shown

### DIFF
--- a/roles/rsyslog/handlers/main.yml
+++ b/roles/rsyslog/handlers/main.yml
@@ -3,6 +3,10 @@
   service:
     name: rsyslog
     state: restarted
+  ignore_errors: true
+
+- name: restart rsyslogd
+  command: journalctl -u rsyslog
 
 - name: stop rsyslogd
   service:
--- a/roles/rsyslog/tasks/main.yml
+++ b/roles/rsyslog/tasks/main.yml
@@ -88,9 +88,9 @@
           when: not __backup_stat.stat.exists
 
         - name: Purge original conf
-          file:
-            state: absent
-            path: "{{ __rsyslog_config_dir }}/*"
+          shell: |-
+            set -euo pipefail
+            /bin/rm -rfv {{ __rsyslog_config_dir }}/*
           when: __rsyslog_purge_original_conf | bool | d(false)
 
         - block:


nhosoi (approved):
lgtm

richm:
Well, I guess this doesn't work. I suspect it is because the `$IncludeConfig /etc/rsyslog.d/*.conf` directive has an error if there are no matching files, and the default configuration includes `21-cloudinit.conf` and/or `listen.conf`.  https://www.rsyslog.com/doc/v8-stable/configuration/global/options/rsconf1_includeconfig.html
I suppose if we were to use the `include()` method instead, we could tell it to ignore files that do not exist https://www.rsyslog.com/doc/v8-stable/rainerscript/include.html although it doesn't explicitly say you can use `mode=optional` with a wildcard, but it should be pretty easy.

So, I'm inclined *not* to merge this feature until we can figure out 
- should we remove the default files?
- if so, should we put some sort of placeholder empty file `/etc/rsyslog.d/empty.conf` when we remove everything?
- should we use `include()` (assuming it works as we desire)?  Does it work on el7 and el8?

nhosoi:
> I guess this doesn't work. I suspect it is because the $IncludeConfig /etc/rsyslog.d/*.conf directive has an error if there are no matching files, ...

@richm, I could not reproduce the problem (in a sense...)
Even if rsyslog.d is empty, if `rsyslog.conf` contains some valid configurations, having `$IncludeConfig /etc/rsyslog.d/*.conf` in `rsyslog.conf` does not cause any error. It successfully starts `rsyslogd`.
```
# ls /etc/rsyslog.d
#
# systemctl restart rsyslog
# systemctl status rsyslog
● rsyslog.service - System Logging Service
   Loaded: loaded (/usr/lib/systemd/system/rsyslog.service; enabled; vendor preset: enabled)
   Active: active (running) since Mon 2021-08-09 22:51:43 UTC; 9s ago
  << snip >>
```
However, `rsyslog.conf` _only_ contains `$IncludeConfig /etc/rsyslog.d/*.conf`, `systemctl start rsyslog` fails if `rsyslog.d` is empty. And it is the same with `include(file="/etc/rsyslog.d/*.conf" mode="optional")`. IMHO, I think this is a valid result that `rsyslog` fails to start if there is no configurations. (Please also note that adding `empty.conf` to `/etc/rsyslog.d` does not help. Starting `rsyslogd` still fails in the same manner.)

My test env is CentOS-8 and rsyslog-8.1911.0-7.el8.x86_64 is installed.

I'm wondering what use case we are trying to support...
In this `tests_default` case, we don't touch rsyslog.conf, thus there's no problem even if the existing files in `rsyslog.d` are purged.
```
  tasks:
    - name: default run
      vars:
        logging_purge_confs: true
      include_role:
```
If we configure something with `logging_inputs`, `logging_outputs`, and `logging_flows`, some config files are generated in /etc/rsyslog.d...

That's said, I think your proposal still looks good to me.

richm:
> However, rsyslog.conf only contains $IncludeConfig /etc/rsyslog.d/*.conf, systemctl start rsyslog fails if rsyslog.d is empty.

Then in the purge case, can we add some sort of harmless directive to /etc/rsyslog.conf in addition to the `$IncludeConfig` directive so that rsyslog will start?

> I'm wondering what use case we are trying to support...

I guess it is the case where you have a certain config, and want to use a different config?  e.g. you deploy using a forwarding config initially, then you want to completely erase that and deploy an ovirt elasticsearch config instead?  Can you be sure that when you do this, the forwarding config is completely erased?

nhosoi:
> > However, rsyslog.conf only contains $IncludeConfig /etc/rsyslog.d/*.conf, systemctl start rsyslog fails if rsyslog.d is empty.
> 
> Then in the purge case, can we add some sort of harmless directive to /etc/rsyslog.conf in addition to the `$IncludeConfig` directive so that rsyslog will start?

Rsyslog requires at least one `action` in the config. This is ugly, but having this file, rsyslog starts. I'd think since 1) no input conf contains `call empty_workaround` and 2) `9999-*` comes the last, it is harmless. Shall we add it when `logging_purge_confs: true`?
```
==> 9999-empty-workaround.conf <==
# Ansible managed
ruleset(name="empty_workaround") {
    *.*                                                 /dev/null
}
```
> 
> > I'm wondering what use case we are trying to support...
> 
> I guess it is the case where you have a certain config, and want to use a different config? e.g. you deploy using a forwarding config initially, then you want to completely erase that and deploy an ovirt elasticsearch config instead? Can you be sure that when you do this, the forwarding config is completely erased?

The concern is the forwarding config could be somehow left? That is, `set -euo pipefail` is not enough? We could add one more check if the rsyslog config directory is empty or not. (but I guess it's redundant...)
```
diff --git a/roles/rsyslog/tasks/main.yml b/roles/rsyslog/tasks/main.yml
index 3896daa..e06e8c6 100644
--- a/roles/rsyslog/tasks/main.yml
+++ b/roles/rsyslog/tasks/main.yml
@@ -87,6 +87,7 @@
           shell: |-
             set -euo pipefail
             /bin/rm -rfv {{ __rsyslog_config_dir }}/*
+            [[ ! $( ls {{ __rsyslog_config_dir }} 2> /dev/null ) ]]
           when: __rsyslog_purge_original_conf | bool | d(false)
       vars:
         __rsyslog_backup_dir: '{{ rsyslog_backup_dir |
```

richm:
> > > However, rsyslog.conf only contains $IncludeConfig /etc/rsyslog.d/*.conf, systemctl start rsyslog fails if rsyslog.d is empty.
> > 
> > 
> > Then in the purge case, can we add some sort of harmless directive to /etc/rsyslog.conf in addition to the `$IncludeConfig` directive so that rsyslog will start?
> 
> Rsyslog requires at least one `action` in the config. This is ugly, but having this file, rsyslog starts. I'd think since 1) no input conf contains `call empty_workaround` and 2) `9999-*` comes the last, it is harmless. Shall we add it when `logging_purge_confs: true`?
> 
> ```
> ==> 9999-empty-workaround.conf <==
> # Ansible managed
> ruleset(name="empty_workaround") {
>     *.*                                                 /dev/null
> }
> ```

Ok - good to know.

> 
> > > I'm wondering what use case we are trying to support...
> > 
> > 
> > I guess it is the case where you have a certain config, and want to use a different config? e.g. you deploy using a forwarding config initially, then you want to completely erase that and deploy an ovirt elasticsearch config instead? Can you be sure that when you do this, the forwarding config is completely erased?
> 
> The concern is the forwarding config could be somehow left? That is, `set -euo pipefail` is not enough?

I think `/bin/rm -rfv {{ __rsyslog_config_dir }}/*` is enough (unless we are concerned about .dot files - and I'm not).

My concern was with the code as it stands today - does it leave behind old config?  If you use `state: absent` to remove the old config, and `state: present` to add the config, does it leave any old config behind?

> We could add one more check if the rsyslog config directory is empty or not. (but I guess it's redundant...)
> 
> ```
> diff --git a/roles/rsyslog/tasks/main.yml b/roles/rsyslog/tasks/main.yml
> index 3896daa..e06e8c6 100644
> --- a/roles/rsyslog/tasks/main.yml
> +++ b/roles/rsyslog/tasks/main.yml
> @@ -87,6 +87,7 @@
>            shell: |-
>              set -euo pipefail
>              /bin/rm -rfv {{ __rsyslog_config_dir }}/*
> +            [[ ! $( ls {{ __rsyslog_config_dir }} 2> /dev/null ) ]]
>            when: __rsyslog_purge_original_conf | bool | d(false)
>        vars:
>          __rsyslog_backup_dir: '{{ rsyslog_backup_dir |
> ```

nhosoi:
> > > > I'm wondering what use case we are trying to support...
> > > 
> > > 
> > > I guess it is the case where you have a certain config, and want to use a different config? e.g. you deploy using a forwarding config initially, then you want to completely erase that and deploy an ovirt elasticsearch config instead? Can you be sure that when you do this, the forwarding config is completely erased?
> > 
> > 
> > The concern is the forwarding config could be somehow left? That is, `set -euo pipefail` is not enough?
> 
> I think `/bin/rm -rfv {{ __rsyslog_config_dir }}/*` is enough (unless we are concerned about .dot files - and I'm not).

Good. *whew* :)

> My concern was with the code as it stands today - does it leave behind old config? If you use `state: absent` to remove the old config, and `state: present` to add the config, does it leave any old config behind?

Yes, it's noted here.
https://github.com/linux-system-roles/logging/#update-and-delete
And this is an example of cleaning up. Most test cases in the logging tests have a clean up code with `state: absent`.
https://github.com/linux-system-roles/logging/blob/master/tests/tests_combination.yml#L169

richm:
> Yes, it's noted here.

Then I guess the only use case for purge is when you aren't sure what is the original config, and you just want to wipe everything out and start over from scratch.  So I think this is a valid bug still, but I don't think it is urgent to fix.  The workaround would be for the user to write a simple play that removed everything from `/etc/rsyslog.d/*` except for `listen.conf` and `21-cloudinit.conf`

nhosoi:
Hi Rich, it's a hack, but how about removing files having "# Ansible managed"? Unfortunately, as rulebase cannot have the comment, we have to remove them without checking, though. Maybe, it's better than deleting files blindly?
```
diff --git a/roles/rsyslog/tasks/main.yml b/roles/rsyslog/tasks/main.yml
index 3896daa..1993c3c 100644
--- a/roles/rsyslog/tasks/main.yml
+++ b/roles/rsyslog/tasks/main.yml
@@ -86,7 +86,8 @@
         - name: Purge original conf
           shell: |-
             set -euo pipefail
-            /bin/rm -rfv {{ __rsyslog_config_dir }}/*
+            find {{ __rsyslog_config_dir }} -type f -exec grep -l "^# Ansible managed" {} \; -delete
+            find {{ __rsyslog_config_dir }} -name "*.rulebase" -delete
           when: __rsyslog_purge_original_conf | bool | d(false)
       vars:
         __rsyslog_backup_dir: '{{ rsyslog_backup_dir |
```

richm:
> Hi Rich, it's a hack, but how about removing files having "# Ansible managed"? Unfortunately, as rulebase cannot have the comment, we have to remove them without checking, though. Maybe, it's better than deleting files blindly?
> 
> ```
> diff --git a/roles/rsyslog/tasks/main.yml b/roles/rsyslog/tasks/main.yml
> index 3896daa..1993c3c 100644
> --- a/roles/rsyslog/tasks/main.yml
> +++ b/roles/rsyslog/tasks/main.yml
> @@ -86,7 +86,8 @@
>          - name: Purge original conf
>            shell: |-
>              set -euo pipefail
> -            /bin/rm -rfv {{ __rsyslog_config_dir }}/*
> +            find {{ __rsyslog_config_dir }} -type f -exec grep -l "^# Ansible managed" {} \; -delete
> +            find {{ __rsyslog_config_dir }} -name "*.rulebase" -delete
>            when: __rsyslog_purge_original_conf | bool | d(false)
>        vars:
>          __rsyslog_backup_dir: '{{ rsyslog_backup_dir |
> ```

What about if we just delete everything except 21-cloudinit.conf and listen.conf?

nhosoi:
> What about if we just delete everything except 21-cloudinit.conf and listen.conf?

If there's no more packages added in the future which contain rsyslog conf file(s), I'm happy with it, too. :)

Kariton:
Sorry for my late reply.  
And thanks for keeping track of this issue.

Does it make sense to utilise `rpm -qf` to check if a file is related to any package?

I missed this (sub)issue due to the fact that i use it to clear everything and reapply my configuration.  
For testing is it easier and faster.

Another idear:
`purge_original_conf` deletes hard coded known files from the default package(s?). 
create `purge_conf` / `purge_all_conf`, as new variable,
with clear description and warning that everything inside `/etc/rsyslog.d` will get deleted
and the hint that `rsyslog` would refuse to work if no valid configuration is present.
this would eliminate unwanted or unexpected behavior by reusing the 'old' `purge_original_conf`.

Kariton:
nevertheless: how to handle `purge_original_conf ` with the `$IncludeConfig /etc/rsyslog.d/*.conf` problem.

nhosoi:
[citest]

richm:
@nhosoi do we still need this PR, or is the issue fixed by https://github.com/linux-system-roles/logging/pull/262 ?

nhosoi:
> @nhosoi do we still need this PR, or is the issue fixed by #262 ?

@richm, this pr is still needed. Actually, I'd like to run the test with #262 rebased. If we do `[citest bad]`, does it happen automatically on github or should we manually rebase against master?

richm:
> > @nhosoi do we still need this PR, or is the issue fixed by #262 ?
> 
> @richm, this pr is still needed. Actually, I'd like to run the test with #262 rebased. If we do `[citest bad]`, does it happen automatically on github or should we manually rebase against master?

You have to rebase manually.  I rebased this PR on top of master.

nhosoi:
lgtm (again :)

Thank you, @richm. Now I see "All checks have passed".

richm:
so it looks like - yes, we do manage /etc/rsyslog.conf:
```
    - name: Generate main rsyslog configuration
      template:
        src: 'rsyslog.conf.j2'
        dest: '/etc/rsyslog.conf'
        mode: "{{ __rsyslog_register_stat_conf.stat.mode | d('0700') }}"
      when:
        - __rsyslog_enabled | bool
        - rsyslog_inputs | d([])
      notify: restart rsyslogd
```
so I think purge, backup, and restore should also manage /etc/rsyslog.conf

richm:
It gets even worse :-(   If we implement this fix, if an existing customer sets purge/restore, they will use *their existing non-default config* as the original config, which is definitely not what we desired.

I think the only way to be sure is to grab the exact original config from the RPM.  One way to do that is to purge config, then uninstall the RPM, then reinstall the RPM.  That would work no matter what currently exists.  The only problem is that there may be customers who run this in "disconnected" mode - the machine is not connected to a network, and so has the packages pre-installed or installed at the time of provisioning.  In that case, we would need to query the RPM database to get the original content of the files - I'm not even sure if that is possible.

nhosoi:
@richm, I merged #264 which covers this pr. Is it ok to close this pr?

PR #243, opened 2021-08-19
Allowing the case, tls is false and key/certs vars are configured.

nhosoi:
- The case used to fail in the checking task with emitting an error.
  The error is transformed to a warning.
- For the test case, TEST CASE 2 in tests_files_elasticsearch.yml
  is added.

In addition, some minor bugs in tests_files_elasticsearch.yml are
fixed.

ref: bz1994580

### DIFF
--- a/tests/tests_files_elasticsearch.yml
+++ b/tests/tests_files_elasticsearch.yml
@@ -19,9 +19,6 @@
       specify all 3 of ca_cert, cert, private_key, or all 3 of
       ca_cert_src, cert_src, private_key_src, or set tls:
       false in the configuration named elasticsearch_output"
-    __expected_err2: "Error: you specified tls: false and also
-      specified one or more cert/key files in the configuration
-      named elasticsearch_output"
 
   tasks:
     - name: "local certs are copied to the target host with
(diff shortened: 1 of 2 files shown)


spetrosi (approved):
lgtm

tests/tests_files_elasticsearch.yml
```diff
@@ -297,44 +278,81 @@
...
+      command: >-
+        /bin/grep 'tls.{{ item.key }}="{{ item.value }}"'
+        {{ __test_outputfiles_conf }}
+      with_dict:
+        - cacert: "{{ __test_ca_cert_target }}"
+        - mycert: "{{ __test_cert_target }}"
+        - myprivkey: "{{ __test_key_target }}"
+      register: __result
+      changed_when: false
+      failed_when: __result.rc != 1
```
richm, line 355: do you use this `__result` anywhere else?  If not, then the default failed condition here is if the `grep` in the `command` returns non-zero, so I think you can get rid of the `register` and the `failed_when` here.

tests/tests_files_elasticsearch.yml
```diff
@@ -297,44 +278,81 @@
...
+      stat:
+        path: "{{ item }}"
+      register: __result
+      failed_when: __result.stat.exists
+      loop:
+        - "{{ __test_ca_cert_target }}"
+        - "{{ __test_cert_target }}"
+        - "{{ __test_key_target }}"
+
+    - name: Check key/certs in {{ __test_outputfiles_conf }}
```
richm, line 345: Should this be "Check key/certs not in {{ __test_outputfiles_conf }}" ?

richm (approved):
lgtm

nhosoi:
Thank you, @richm. Merging.

PR #247, opened 2021-09-10
Replacing seport module with the semanage command line.

nhosoi:
In the replacement, a serous redundancy was found in the sub-role
handling. This commit eliminated the extra loop on the ovirt input.
Other inputs and outputs share the same redundancy. They are going
to be cleaned up in a separate commit.

### DIFF
--- a/roles/rsyslog/tasks/inputs/ovirt/main.yml
+++ b/roles/rsyslog/tasks/inputs/ovirt/main.yml
@@ -1,4 +1,18 @@
 ---
+# yamllint disable rule:line-length
+- name: Set platform/version specific variables
+  include_vars: "{{ item }}"
+  loop:
+    - "{{ role_path }}/vars/inputs/ovirt/default.yml"
+    - "{{ role_path }}/vars/inputs/ovirt/{{ ansible_facts['os_family'] }}.yml"
+    - "{{ role_path }}/vars/inputs/ovirt/{{ ansible_facts['distribution'] }}.yml"
+    - "{{ role_path }}/vars/inputs/ovirt/{{ ansible_facts['distribution'] }}_\
+      {{ ansible_facts['distribution_major_version'] }}.yml"
+    - "{{ role_path }}/vars/inputs/ovirt/{{ ansible_facts['distribution'] }}_\
+      {{ ansible_facts['distribution_version'] }}.yml"
+  when: item is file
+
+# yamllint enable rule:line-length
 # Deploy configuration files
 - name: "Install/Update oVirt input packages and generate
   configuration files in /etc/rsyslog.d"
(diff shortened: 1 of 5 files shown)


richm:
> In the replacement, a serous redundancy was found in the sub-role
> handling. This commit eliminated the extra loop on the ovirt input.
> Other inputs and outputs share the same redundancy. They are going
> to be cleaned up in a separate commit.

I think all of this cleanup should be in a separate PR

roles/rsyslog/tasks/inputs/ovirt/main.yml
```diff
@@ -9,30 +9,48 @@
...
+  register: __result1
+  changed_when: false
+
+- name: "Allow rsyslog to listen on the default collectd port if no
+  other port is available and the port is not yet configured"
+  command: >-
+    semanage port -a -t syslogd_port_t -p tcp 44514
+  when:
+    - __result0.skipped
+    - __result1.rc != 0
```
richm, line 54: I think this sequence should be something like this (pseudocode):
```
if semanage port -l lists port {{ item.rsyslog_read_collectd_port | d(44514) }} && type syslogd_port_t && protocol tcp; then
  do nothing - already set
else
  semanage port -a -t syslogd_port_t -p tcp {{ item.rsyslog_read_collectd_port | d(44514) }} 
```
Can `semanage port -a` operate idempotently?  That is - it would be nice if `semanage port -a` would do something like this: return 0 if changed successfully; return 1 if the provided settings are already set; return -1 if there is an error - I don't see anything like that, nor does the man page give any information
Does the `seport` module reload the policy, or does it work like `semanage port --noreload`?
What does `seport` do if port `{{ item.rsyslog_read_collectd_port | d(44514) }}` is already set to some other type other than `syslogd_port_t`?  What does `semanage port -a -t syslogd_port_t -p tcp {{ item.rsyslog_read_collectd_port | d(44514) }}` do?

nhosoi:
Unfortunately, `semanage` is not in `policycoreutils-python-utils` for rhel-/centos-7.
```
TASK [/path/to/linux-system-roles/logging/roles/rsyslog : Install/Update required packages] ***
task path: /path/to/linux-system-roles/logging/roles/rsyslog/tasks/deploy.yml:4
fatal: [/path/to/Downloads/CentOS-7-x86_64-GenericCloud.qcow2c]: FAILED! => {
    "changed": false,
    "rc": 126,
    "results": [
        "No package matching 'policycoreutils-python-utils' found available, installed or updated"
    ]
}
MSG:
No package matching 'policycoreutils-python-utils' found available, installed or updated
```

richm (approved):
lgtm

nhosoi:
Thank you for reviewing this pr, @richm !

PR #248, opened 2021-09-11
Eliminating redundant loop.

nhosoi:
When including an input or an output subrole, it is called with
one configuration item (not with the list). Thus, it should not
loop on the logging_inputs or logging_outputs items in the sub-
role task. This commit removes the loops and updated templates
according to the change.


roles/rsyslog/tasks/inputs/basics/main.yml
```diff
@@ -7,22 +7,17 @@
...
-      - name: "input-basics-{{ item.name }}"
+      - name: "input-basics-{{ _input.name }}"
         type: input
-        state: "{{ item.state | d('present') }}"
+        state: "{{ _input.state | d('present') }}"
         sections:
           - options: "{{ lookup('template', __rsyslog_input_basics) }}"
   include_tasks:
     file: "{{ role_path }}/tasks/deploy.yml"
-  loop: '{{ logging_inputs | sort(attribute="type") }}'
```
richm, line 21: this is looping in sorted order, but the outer loop does not https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/tasks/main.yml#L296 - do you want the outer loop to be sorted?

nhosoi: Thank you, @richm. I totally forgot about the special treatment for the basic input (imjournal), which conf file should be guaranteed there is just one file.

richm (approved):
lgtm

nhosoi:
Thank you for your reviews, @richm.

PR #253, opened 2021-10-07
Support ansible-core 2.11

nhosoi:
Updated .ansible-lint to skip a role name warning in the meta file.

Added "name" to debug tasks to satisfy ansible-lint.


richm:
@nhosoi note that the tox CI here is still using ansible-lint 2.9, which requires the numeric identifiers in the .ansible-lint like `'303'`, not the string identifiers like `role-name`, which are used by ansible-lint 2.11.  In my testing, ansible-lint 2.9 will ignore the string identifiers, and ansible-lint 2.11 will ignore the numeric identifiers, so you can use both in .ansible-lint.

We still have to support both ansible-lint 2.9 and ansible-lint 2.11 for some time.

nhosoi:
> @nhosoi note that the tox CI here is still using ansible-lint 2.9, which requires the numeric identifiers in the .ansible-lint like `'303'`, not the string identifiers like `role-name`, which are used by ansible-lint 2.11. In my testing, ansible-lint 2.9 will ignore the string identifiers, and ansible-lint 2.11 will ignore the numeric identifiers, so you can use both in .ansible-lint.
> 
> We still have to support both ansible-lint 2.9 and ansible-lint 2.11 for some time.

Sorry, I did not know the string identifiers are not available for 2.9... I'm going to replace them with the numerical ones. (there is one issue which may not work well... but it's not in logging, but in vpn. so, i'm switching to the pr.)

nhosoi:
Thank you, @richm. Merging.

PR #262, opened 2022-01-07
Add logging_restore_confs variable to restore backup.

nhosoi:
When `logging_purge_confs` is set to `true` and the configuration is cleaned up at the end of the test (e.g., tests_basics_files.yml), the rsyslog configuration on the target nodes are destroyed and rsyslog service fails to restart as shown in this ansible.log.
https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/linux-system-roles-logging-pull-linux-system-roles_logging-240-618f22e-rhel-8-y-20220105-174527/artifacts/ansible.log

To mitigate the issue, I'd like to propose to add `logging_restore_confs` variable to restore the conf files from backup if it is set to `true`. In the addition, I'm switching the default backup location from /tmp to /var/tmp to keep the backup in a bit safer place.

### DIFF
--- a/roles/rsyslog/tasks/main.yml
+++ b/roles/rsyslog/tasks/main.yml
@@ -74,25 +74,42 @@
             state: directory
             path: '{{ __rsyslog_backup_dir }}'
             mode: '0700'
-          changed_when: false
 
-        - name: Archive the contents of {{ __rsyslog_config_dir }} to
-            the backup dir
+        - name: Get backup file stat
+          stat:
+            path: "{{ __rsyslog_backup_dir }}/backup.tgz"
+          register: __backup_stat
+
+        - name: Archive the original contents of {{ __rsyslog_config_dir }}
+            to the backup dir
           command: >
             tar -cvzPf "{{ __rsyslog_backup_dir }}/backup.tgz"
-                /etc/rsyslog.conf "{{ __rsyslog_config_dir }}"
-          args:
-            warn: false  # suppress 'unarchive' warning
-          changed_when: false
+              /etc/rsyslog.conf "{{ __rsyslog_config_dir }}"
+          when: not __backup_stat.stat.exists
 
         - name: Purge original conf
           file:
             state: absent
             path: "{{ __rsyslog_config_dir }}/*"
           when: __rsyslog_purge_original_conf | bool | d(false)
+
+        - block:
+            - name: Get backup file stat
+              stat:
+                path: "{{ __rsyslog_backup_dir }}/backup.tgz"
+              register: __backup_stat
+
+            - name: Restore original conf
+              command: >
+                tar -xvzPf "{{ __rsyslog_backup_dir }}/backup.tgz"
+              when: __backup_stat.stat.exists
+          when:
+            - __rsyslog_restore_original_conf | bool | d(false)
+
       vars:
-        __rsyslog_backup_dir: '{{ rsyslog_backup_dir |
-          d("/tmp") }}/rsyslog.d-{{ ansible_date_time.iso8601_basic_short }}'
+        __rsyslog_backup_dir: '{{ logging_backup_dir
+                               if logging_backup_dir|length > 0
+                               else "/var/lib/rsyslog" }}'
 
     - name: "Create logging directory if it does not exist or
       the ownership and/or modes are different."
--- a/README.md
+++ b/README.md
@@ -404,6 +404,8 @@ These variables are set in the same level of the `logging_inputs`, `logging_outp
 - `logging_mark`: Mark message periodically by immark, if set to `true`. Default to `false`.
 - `logging_mark_interval`: Interval for `logging_mark` in seconds. Default to `3600`.
 - `logging_purge_confs`: `true` or `false`. If set to `true`, files in /etc/rsyslog.d are purged.
+- `logging_backup_dir`: Path to specify the directory to store backup file which contains pre-logging system role run. The backup is used when `logging_restore_confs` is set to `true`.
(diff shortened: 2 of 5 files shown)


roles/rsyslog/tasks/main.yml
```diff
@@ -90,9 +92,28 @@
             state: absent
             path: "{{ __rsyslog_config_dir }}/*"
           when: __rsyslog_purge_original_conf | bool | d(false)
+
+        - block:
+            - name: Get backup files
+              shell: ls "{{ __rsyslog_backup_dir }}"
```
richm, line 98: use `command` instead of `shell`?

roles/rsyslog/tasks/main.yml
```diff
@@ -78,8 +78,10 @@
 
         - name: Archive the contents of {{ __rsyslog_config_dir }} to
             the backup dir
+          vars:
+            __now: "{{ lookup('pipe', 'date +%s') }}"
           command: >
-            tar -cvzPf "{{ __rsyslog_backup_dir }}/backup.tgz"
+            tar -cvzPf "{{ __rsyslog_backup_dir }}/backup{{ __now }}.tgz"
```
richm, line 84: lookup will return the timestamp on the controller, not on the managed host where the backup will be created.  Is there an ansible fact which contains the current time on the managed host?

roles/rsyslog/tasks/main.yml
```diff
@@ -90,9 +92,28 @@
...
+
+            - name: Get backup file stat
+              stat:
+                path: "{{ __rsyslog_backup_dir }}/{{ __backups.stdout_lines[0] }}"
+              register: __backup_stat
+
+            - name: Restore original conf
+              shell: |-
+                # set -euo pipefail
+                tar -xvzPf "{{ __rsyslog_backup_dir }}/{{ __backups.stdout_lines[0] }}"
```
richm, line 109: use `command` instead of `shell`?

richm: is `__backups.stdout_lines[0]` the earliest file or the latest file?

roles/rsyslog/tasks/main.yml
richm: If there are no backups, what happens?

roles/rsyslog/tasks/main.yml
richm: I guess you can't use `ansible_date_time.iso8601_basic_short` here because the method below relies on `__now` being in seconds?  here are the time fields in ansible_date_time`:
```
    "epoch": "1641649779",
    "hour": "08",
    "iso8601": "2022-01-08T13:49:39Z",
    "iso8601_basic": "20220108T084939145434",
    "iso8601_basic_short": "20220108T084939",
    "iso8601_micro": "2022-01-08T13:49:39.145434Z",
```
I think you could use `iso8601_basic_short` here - it gives the same second resolution as `date +%s`, and I think it sorts/collates the same way

roles/rsyslog/tasks/main.yml
nhosoi: > use `command` instead of `shell`?

:p Thanks!

> If there are no backups, what happens?

The backup dir is guaranteed to exist.
https://github.com/linux-system-roles/logging/blob/master/roles/rsyslog/tasks/main.yml#L73-L76

And if no backup files exist, extracting from the backup file is skipped.
https://github.com/nhosoi/logging/blob/restore_confs/roles/rsyslog/tasks/main.yml#L110

roles/rsyslog/tasks/main.yml
nhosoi: > is `__backups.stdout_lines[0]` the earliest file or the latest file?

The earliest. I was hoping to restore from the one prior to running the first rhel-system-roles.logging. That is, the one from the original rsyslog rpm, if available.

Rethinking of it, do we want to have two variables `logging_restore_confs_from_prev` and `logging_restore_confs_from_first` instead of one?

richm: > > is `__backups.stdout_lines[0]` the earliest file or the latest file?
> 
> The earliest. I was hoping to restore from the one prior to running the first rhel-system-roles.logging. That is, the one from the original rsyslog rpm, if available.

So we really only need 1 backup - something like
```
if not backup exists in backup_dir; then
  create backup in backup_dir
fi
```
The problem with creating an indefinite number of backups is that there is no way to clean them up.

> 
> Rethinking of it, do we want to have two variables `logging_restore_confs_from_prev` and `logging_restore_confs_from_first` instead of one?

No.  I think just "restore_confs" meaning "restore to original state" is sufficient.

roles/rsyslog/tasks/main.yml
```diff
@@ -76,16 +76,18 @@
...
+        - name: Archive the original contents of {{ __rsyslog_config_dir }}
+            to the backup dir
           command: >
-            tar -cvzPf "{{ __rsyslog_backup_dir }}/backup{{ __now }}.tgz"
-                /etc/rsyslog.conf "{{ __rsyslog_config_dir }}"
-          args:
-            warn: false  # suppress 'unarchive' warning
+            tar -cvzPf "{{ __rsyslog_backup_dir }}/backup.tgz"
+              /etc/rsyslog.conf "{{ __rsyslog_config_dir }}"
           changed_when: false
```
richm, line 89: do you still need the `changed_when` here?

nhosoi: If we consider creating a backup file is a `change`, we don't need it. And without the statement, it doesn't break the idempotence. So, right. it's not needed any more. Removing it. Thanks!

richm (approved):
lgtm

nhosoi:
Thank you for reviewing the pr, @richm. Squashed and merged.

PR #264, opened 2022-01-13
Refactor logging_purge_confs and logging_restore_confs.

nhosoi:
New Variables - [0]
- logging_purge_confs: true or false.
      If logging_purge_confs is set to true, remove files in rsyslog.d
      which do not belong to any rpm packages. That includes config files
      generated by the previous logging role run.
      Default to false.
- logging_reset_confs: true or false.
      If logging_reset_confs is set to true, reinstall rsyslog package,
      and restore the pre-existing config files, which restores the
      default /etc/rsyslog.conf.
      Default to false.
    
Other changes in the refactor:
- Variable logging_restore_confs is replaced with logging_reset_confs.
- A test case is added to tests_basics_forwards TEST CASE 2, where
      the even if the config files installed by other packages are
      accidentally removed, it's verified that the logging system
      (rsyslog) can be reset and restarted with "logging_reset_confs:
      true" in the cleanup playbook.
- A test case only sets purge and reset is added.

Note: This PR takes over https://github.com/linux-system-roles/logging/pull/240.

[0] - Thanks to @richm and Brian Smith for their ideas. This refactoring is based on them.


roles/rsyslog/tasks/main.yml
```diff
@@ -66,46 +66,72 @@
...
-            to the backup dir
-          command: >
-            tar -cvzPf "{{ __rsyslog_backup_dir }}/backup.tgz"
-              /etc/rsyslog.conf "{{ __rsyslog_config_dir }}"
-          when: not __backup_stat.stat.exists
+        - block:
+            - name: Purge original conf
+              file:
+                state: absent
+                path: "{{ __rsyslog_config_dir }}/*"
```
spetrosi, line 104: Isn't this approach described as not working in #240?

README.md
```diff
@@ -404,8 +404,8 @@ These variables are set in the same level of the `logging_inputs`, `logging_outp
 - `logging_mark`: Mark message periodically by immark, if set to `true`. Default to `false`.
 - `logging_mark_interval`: Interval for `logging_mark` in seconds. Default to `3600`.
 - `logging_purge_confs`: `true` or `false`. If set to `true`, files in /etc/rsyslog.d are purged.
-- `logging_backup_dir`: Path to specify the directory to store backup file which contains pre-logging system role run. The backup is used when `logging_restore_confs` is set to `true`.
-- `logging_restore_confs`: `true` or `false`. If set to `true`, config files in the backup file are restored. This variable is to be used in the cleaning up.
+- `logging_backup_dir`: Path to specify the directory to store backup file which contains the config files installed by the other packages. Default to `/var/lib/rsyslog` for `rsyslog`.
```
spetrosi, line 407: What other packages?

roles/rsyslog/tasks/main.yml
```diff
@@ -66,46 +66,72 @@
         path: '{{ __rsyslog_work_dir }}'
         mode: '0700'
 
-    # Back up the pre-existing rsyslog config files in the backup dir,
-    # then removing the files/dirs in rsyslog.d.
+    # Back up the pre-existing rsyslog config files in the backup dir.
+    # If logging_purge_confs is set to true, remove files in rsyslog.d.
+    # Then restore the pre-existing config files.
+    # If logging_reset_confs is set to true, reinstall rsyslog package,
```
spetrosi, line 72: Where are lines that define what happens when logging_reset_confs == true?

README.md
nhosoi: Could you please elaborate your question?

If it helps, here's what's happening then.
Let's supposed `/etc/rsyslog.d` has these config files.
```
# ls /etc/rsyslog.d
00-global.conf                21-cloudinit.conf                   30-output-forwards-forwards_facility_only.conf
05-common-defaults.conf       30-output-files-files_output0.conf  30-output-forwards-forwards_severity_and_facility.conf
10-output-files-modules.conf  30-output-files-files_output1.conf  90-input-basics-basic_input0.conf
```
All files except `21-cloudinit.conf` do not belong to any rpm packages since they were generated by the logging role. The backup file in `logging_backup_dir` would contain just `21-cloudinit.conf` in this example and the file would be restored when purge is correctly executed (#240) or when `logging_reset_confs` is set to `true`.

roles/rsyslog/tasks/main.yml
nhosoi: https://github.com/nhosoi/logging/blob/restore_confs/roles/rsyslog/tasks/main.yml#L112-L134

roles/rsyslog/tasks/main.yml
spetrosi: In #240, @richm says _"file module state=absent does not work with path=/some/dir/* (wildcard)"_. `And changes file: state: absent` to `shell: /bin/rm ...` in 
https://github.com/linux-system-roles/logging/pull/240/files#diff-a83292ff37790599fb1a73e2e3a0f69a0f699fdfa00d9af6f57a7b039ea1ddf1L91-L93
I wonder if `path: "{{ __rsyslog_config_dir }}/*"` works in your case.

README.md
spetrosi: I am nitpicking the docs :) If you could specify exactly what packages override config files it would help. But it's up to you, if users does not need to know it, it is fine to keep this as is.

roles/rsyslog/tasks/main.yml
spetrosi: Ah sorry I am blind, that is fine

README.md
richm: This is how you find out which packages have files installed in `/etc/rsyslog.d`
```
# rpm -qf /etc/rsyslog.d
cloud-init-20.3-10.el8_4.2.noarch
rsyslog-8.1911.0-7.el8.x86_64
```
This is how you find out which files are owned by those packages in `/etc/rsyslog.d`
```
# for pkg in $(rpm -qf /etc/rsyslog.d); do echo "$pkg"; rpm -ql "$pkg" | grep \^/etc/rsyslog.d; done
cloud-init-20.3-10.el8_4.2.noarch
/etc/rsyslog.d
/etc/rsyslog.d/21-cloudinit.conf
rsyslog-8.1911.0-7.el8.x86_64
/etc/rsyslog.d
```

richm: There may be more - we can't assume that only cloud-init provides files in /etc/rsyslog.d - on some systems there is a listen.conf

richm: so the correct way to do this would be:
* rm -rf /etc/rsyslog.conf /etc/rsyslog.d/*
* find packages that provide files in /etc/rsyslog.d
* remove and reinstall those packages

nhosoi: > so the correct way to do this would be:

At what timing, should these steps be executed? In this pr, it is done when `logging_reset_confs` is set to true and the code exists prior to the primary rsyslog role. Basically, it is "independent" from `logging_purge_confs`.

>     * rm -rf /etc/rsyslog.conf /etc/rsyslog.d/*
> 
>     * find packages that provide files in /etc/rsyslog.d
> 
>     * remove and reinstall those packages

In this pr, "find config files that belong to rpm packages except rsyslog and the config files are stored in `backup_additional.tgz`" at https://github.com/nhosoi/logging/blob/restore_confs/roles/rsyslog/tasks/main.yml#L81-L93. It's always done prior to the logging/rsyslog role execution. (This may be better to be done only when `backup_additional.tgz` does not exist.)

If `logging_reset_confs` is set to `true` and `backup_additional.tgz` exists, `rsyslog` is uninstalled and reinstalled, then the config files from `backup_additional.tgz` are extracted and placed in `/etc/rsyslog.d`. Please note this process does not clean up config files, e.g., config files generated by the previous logging role run. I guess you suggest to remove all of them by `rm -rf /etc/rsyslog.conf /etc/rsyslog.d/*`, right? > @richm. If yes, I'm updating this pr. Thanks.

richm: creating the backup.tgz if it does not exist will work fine only if you can guarantee that /etc/rsyslog.conf and the contents of the files in /etc/rsyslog.d are from a brand new system and have never been altered - also - what happens if you do the following:
```
install rsyslog
run logging role - creates backup first time
install cloud-init - installs /etc/rsyslog.d/21-cloudinit.conf
run logging role with purge/restore set - removes /etc/rsyslog.d/21-cloudinit.conf and restores from backup without /etc/rsyslog.d/21-cloudinit.conf
```
so you have a situation where a package is missing a file

nhosoi:
@richm, could you please review once again? I think I implemented based on the ideas brought up in the brainstorming.

roles/rsyslog/handlers/main.yml
```diff
@@ -3,6 +3,10 @@
   service:
     name: rsyslog
     state: restarted
+  ignore_errors: true
+
+- name: restart rsyslogd
+  command: journalctl -u rsyslog
```
richm, line 9: please get rid of this testing code

roles/rsyslog/tasks/main.yml
```diff
@@ -66,50 +66,45 @@
...
+    # and restore the pre-existing config files, which restores the
+    # default /etc/rsyslog.conf.
+    - name: >-
+        Purge original conf - remove confs with no owner package and
+        confs generated by previous logging role.
+      shell: |-
+        set -euo pipefail
+         for conf in $( ls "{{ __rsyslog_config_dir }}" ); do
+           rstr=$( rpm -qf "{{ __rsyslog_config_dir }}"/$conf 2>&1 || : )
```
richm, line 81: `"{{ __rsyslog_config_dir }}/$conf"`

roles/rsyslog/tasks/main.yml
```diff
@@ -66,50 +66,45 @@
...
+           if [[ $rstr == *"not owned by any package"* ]]; then
+             # confs generated by the logging role do not belong to
+             # any rpm packages.
+             /usr/bin/rm -f "{{ __rsyslog_config_dir }}"/$conf
```
richm, line 85: `"{{ __rsyslog_config_dir }}/$conf"`

defaults/main.yml
```diff
@@ -14,17 +14,15 @@ logging_inputs: []
...
-# set to true.
-# Default to '/var/lib/rsyslog' for rsyslog.
-logging_backup_dir: ""
-
-# If logging_restore_confs is set to true, backed up config files are restored.
-logging_restore_confs: false
+# If `logging_reset_confs` is set to `true`, reinstall rsyslog package, and
+# restore the pre-existing config files, which restores the default
+# `/etc/rsyslog.conf`.
+logging_reset_confs: false
```
richm, line 25: This represents a breaking API change - we are removing an existing variable.  Instead, I think we should reuse the variable `logging_restore_confs` and just change the way it makes the role behave.

nhosoi: `logging_restore_confs` was added in this commit (tag: 1.7.0)
```
commit 522da0a0a504d50f7f4397ac36a038bdac390c0d (tag: 1.7.0, origin/master, origin/HEAD, master)
Author: Noriko Hosoi <nhosoi@redhat.com>
Date:   Thu Jan 6 15:33:10 2022 -0800
    Add logging_restore_confs variable to restore backup.
```
The tagged logging source has not been used in any linux-system-roles build yet.
This is the snippet of the rawhide spec file.
```
#%%defcommit 10 20dd3e5520ca06dcccaa9b3f1fb428d055e0c23f
%global rolename10 logging
%deftag 10 1.6.1
```
So, I think it's safe to rename the variable name. (And `logging_reset_confs` represents the behavior better... :)

richm (approved):
lgtm

nhosoi:
Thank you, @richm. Merged.

PR #265, opened 2022-01-21
make purge and reset idempotent

richm:
The use of `logging_purge_confs` and `logging_reset_confs` should
be idempotent.  For `logging_purge_confs`, this is done by first
getting a list of all unowned config files to see if there are
any.  For `logging_reset_confs`, this is done by checking if
`/etc/rsyslog.conf` has been modified.


richm:
[citest]

richm:
[citest]

tests/tests_purge_reset.yml
```diff
@@ -1,11 +1,86 @@
...
+    - name: create some unowned config files
+      copy:
+        dest: "{{ item }}"
+        content: "# this is file {{ item }}"
+        mode: "0444"
+      loop:
+        - "/etc/rsyslog.d/unowned1.conf"
+        - "/etc/rsyslog.d/unowned2.conf"
+
+    - name: modify /etc/rsyslog.conf
```
nhosoi, line 23: I think this task is not needed since the following task "configure with basics + purge + reset" modifies rsyslog.conf anyway. Actually, the effect of this task is overridden by the next task.

tests/tests_purge_reset.yml
```diff
@@ -1,11 +1,86 @@
...
+        rsyslog_outputs: []
+        rsyslog_inputs: []
+        rsyslog_custom_config_files: []
+        __rsyslog_common_rules: []
+    # HACK HACK HACK
+
+    - name: run again with purge + reset to erase all config
       vars:
         logging_purge_confs: true
         logging_reset_confs: true
```
nhosoi, line 86: Actually, eliminating the "HACK - reset global variables" task and add the infamous `logging_enabled: false` to this task does the trick. ;)
```
     - name: run again with purge + reset to erase all config
       vars:
+        logging_enabled: false
         logging_purge_confs: true
         logging_reset_confs: true
       include_role:
```

tests/tests_purge_reset.yml
richm: But the code won't hit https://github.com/linux-system-roles/logging/pull/265/files#diff-a83292ff37790599fb1a73e2e3a0f69a0f699fdfa00d9af6f57a7b039ea1ddf1R115 and report that rsyslog.conf is modified unless I somehow modify rsyslog.conf first?

tests/tests_purge_reset.yml
richm: hmm - ok - note that this is only necessary because we ran the role with actual settings previously in the same process - if a user wants to use reset to revert back to the system defaults, they will have to do it in a separate run of ansible-playbook, or those global variables will still be set (or use logging_enabled: false)

richm: I have another branch which gets rid of the global variables - but I think that change is too invasive for the purge/reset issue

tests/tests_purge_reset.yml
nhosoi: You are right. This "modify /etc/rsyslog.conf" is needed to execute the block. Sorry for the noise.

tests/tests_purge_reset.yml
nhosoi: > hmm - ok - note that this is only necessary because we ran the role with actual settings previously in the same process - if a user wants to use reset to revert back to the system defaults, they will have to do it in a separate run of ansible-playbook, or those global variables will still be set (or use logging_enabled: false)

I agree. The logging role tests try to revert the config files as much as possible in "END TEST CASE #; Clean up the deployed config" to allow to run multiple tests repeatedly in one VM, in which the same global variables are given with `state: absent` to delete the previously generated files.  (Right. `/etc/rsyslog.conf` is not reverted. So, the configuration is almost empty and useless as `rsyslog`.) The cleanup code could be replaced with:
```
       vars:
+        logging_enabled: false
         logging_purge_confs: true
         logging_reset_confs: true
```
Having the simple(?) reset method would be useful for our CI which repeats a series of tests. (Not sure how it is in the real use cases, though...)

richm:
[citest]

richm:
[citest bad]

nhosoi (approved):
lgtm
Thanks!

richm:
[citest pending]

richm:
[citest]

richm:
@nhosoi @spetrosi please review the latest commit

nhosoi:
> @nhosoi @spetrosi please review the latest commit

lgtm

I'd like to use `logging_enabled: false` and `logging_purge_confs: true` in all the END TEST CASE tasks like this. 
```
diff --git a/tests/tests_basics_files.yml b/tests/tests_basics_files.yml
index ac04a28..f64c96b 100644
--- a/tests/tests_basics_files.yml
+++ b/tests/tests_basics_files.yml
@@ -77,14 +77,7 @@
     - name: END TEST CASE 0; Clean up the deployed config
       vars:
         logging_enabled: false
-        logging_inputs:
-          - name: basic_input
-            type: basics
-            state: absent
-        logging_outputs:
-          - name: default_files
-            type: files
-            state: absent
+        logging_purge_confs: true
       include_role:
```
Also, I'd like to update this README sentence followed by an example to use `logging_purge_confs`.
```
If you want to remove all the configuration files previously configured, in addition to setting `state: absent` to each logging_inputs and logging_outputs item, add `logging_enabled: false` to the configuration variables as follows. It will eliminate the global and common configuration files, as well.
```
Had we better make the changes in a separate pr?

richm:
> > @nhosoi @spetrosi please review the latest commit
> 
> lgtm
> 
> I'd like to use `logging_enabled: false` and `logging_purge_confs: true` in all the END TEST CASE tasks like this.

1) I'd like to get rid of `logging_enabled` - I don't understand why it is needed - imo if you use
```
        logging_inputs:
          - name: basic_input
            type: basics
            state: absent
```
it should simply remove that config - how is this different if you use `logging_enabled: true` vs. `logging_enabled: false`?
but removing `logging_enabled` will have to wait for another PR

2) Maybe some of the test cases could use logging_purge_confs, but we still need to test the `state: absent` functionality somehow.

> 
> ```
> diff --git a/tests/tests_basics_files.yml b/tests/tests_basics_files.yml
> index ac04a28..f64c96b 100644
> --- a/tests/tests_basics_files.yml
> +++ b/tests/tests_basics_files.yml
> @@ -77,14 +77,7 @@
>      - name: END TEST CASE 0; Clean up the deployed config
>        vars:
>          logging_enabled: false
> -        logging_inputs:
> -          - name: basic_input
> -            type: basics
> -            state: absent
> -        logging_outputs:
> -          - name: default_files
> -            type: files
> -            state: absent
> +        logging_purge_confs: true
>        include_role:
>          name: linux-system-roles.logging
> ```
> 
> Also, I'd like to update this README sentence followed by an example to use `logging_purge_confs`.
> 
> ```
> If you want to remove all the configuration files previously configured, in addition to setting `state: absent` to each logging_inputs and logging_outputs item, add `logging_enabled: false` to the configuration variables as follows. It will eliminate the global and common configuration files, as well.
> ```
> 
> Had we better make the changes in a separate pr?

I think I need to at least fix the README in this PR.

richm:
I think this current fix is also incomplete, because it doesn't handle all of the cases in a way such that are idempotent and work in check mode.  Consider the following case:
run 1
```
logging_inputs: ... some inputs which will cause edits to /etc/rsyslog.conf
logging_outputs: ... which will create some files in /etc/rsyslog.d/
logging_flows: ...
```
run 2
```
logging_purge_confs: true
logging_inputs: ... other inputs which will cause edits to /etc/rsyslog.conf
logging_outputs: ... other which will create some files in /etc/rsyslog.d/
logging_flows: ...
```
This should a) report changed: true b) refresh /etc/rsyslog.conf by reinstalling c) remove all files in /etc/rsyslog.d not owned by any package

now for the fun part:
run 2
```
logging_purge_confs: true
logging_inputs: ... other inputs which will cause edits to /etc/rsyslog.conf
logging_outputs: ... other which will create some files in /etc/rsyslog.d/
logging_flows: ...
```
This should report changed: false - because we *did not* change the state of the system - we gave the exact same inputs, as previously, so the role should report that the given inputs leave the system in exactly the same state as it is currently in

For /etc/rsyslog.conf, it's pretty simple - if purge is true and no inputs are given, we reinstall to reset /etc/rsyslog.conf back to default state (we currently also have to take logging_enabled into account).
For /etc/rsyslog.d/, we have to apply the templates first, then get a list of files that were touched by those templates - call this template_list.  We then have to get a list of the files in /etc/rsyslog.d/ that are 1) not owned by any package 2) not in the list of files added/changed - call this other_list.  If 1) we are purging 2) template_list has no changes 3) other_list is empty - then the role should report changed: false.  If other_list is not empty, then the role should report changed: true.  If other_list is empty, but template_list has changes, then the role should report changed: true.  This way you could use the role for drift checking - you could just always use
```
logging_purge_confs: true
logging_inputs: ... other inputs which will cause edits to /etc/rsyslog.conf
logging_outputs: ... other which will create some files in /etc/rsyslog.d/
logging_flows: ...
```
if someone went into the system and modified /etc/rsyslog.conf, or added/modified/removed a file in /etc/rsyslog.d/, then ansible would report that something changed, and would fix it. Otherwise, ansible would report no changes, and you could be assured that no one has tampered with the system.

nhosoi:
> now for the fun part:
> run 3
> logging_purge_confs: true
> logging_inputs: ... other inputs which will cause edits to /etc/rsyslog.conf
> logging_outputs: ... other which will create some files in /etc/rsyslog.d/
> logging_flows: ...
> 
> This should report changed: false - because we did not change the state of the system - we gave the exact same inputs, as previously, so the role should report that the given inputs leave the system in exactly the same state as it is currently in

Understand the expectation, but this run 3 is actually a merge of two playbooks run 3' and run 4', isn't it...? For instance, it might be meaningful if I get `changed: true` even if run 2 and run 4' are identical to learn run 4' is executed in the clean state...
run 3'
logging_purge_confs: true

run 4'
logging_inputs: ... other inputs which will cause edits to /etc/rsyslog.conf
logging_outputs: ... other which will create some files in /etc/rsyslog.d/
logging_flows: ...

richm:
> > now for the fun part:
> > run 3
> > logging_purge_confs: true
> > logging_inputs: ... other inputs which will cause edits to /etc/rsyslog.conf
> > logging_outputs: ... other which will create some files in /etc/rsyslog.d/
> > logging_flows: ...
> > This should report changed: false - because we did not change the state of the system - we gave the exact same inputs, as previously, so the role should report that the given inputs leave the system in exactly the same state as it is currently in
> 
> Understand the expectation, but this run 3 is actually a merge of two playbooks run 3' and run 4', isn't it...?For instance, it might be meaningful if I get `changed: true` even if run 2 and run 4' are identical to learn run 4' is executed in the clean state... run 3' logging_purge_confs: true
> 
> run 4' logging_inputs: ... other inputs which will cause edits to /etc/rsyslog.conf logging_outputs: ... other which will create some files in /etc/rsyslog.d/ logging_flows: ...

The user shouldn't have to run two playbooks 3' and 4'.  Also, just re-running 4' does not tell you if someone added an extra file in /etc/rsyslog.d/

richm:
[citest]

nhosoi:
lgtm to all the commits including the last note.

richm:
[citest bad]

PR #273, opened 2022-03-31
Fix failures reported by "tox -e qemu_option -- --remove-cloud-init ..."

nhosoi:
If no additional config files such as 21-cloudinit.conf exist in
/etc/rsyslog.d; in the cleaning up phase in the test playbooks,
by removing all the generated files by the logging role, it ends
up /etc/rsyslog.d is empty with "$IncludeConfig /etc/rsyslog.d/*.conf"
in /etc/rsyslog.conf. It makes the following rsyslog service restart
fail.

This patch replaces the old cleaning up method - removing the
files in /etc/rsyslog.d with the better one - resetting the
rsyslog config files to the ones from rpm.

### DIFF
--- a/tests/tests_basics_files.yml
+++ b/tests/tests_basics_files.yml
@@ -76,15 +76,7 @@
 
     - name: END TEST CASE 0; Clean up the deployed config
       vars:
-        logging_enabled: false
-        logging_inputs:
-          - name: basic_input
-            type: basics
-            state: absent
-        logging_outputs:
-          - name: default_files
-            type: files
-            state: absent
+        logging_purge_confs: true
       include_role:
         name: linux-system-roles.logging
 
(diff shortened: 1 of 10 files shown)


nhosoi:
> ok - but there should be at least one test that tests `logging_enabled: false` and configs with `state: absent` since that is a feature supported by the role

It turned out there is one left in tests_relp.yml. The test playbook did not fail since it contains only one test case and there is no service restarted after the cleaning up... :p If restarting the service, it'd have failed there... Do you think it's ok as it is or shall I make a trick like adding the restart service code plus a dummy config file just not to cause the failure?

richm:
> > ok - but there should be at least one test that tests `logging_enabled: false` and configs with `state: absent` since that is a feature supported by the role
> 
> It turned out there is one left in tests_relp.yml. The test playbook did not fail since it contains only one test case and there is no service restarted after the cleaning up... :p If restarting the service, it'd have failed there... Do you think it's ok as it is or shall I make a trick like adding the restart service code plus a dummy config file just not to cause the failure?

I think it is ok as-is

nhosoi:
[citest bad]

richm:
so the only change is to the tests?  is there any way for a regular user to end up in a situation where /etc/rsyslog.conf contains `$IncludeConfig /etc/rsyslog.d/*.conf` but `/etc/rsyslog.d` is empty?

nhosoi:
> so the only change is to the tests? is there any way for a regular user to end up in a situation where /etc/rsyslog.conf contains `$IncludeConfig /etc/rsyslog.d/*.conf` but `/etc/rsyslog.d` is empty?

I cannot think of any... (of course, unless explicitly setting `state: absent` from the beginning, though...)

And even if it happens, there is a way to workaround the problem, isn't there? By setting `logging_purge_confs: true`, the rsyslog service would be recovered. (I think having the resetting method is quite useful for the system roles, in general.)

richm (approved):
lgtm

nhosoi:
Thanks, @richm. Merged.

PR #274, opened 2022-04-18
support gather_facts: false; support setup-snapshot.yml

nhosoi:
Some users use `gather_facts: false` in their playbooks.  This changes
the role to work in that case, by gathering only the facts it requires
to run.
CI testing can be sped up by creating a snapshot image pre-installed
with packages.  tests/setup-snapshot.yml can be used by a CI system
to do this.

Logging role tests used to check installed packages are necessary and
sufficient. Also, they uninstalled the installed packages at the end
of each test. They are eliminated since it'd get broken when the pre-
installed snapshot image is used.


tests/setup-snapshot.yml
```diff
@@ -0,0 +1,32 @@
+- hosts: all
+  tasks:
+    - name: Set platform/version specific variables
+      include_role:
+        name: linux-system-roles.logging
+        tasks_from: "../roles/rsyslog/tasks/set_vars.yml"
```
richm, line 6: I don't think relative paths like this will work for collections

tests/tests_include_vars_from_parent.yml
```diff
@@ -1,5 +1,15 @@
...
   tasks:
+    - name: Ensure ansible_facts used by role
+      setup:
+        gather_subset:
+          - '!all'
+          - '!any'
+          - os_family
+          - distribution
+          - distribution_version
+          - distribution_major_version
```
richm, line 11: ?  https://docs.ansible.com/ansible/latest/collections/ansible/builtin/setup_module.html#setup-module - I guess `!all` and `!any` should be under `gather_subset`  - but the names of individual facts should be in `filter` - and why is this needed here?

tests/setup-snapshot.yml
richm: What if the rsyslog subrole had its own setup-snapshot.yml, and  each input and output had their own setup-snapshot.yml?

richm:
> Some users use `gather_facts: false` in their playbooks. This changes the role to work in that case, by gathering only the facts it requires to run. CI testing can be sped up by creating a snapshot image pre-installed with packages. tests/setup-snapshot.yml can be used by a CI system to do this.
> 
> Logging role tests used to check installed packages are necessary and sufficient.

Yeah - I don't think this is necessary to do.  I don't like having the hard coded list of packages anyway.

> Also, they uninstalled the installed packages at the end of each test. They are eliminated since it'd get broken when the pre- installed snapshot image is used.

I think we should leave the uninstall.  It won't "break" the pre-installed snapshot image - the packages will still be in the local cache and will be re-installed from the local cache.  In fact other roles do something similar - they uninstall some packages after some tests - this works even when using the pre-installed snapshot image.

tests/tests_include_vars_from_parent.yml
nhosoi: Sorry, I thought the test failed without this task, but I cannot reproduce it any more... There should have been some other bug, then...   I'm removing this setup task.

tests/setup-snapshot.yml
nhosoi: > What if the rsyslog subrole had its own setup-snapshot.yml, and each input and output had their own setup-snapshot.yml?

I'm running my test like this.
```
tox -e qemu-ansible-core-2.12 -- --image-name centos-8 --erase-old-snapshot \
  --setup-yml tests/setup-snapshot.yml --use-snapshot --use-yum-cache --log-level info \
  tests/tests_test_name.yml
```
The rsyslog subrole has 5 inputs and 5 outputs. Instead of having one setup-snapshot.yml, are you suggesting to generate 5 setup-snapshot_input-somename.yml and 5 setup-snapshot_output-somename.yml. And when we run the command line, we list all the setup-snapshot in front something like this? (The setup-snapshot inputs and outputs are the ones used in the test_name playbook.) Maybe, I misunderstood your suggestion...
```
tox -e qemu-ansible-core-2.12 -- --image-name centos-8 --erase-old-snapshot \
  --use-snapshot --use-yum-cache --log-level info \
  tests/setup-snapshot_input-somename0.yml ... tests/setup-snapshot_input-somenameM.yml \
  tests/setup-snapshot_output-somename0.yml ... tests/setup-snapshot_output-somenameN.yml \
  tests/tests_test_name.yml
```

richm: No.  I guess the logging/tests/setup-snapshot.yml should handle that.  I just don't like the fact that if we add another input or output, we'll have to update several places in the code.  I would like to minimize that as much as possible.

tasks/set_vars.yml
```diff
@@ -0,0 +1,22 @@
+---
+- name: Set Rsyslog facts then include rsyslog role
+  block:
+    - name: test
+      include_tasks:
+        file: "{{ role_path }}/roles/rsyslog/tasks/set_vars.yml"
```
nhosoi, line 6: This does not work either for the collections format, of course... :(
```
$ls .tox/ansible_collections/fedora/linux_system_roles/roles
logging/  private_logging_subrole_rsyslog/
```

roles/rsyslog/tasks/set_vars.yml
```diff
@@ -16,3 +16,19 @@
...
+    - "{{ role_path }}/vars/inputs/basics/main.yml"
+    - "{{ role_path }}/vars/inputs/files/main.yml"
+    - "{{ role_path }}/vars/inputs/ovirt/main.yml"
+    - "{{ role_path }}/vars/inputs/relp/main.yml"
+    - "{{ role_path }}/vars/inputs/remote/main.yml"
+    - "{{ role_path }}/vars/outputs/elasticsearch/main.yml"
+    - "{{ role_path }}/vars/outputs/files/main.yml"
+    - "{{ role_path }}/vars/outputs/forwards/main.yml"
+    - "{{ role_path }}/vars/outputs/relp/main.yml"
+    - "{{ role_path }}/vars/outputs/remote_files/main.yml"
```
richm, line 33: You could use lookup fileglob here like this:
```yaml
- name: Include rsyslog vars files
  include_vars: "{{ item }}"
  loop: "{{ ['{{ myrole_path }}/vars/main.yml'] +
            lookup('lines', '/bin/ls -1 {{ myrole_path }}/vars/*/*/main.yml').split(',') }}"
```
then, you could install the packages defined in the vars from these files like this:
```
- name: Install rsyslog packages for tests
  package:
    name: "{{ lookup('vars', *lookup('varnames', '^__rsyslog_.*packages$', wantlist=True)) | flatten | unique | sort }}"
    state: present
```
The `varnames` lookup files all vars matching `^__rsyslog_.*packages$` which is the pattern used to define the packages in the rsyslog/vars/*/*/main.yml files.  Then, the `vars` lookup gets the values of those named variables.

nhosoi: Thank you, @richm! Your code is much cleaner. I replaced the hardcoded version with your suggestion.

roles/rsyslog/tasks/inputs/ovirt/set_vars.yml
```diff
@@ -0,0 +1,21 @@
+---
+- name: Ensure ansible_facts used by role
+  setup:
+    gather_subset:
+      - 'min'
+      - default_ipv4
```
richm, line 6: You don't need to quote `min` here.  Also, `default_ipv4` should go under `filter` - it isn't a `gather_subset` - https://docs.ansible.com/ansible/latest/collections/ansible/builtin/setup_module.html#setup-module
also, if `default_ipv4` is included with `min`, then you don't need to specify it

tests/setup-snapshot.yml
```diff
@@ -0,0 +1,15 @@
+- hosts: all
+  tasks:
+    - name: Set platform/version specific variables
+      include_role:
+        name: linux-system-roles.logging
+        public: true
```
richm, line 6: don't you need to use `tasks_from:` here?

roles/rsyslog/tasks/inputs/ovirt/set_vars.yml
nhosoi: Is this your suggestion?
```
diff --git a/roles/rsyslog/tasks/inputs/ovirt/set_vars.yml b/roles/rsyslog/tasks/inputs/ovirt/set_vars.yml
index d6bdcfe..686aeb6 100644
--- a/roles/rsyslog/tasks/inputs/ovirt/set_vars.yml
+++ b/roles/rsyslog/tasks/inputs/ovirt/set_vars.yml
@@ -2,7 +2,8 @@
 - name: Ensure ansible_facts used by role
   setup:
     gather_subset:
-      - 'min'
+      - min
+    filter:
       - default_ipv4
   when: not ansible_facts.keys() | list |
     intersect(__logging_required_facts) == __logging_required_facts
```
If so, it makes the ovirt input fail due to missing ansible_default_ipv4...
```
  set $!ipaddr4 = '{{ ansible_default_ipv4.address }}';
```
Please note that `min` does not include `default_ipv4`.

So far, only this param set works for me.
```
  setup:
    gather_subset:
      - min
      - default_ipv4
```

richm: ok - using `default_ipv4` in gather_subset is definitely not documented - looks like you should use `network` to get ipv4 and other network facts.

tests/setup-snapshot.yml
nhosoi: That's what I've been struggling with... 
In the second commit: 03b00f557cc525e54eebb3af3bc6b3d7f210009e, I had tasks_from like this: https://github.com/linux-system-roles/logging/pull/274/commits/03b00f557cc525e54eebb3af3bc6b3d7f210009e#diff-7eb2a312068fa9ec1be5e435acb86b2def3a99f30b6ff3092f42964283bd9fa8.

The top level task has to have some way to point another task in rsyslog, it contained the "relative" path to another task in rsyslog task like this: https://github.com/linux-system-roles/logging/pull/274/commits/03b00f557cc525e54eebb3af3bc6b3d7f210009e#diff-fde1cadb053b9da84a369cc078950e138d3ccbdb9d3c2f6a6aef1e482fad2f92

It obviously breaks the collection converted format... :(  So, I gave it up and did include_role with no options just to get the ansible facts and the package names.

roles/rsyslog/tasks/inputs/ovirt/set_vars.yml
richm: so try
```yaml
 - name: Ensure ansible_facts used by role
   setup:
     gather_subset:
       - min
       - network
   when: not ansible_facts.keys() | list |
     intersect(__logging_required_facts) == __logging_required_facts

File: tests/setup-snapshot.yml (hunk 6 above)
Comment (richm) [reply tests/setup-snapshot.yml:6]: hmm - Then I guess add some sort of logic to logging tasks/main.yml so that when `__snapshot_gather_vars` is `true` the role does nothing except set the facts

File: roles/rsyslog/tasks/inputs/ovirt/set_vars.yml (hunk 5 above)
Comment (nhosoi) [reply roles/rsyslog/tasks/inputs/ovirt/set_vars.yml:6]: Hmmm, I thought we want to gather facts as less as possible. So, I put `default_ipv4` instead of `network`. I guess my assumption was wrong?

File: tests/setup-snapshot.yml (hunk 6 above)
Comment (nhosoi) [reply tests/setup-snapshot.yml:6]: If the performance is the issue, I compared the duration time between the 2nd commit version https://github.com/linux-system-roles/logging/commit/03b00f557cc525e54eebb3af3bc6b3d7f210009e#diff-7eb2a312068fa9ec1be5e435acb86b2def3a99f30b6ff3092f42964283bd9fa8 and this one. There were almost no difference... But probably, we need some way to avoid the risk to bring in an unexpected bug... Let me think about it...

File: roles/rsyslog/tasks/inputs/ovirt/set_vars.yml (hunk 5 above)
Comment (richm) [reply roles/rsyslog/tasks/inputs/ovirt/set_vars.yml:6]: > Hmmm, I thought we want to gather facts as less as possible. So, I put `default_ipv4` instead of `network`. I guess my assumption was wrong?

The way it was explained to me by one of the Ansible guys is that gather_subset is the name of a collection of facts - for example, `min` uses these subsets - https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/setup.py#L192-L195 - `network` is another subset.  Specifying the fewest subsets possible will speed up Ansible processing.  You can also use `filter` to reduce the amount of network traffic by specifying only those parameters you want to be returned.  In the case of logging, you could use `filter` and specify the list of `default_ipv4`, `distribution`, etc. so that only those values will be returned.  AFAICT by looking at the ansible code, `default_ipv4` is *not* a fact subset, it is a parameter, so it should not be used in `gather_subset`, only in `filter`, and I'm not sure why it works, but since this is undocumented behavior, we should not use it.  So to be maximally performant, we should use something like this:
```yaml
 - name: Ensure ansible_facts used by role
   setup:
     gather_subset:
       - min
       - network
     filter: "{{ ['default_ipv4'] + __logging_required_facts }}"
   when: not ansible_facts.keys() | list |
     intersect(__logging_required_facts) == __logging_required_facts
```
that is - gather the minimal and network facts, and only return the values specified in the `filter` list, which are the ones needed by the role.

I'm not sure how much of a performance gain it is to specify `filter`, but in this case, it is pretty straightforward to specify the list of parameters.

One problem is that I don't know if specifying a list of parameters to `filter` is compatible with 2.9.  The documentation says this: https://docs.ansible.com/ansible/2.9/modules/setup_module.html#setup-module
which implies that `filter` is a string, not a list.  I didn't look at the 2.9 code, but the 2.12 code will accept `filter` as a list or as a string.  This is one of the reasons I didn't want to use `filter` in general, because the API is unclear.  However, if you want to do some investigation to determine if it is safe to use a list for 2.9, then we can use `filter` with a list.

tests/setup-snapshot.yml
richm: > If the performance is the issue, I compared the duration time between the 2nd commit version [03b00f5#diff-7eb2a312068fa9ec1be5e435acb86b2def3a99f30b6ff3092f42964283bd9fa8](https://github.com/linux-system-roles/logging/commit/03b00f557cc525e54eebb3af3bc6b3d7f210009e#diff-7eb2a312068fa9ec1be5e435acb86b2def3a99f30b6ff3092f42964283bd9fa8) and this one. There were almost no difference... But probably, we need some way to avoid the risk to bring in an unexpected bug... Let me think about it...

Right.  The role should do nothing if all we want to do is set some vars.

roles/rsyslog/tasks/inputs/ovirt/set_vars.yml
nhosoi: Thank you for your detailed explanation, @richm. You are right. `gather_subset` is supposed take only specific values... Somehow I thought it also took individual fact (and it's unfortunate it "accidentally" does...) Since it is not documented, it could stop working any time. So, I should replace it with `network`. And I don't think the `filter` part is needed. The reason I thought having `default_ipv4` was a good idea was it could just retrieve the ipv4 fact, instead of all the `network` facts. Retrieving all then dropping the rest does not provide us much benefit, I guess.

richm (approved):
lgtm

nhosoi:
[citest pending]

spetrosi (approved):
lgtm

nhosoi:
[citest pending]

nhosoi:
Thank you for the reviews, @richm and @spetrosi!

PR #277, opened 2022-06-22
use gather_facts: true in tests_include_vars_from_parent

richm:
This test does not work with ANSIBLE_GATHERING=explicit - add
explicit gather_facts: true


nhosoi (approved):
lgtm

PR #292, opened 2022-08-16
Use the firewall role and the selinux role to manage the ports; use the certificate role to generate certificates.

nhosoi:
- Introduce `logging_firewall` to use the firewall role to manage the ports.
- Introduce `logging_selinux_ports` to use the selinux role to manage the ports.  
- Introduce `logging_certificates` to generate certificates used in the role
- Add meta/collection-requirements.yml

### DIFF
--- a/roles/rsyslog/tasks/main_core.yml
+++ b/roles/rsyslog/tasks/main_core.yml
@@ -308,6 +308,14 @@
         selinux_ports: "{{ logging_selinux_ports }}"
       when: logging_selinux_ports | length > 0
 
+    # certificates
+    - name: Generate certificates
+      include_role:
+        name: fedora.linux_system_roles.certificate
+      vars:
+        certificate_requests: "{{ logging_certificates }}"
+      when: logging_certificates | length > 0
+
     - block:
         # Check if tls is enabled in forwards output or remote input and
         #          logging_pki_files.{ca_cert_src,ca_cert} is defined
--- a/roles/rsyslog/tasks/main_core.yml
+++ b/roles/rsyslog/tasks/main_core.yml
@@ -292,12 +292,21 @@
       when: __rsyslog_enabled | bool
       notify: restart rsyslogd
 
+    # firewall role
     - name: Manage firewall for specified ports
       include_role:
         name: fedora.linux_system_roles.firewall
       vars:
         firewall: "{{ logging_firewall }}"
       when: logging_firewall | length > 0
+
+    # selinux role
+    - name: Manage selinux for specified ports
+      include_role:
+        name: fedora.linux_system_roles.selinux
+      vars:
+        selinux_ports: "{{ logging_selinux_ports }}"
+      when: logging_selinux_ports | length > 0
 
     - block:
         # Check if tls is enabled in forwards output or remote input and
--- a/roles/rsyslog/tasks/main_core.yml
+++ b/roles/rsyslog/tasks/main_core.yml
@@ -292,6 +292,13 @@
       when: __rsyslog_enabled | bool
       notify: restart rsyslogd
 
+    - name: Manage firewall for specified ports
+      include_role:
+        name: fedora.linux_system_roles.firewall
+      vars:
+        firewall: "{{ logging_firewall }}"
+      when: logging_firewall | length > 0
+
     - block:
         # Check if tls is enabled in forwards output or remote input and
         #          logging_pki_files.{ca_cert_src,ca_cert} is defined
--- a/README.md
+++ b/README.md
@@ -36,7 +36,15 @@ To satisfy such requirements, logging role introduced 3 primary variables `loggi
 
 ## Requirements
 
-This role is supported on RHEL/CentOS-7, RHEL/CentOS-8 and Fedora distributions.
+This role is supported on CentOS-7, CentOS-8-Stream+ and Fedora distributions
+as well as on RHEL-7+.
+
+The role requires the following collections:
+* `fedora.linux_system_roles`
+Use this to install the collections:
+```
+ansible-galaxy collection install -vv -r meta/collection-requirements.yml
+```
 
 ## Definitions
 
@@ -422,6 +430,33 @@ These variables are set in the same level of the `logging_inputs`, `logging_outp
   will be uninstalled and reinstalled in order to revert back to the original
   system default configuration.
 - `logging_system_log_dir`: Directory where the local log output files are placed. Default to `/var/log`.
+- `logging_firewall`: This is a `list` of `dict` in the same format as used by the
+  `fedora.linux_system_role.firewall` role.  Use this to specify ports that you
+  want the role to manage in the firewall.
+```yaml
+    # Manage ports 514/tcp, 514/udp, 20514/tcp, and 20514/udp
+    logging_firewall:
+      - port: 514/tcp
+        state: enabled
+      - port: 514/udp
+        state: enabled
+      - port: 20514/tcp
+        state: enabled
+      - port: 20514/udp
+        state: enabled
+```
+```yaml
+    # Stop managing ports 514/tcp, 514/udp, 20514/tcp, and 20514/udp
+    logging_firewall:
+      - port: 514/tcp
+        state: disabled
+      - port: 514/udp
+        state: disabled
+      - port: 20514/tcp
+        state: disabled
+      - port: 20514/udp
+        state: disabled
+```
 
 ### Update and Delete
 
@@ -645,6 +680,11 @@ Deploying `basics input` reading logs from systemd journal and `forwards output`
   roles:
     - linux-system-roles.logging
   vars:
+    logging_firewall:
+      - port: 514/udp
+        state: enabled
+      - port: 514/tcp
+        state: enabled
     logging_inputs:
       - name: basic_input
         type: basics
@@ -706,6 +746,15 @@ Deploying `remote input` reading logs from remote rsyslog and `remote_files outp
   roles:
     - linux-system-roles.logging
   vars:
+    logging_firewall:
+      - port: 514/udp
+        state: enabled
+      - port: 1514/udp
+        state: enabled
+      - port: 514/tcp
+        state: enabled
+      - port: 1514/tcp
+        state: enabled
     logging_inputs:
       - name: remote_udp_input
         type: remote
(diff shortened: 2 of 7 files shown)


richm:
not including the logging_firewall parameter, what are the other parameters where the user specifies a port number?  In what format does the user specify the port i.e. is it a simple number like `relp_port: 8765` or is it part of some other value `relp_url: https://host:port`?

nhosoi:
> not including the logging_firewall parameter, what are the other parameters where the user specifies a port number? In what format does the user specify the port i.e. is it a simple number like `relp_port: 8765` or is it part of some other value `relp_url: https://host:port`?

A good question... `rsyslog relp port` is `20514` by default. And I cannot find the port number in the `/usr/lib/firewalld/services` dir. Just these two syslog*.xml are found.
```
==> syslog-tls.xml <==
<?xml version="1.0" encoding="utf-8"?>
<service>
  <short>syslog-tls</short>
  <description>Syslog is a client/server protocol: a logging application transmits a text message to the syslog receiver. The receiver is commonly called syslogd, syslog daemon or syslog server. Syslog-tls uses TLS encryption to protect the messages during transport.</description>
  <port protocol="tcp" port="6514"/>
  <port protocol="udp" port="6514"/>
</service>

==> syslog.xml <==
<?xml version="1.0" encoding="utf-8"?>
<service>
  <short>syslog</short>
  <description>Syslog is a client/server protocol: a logging application transmits a text message to the syslog receiver. The receiver is commonly called syslogd, syslog daemon or syslog server.</description>
  <port protocol="udp" port="514"/>
</service>
```
When we configure `firewalld` for a role like `logging`, we are supposed to set up the standard ports? Then, in addition, we configure the custom ones like `relp_port: 8765` if any?

richm:
> > not including the logging_firewall parameter, what are the other parameters where the user specifies a port number? In what format does the user specify the port i.e. is it a simple number like `relp_port: 8765` or is it part of some other value `relp_url: https://host:port`?
> 
> A good question... `rsyslog relp port` is `20514` by default. And I cannot find the port number in the `/usr/lib/firewalld/services` dir. Just these two syslog*.xml are found.
> 
> ```
> ==> syslog-tls.xml <==
> <?xml version="1.0" encoding="utf-8"?>
> <service>
>   <short>syslog-tls</short>
>   <description>Syslog is a client/server protocol: a logging application transmits a text message to the syslog receiver. The receiver is commonly called syslogd, syslog daemon or syslog server. Syslog-tls uses TLS encryption to protect the messages during transport.</description>
>   <port protocol="tcp" port="6514"/>
>   <port protocol="udp" port="6514"/>
> </service>
> 
> ==> syslog.xml <==
> <?xml version="1.0" encoding="utf-8"?>
> <service>
>   <short>syslog</short>
>   <description>Syslog is a client/server protocol: a logging application transmits a text message to the syslog receiver. The receiver is commonly called syslogd, syslog daemon or syslog server.</description>
>   <port protocol="udp" port="514"/>
> </service>
> ```
> 
> When we configure `firewalld` for a role like `logging`, we are supposed to set up the standard ports?

If the user specifies the logging role to manage the firewall, and use forwarding, and does not specify a custom port, then the logging role should do
```
include_role:
  name: fedora.linux_system_roles.firewall
vars:
  firewall:
    - service: syslog
      state: enabled
```
same with forwarding with tls - logging should enable service `syslog-tls`

If the user specifies the logging role to manage the firewall, and use forwarding, and does specify a custom port, then the logging role should parse the port number out of the specification and do
```
include_role:
  name: fedora.linux_system_roles.firewall
vars:
  firewall:
    - port: PORTNUM/tcp
      state: enabled
```

> Then, in addition, we configure the custom ones like `relp_port: 8765` if any?

Yes.

IMO the role should not require the user to specify `logging_firewall` - the role should only require the user to specify `logging_manage_firewall: true` - the role should figure out what ports/services are required in the firewall and call the firewall role to enable them.  This depends on the questions I asked above - how hard is it to determine which ports and services need to be enabled in the firewall based on the inputs to the logging role?

nhosoi:
@richm, as you see in https://github.com/linux-system-roles/logging/pull/292/files#diff-3d0ff1709ca48add100327bb2a468e6c508fb92a159c64c4f99ad1df89d9bddeR53-R151, it's quite tedious to gather the port numbers. Do you think there's easier and cleaner way to achieve it? 

Luckily, we can use the port lists for the selinux role, too, which is good. :)

nhosoi:
[citest]

roles/rsyslog/tasks/main_core.yml
```diff
@@ -292,6 +292,37 @@
...
+      vars:
+        firewall: "{{ logging_firewall }}"
+      when: logging_firewall | length > 0
+
+    - name: "Prepare _logging_selinux_ports to drop ports defined
+            in the selinux policy"
+      set_fact:
+        _logging_selinux_ports: []
+
+    - name: Drop ports defined in the selinux policy
```
richm, line 307: What happens when you pass in ports that are already defined in the selinux policy?

nhosoi: > What happens when you pass in ports that are already defined in the selinux policy?

Add is ok. Just silently skipped ("changed": false). But if I try to delete them with these parameters:
```
logging_selinux_ports:
   - ports: 1514
     proto: tcp
     setype: syslogd_port_t
     state: absent
   - ports: 20514
     proto: tcp
     setype: syslogd_port_t
     state: absent
```
the selinux role (in the collection) fails like this:
```
TASK [fedora.linux_system_roles.selinux : Set an SELinux label on a port] ******
<<snip>>
failed: [/home/nhosoi/.cache/linux-system-roles/rhel-9.qcow2] (item={'ports': 20514, 'proto': 'tcp', 'setype': 'syslogd_port_t', 'state': 'absent'}) => {
    "ansible_loop_var": "item",
    "changed": false,
    "item": {
        "ports": 20514,
        "proto": "tcp",
        "setype": "syslogd_port_t",
        "state": "absent"
    }
}

MSG:
ValueError: Port tcp/20514 is defined in policy, cannot be deleted

task path: /home/nhosoi/linux-system-roles/logging/roles/rsyslog/handlers/main.yml:2
Friday 19 August 2022  15:57:30 -0700 (0:00:02.422)       0:01:53.744 ********* 

/home/nhosoi/.cache/linux-system-roles/rhel-9.qcow2 : ok=126  changed=20   unreachable=0    failed=1    skipped=114  rescued=0    ignored=0   
```

nhosoi:
[citest]

roles/rsyslog/tasks/main_core.yml
richm: > > What happens when you pass in ports that are already defined in the selinux policy?
> 
> Add is ok. Just silently skipped ("changed": false). But if I try to delete them with these parameters:
> 
> ```
> logging_selinux_ports:
>    - ports: 1514
>      proto: tcp
>      setype: syslogd_port_t
>      state: absent
>    - ports: 20514
>      proto: tcp
>      setype: syslogd_port_t
>      state: absent
> ```
> 
> the selinux role (in the collection) fails like this:
> 
> ```
> TASK [fedora.linux_system_roles.selinux : Set an SELinux label on a port] ******
> <<snip>>
> failed: [/home/nhosoi/.cache/linux-system-roles/rhel-9.qcow2] (item={'ports': 20514, 'proto': 'tcp', 'setype': 'syslogd_port_t', 'state': 'absent'}) => {
>     "ansible_loop_var": "item",
>     "changed": false,
>     "item": {
>         "ports": 20514,
>         "proto": "tcp",
>         "setype": "syslogd_port_t",
>         "state": "absent"
>     }
> }
> 
> MSG:
> ValueError: Port tcp/20514 is defined in policy, cannot be deleted
> 
> RUNNING HANDLER [/home/nhosoi/linux-system-roles/logging/tests/roles/linux-system-roles.logging/roles/rsyslog : restart rsyslogd] ***
> task path: /home/nhosoi/linux-system-roles/logging/roles/rsyslog/handlers/main.yml:2
> Friday 19 August 2022  15:57:30 -0700 (0:00:02.422)       0:01:53.744 ********* 
> 
> PLAY RECAP *********************************************************************
> /home/nhosoi/.cache/linux-system-roles/rhel-9.qcow2 : ok=126  changed=20   unreachable=0    failed=1    skipped=114  rescued=0    ignored=0   
> ```

I see.  This is going to be tricky.  I can think of a couple of ways to do this:

* use a `register` and a `failed_when`
```
  register: __logging_seport
  failed_when: not __logging_seport.msg is search('is defined in policy, cannot be deleted')
```
I'm not sure which field in the register variable holds the message.  If register/failed_when will simply not work, then try a block/rescue

* use `semanage port -C --list` - parse the list into a form you can use to compare against the values in logging_selinux_ports, then only apply the ones in that list

nhosoi: Thank you, @richm. Let me try your ideas.

BTW, I guess it's too early for me to set `[citest]`. :) We need our citool pr merged first to make the ci tests in this pr pass...

richm: > Thank you, @richm. Let me try your ideas.

Thinking about this some more
* Using register/failed_when probably won't work correctly because we're using it with the `include_role`, and there may be multiple ports we want to try to disable - would have to be in a loop to try each port separately - same with block/rescue
* Using `semanage port -C --list` means having to ensure that the `semanage` command is installed - not sure if the `selinux` role does that
* The easiest way would be if the `selinux` role had an option to allow you to ignore errors when trying to remove built-in policy - that way you could just pass in all of them and ignore all of the errors

> 
> BTW, I guess it's too early for me to set `[citest]`. :) We need our citool pr merged first to make the ci tests in this pr pass...

There are several roles now that need collection-requirements.yml

nhosoi:
This PR will be updated once this selinux https://github.com/linux-system-roles/selinux/pull/122 is completed and merged.

nhosoi:
Updated the `selinux` role part to use the [new implementation](https://github.com/bachradsusi/linux-system-roles-selinux/tree/seport_local) proposed by @bachradsusi.
Verified the tests passed using the branch.

richm:
[citest]

nhosoi:
Closing this pr in favor of https://github.com/linux-system-roles/logging/pull/293.

PR #295, opened 2022-10-20
Elaborate the logging_certificates usage in README.

nhosoi:


nhosoi:
@richm, @rjeffman,
I'd think to use the `certificate` role to generate a key, cert pair for encrypting the logs over TLS, a CA cert issued by the FreeIPA server is needed. (self-signed cert does not work for the logging cases.) I'd like to emphasize it in README...

If you could share your comments with me, I'd greatly appreciate it. Thanks!

nhosoi:
[citest]

README.md
```diff
@@ -445,25 +445,51 @@ These variables are set in the same level of the `logging_inputs`, `logging_outp
...
+     - name: relp_server
+       type: relp
+       tls: true
+       ca_cert: /etc/ipa/ca.crt
+       cert: /etc/pki/tls/certs/logging_cert.crt
+       private_key: /etc/pki/tls/private/logging_cert.key
+       [snip]
+```
+  NOTE: The `certificate` role supports self-signed CA, but it may lead to
+  security problems. Therefore, it must not be used in the public system.
```
richm, line 492: Maybe something like this: "NOTE: The `certificate` role, unless using IPA and joining the systems to an IPA domain, creates self-signed certificates, so you will need to explicitly configure trust, which is not currently supported by the system roles."

nhosoi: Thanks a lot, @richm. Updated the PR.

richm (approved):

PR #300, opened 2022-11-18
cert cleanup needs to use getcert stop-tracking

richm:
Need to use `getcert stop-tracking` when removing the cert, otherwise,
subsequent requests will hang if certmonger is tracking a non-existent
cert.


richm:
[citest]

richm:
[citest]

richm:
[citest]

richm:
[citest]

PR #304, opened 2022-12-05
Version 1.11.3 - CHANGELOG.md [citest skip]

richm:
[1.11.3] - 2022-12-05
--------------------

### New Features

- none

### Bug Fixes

- use logging_purge_confs in relp test (#303)

On some platforms, there are no packages that put config files in
/etc/rsyslog.d - using `logging_enabled: false` after a test leaves
no files in /etc/rsyslog.d and the `$IncludeConfig /etc/rsyslog.d/*.conf`
then leaves rsyslogd in a state that it will not start and cause test
failures.
The solution is to use `logging_purge_confs: true` which "does the right
thing" when /etc/rsyslog.d is empty.
Fix the expected number of config files check.

### Other Changes

- none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,30 @@
 Changelog
 =========
 
+[1.11.3] - 2022-12-05
+--------------------
+
+### New Features
+
+- none
+
+### Bug Fixes
+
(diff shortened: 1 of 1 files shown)


nhosoi (approved):
Lgtm

PR #308, opened 2022-12-13
specify empty inputs, outputs, flows with purge

richm:
If there are any leftover defined variables that specify inputs, outputs, or flows,
purge will not remove those files since they are referred to.  This typically
happens when using `public: true` which leaves all of those role variables
defined in the public namespace.  The solution when cleaning up is to ensure
that none of the inputs, outputs, or flows is defined when using purge, to
guarantee that all of the config files are removed and none are leftover.


richm:
[citest]

nhosoi (approved):
lgtm

PR #310, opened 2022-12-15
Add check for non-inclusive language

nhosoi:
Add a check for usage of terms and language that is considered
non-inclusive. We are using the woke tool for this with a wordlist
that can be found at
https://github.com/linux-system-roles/tox-lsr/blob/main/src/tox_lsr/config_files/woke.yml

CHANGELOG.md - cleanup non-inclusive words.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -146,14 +146,9 @@ Adds startmsg_regex and endmsg_regex options:
 
 - Fix a bash bug in changelog_to_tag.yml, which unexpectedly expanded "*"
 
-- changelog_to_tag action - support other than "master" for the main branch name, as well
+- changelog_to_tag action - github action ansible test improvements
 
-- Use GITHUB_REF_NAME as name of push branch; fix error in branch detection [citest skip]
-
-We need to get the name of the branch to which CHANGELOG.md was pushed.
-For now, it looks as though `GITHUB_REF_NAME` is that name.  But don't
-trust it - first, check that it is `main` or `master`.  If not, then use
-a couple of other methods to determine what is the push branch.
+- Use GITHUB_REF_NAME as name of push branch; fix error in branch detection
 
 Signed-off-by: Rich Megginson <rmeggins@redhat.com>
 
@@ -445,7 +440,7 @@ must be a `string` value like `"2.9"`, not a `float` value like `2.9`.
 ### Other Changes
 
 - Remove python-26 environment from tox testing
-- update to tox-lsr 2.4.0 - add support for ansible-test sanity with docker
+- update to tox-lsr 2.4.0 - add support for ansible-test with docker
 - use tox-lsr 2.3.0 and enable ansible-test
 - CI: Add support for RHEL-9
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -146,13 +146,13 @@ Adds startmsg_regex and endmsg_regex options:
...
 
-- changelog_to_tag action - support other than "master" for the main branch name, as well
+- changelog_to_tag action - support other than "master" for the main branch name, as well // wokeignore:rule=master
 
 - Use GITHUB_REF_NAME as name of push branch; fix error in branch detection [citest skip]
 
 We need to get the name of the branch to which CHANGELOG.md was pushed.
 For now, it looks as though `GITHUB_REF_NAME` is that name.  But don't
-trust it - first, check that it is `main` or `master`.  If not, then use
+trust it - first, check that it is `main` or `master`.  If not, then use // wokeignore:rule=master
```
richm, line 155: ok - but I'd like to see what @Jakuje says about https://github.com/linux-system-roles/crypto_policies/pull/54#issuecomment-1352004655 - if the right answer is "just edit CHANGELOG.md to remove the problematic terms rather than put `wokeignore` in CHANGELOG.md" then this PR becomes much smaller and simpler i.e. can use the official woke instead of the custom

nhosoi:
Test was successful. Warning is not counted as an error.
```
Prepare all required actions
Run ./.github/actions/custom-woke-action
Run $GITHUB_ACTION_PATH/entrypoint.sh
 Installing woke ... https://github.com/nhosoi/woke
 Running woke ...
  Warning: `sanity` may be insensitive, use `confidence`, `quick check`, `coherence check` instead
  No findings found.
```

CHANGELOG.md
nhosoi: Yeah, but this changelog item specifically refers to the `master` branch... How we could describe it without using the word `master`... In this case, I can just remove it, as it does not look that important(?), but there could be some cases that do not allow it???

richm: > Yeah, but this changelog item specifically refers to the `master` branch... How we could describe it without using the word `master`... In this case, I can just remove it, as it does not look that important(?), but there could be some cases that do not allow it???

I think we can remove these entire entries from the changelog.

The problem with the auto-changelog generation that I wrote is
* it is pretty stupid - it just collects all of the git commit messages
* it requires a non-lazy human to edit - I should really have just summarized all of these changes as something like "fix github publish action" or something like that

Especially in this case - these can be replaced with `- fix github action for publishing roles` or similar

richm: Same below - just say `- github action ansible test improvements`

nhosoi: Thanks, @richm! I'm updating CHANGELOG as you suggested.

richm:
please remove/change the reference at https://github.com/linux-system-roles/logging/pull/310/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edL448 also - then you can just use the standard woke action and don't have to use the custom action

nhosoi:
[citest]

richm (approved):

nhosoi:
Thank you for your review, @richm. Merged.

PR #312, opened 2023-01-04
additional ansible-lint 6.x fixes

richm:
All tasks must be named
Task names must begin with uppercase letter
vars and when should come before block/rescue/always
There must be only 1 template in a name, and it must come last


richm:
[citest]

PR #320, opened 2023-02-10
Use custom shellcheck github action

nhosoi:
Remove shellcheck github action since there are no shell scripts in the role.


richm:
I guess you'll need to add a shell script with some problems to see what happens.

Also, this should go into https://github.com/linux-system-roles/.github/blob/main/playbooks/templates/.github/workflows/shellcheck.yml

nhosoi:
I noticed `Code scanning` also checks the shellcheck errors/warnings.

I put a variable with no double-quotes, which makes `Differential ShellCheck` fail; not `Code scanning`. In `Code scanning`, the error is defined as a warning and it issues 1 alert, but the test passes successfully.

Another difference is `Code scanning` reports multiple errors/warnings, but it seems `Differential ShellCheck` quits at the first error. (There could be an option to run through the end even if it encountered errors...?)

.github/workflows/shellcheck.yml
```diff
@@ -16,10 +16,6 @@ jobs:
       # required for all workflows
       security-events: write
 
-      # only required for workflows in private repositories
-      actions: read
-      contents: read
```
richm: Why are these unnecessary?

richm:
Not sure what `Code scanning results` is - it just seems to have been added with Differential ShellCheck

.github/workflows/shellcheck.yml
nhosoi: I thought we don't use "private repositories"... (and removing them did not break the test...)

richm: I'm not sure what that means, but I recently added permissions based on a suggestion from the journald team to all of our github actions - https://github.com/linux-system-roles/.github/pull/8/files

nhosoi: It seems the examples listed here do not have them... (but the permissions are not identical to each other...)
https://github.com/redhat-plumbers-in-action/differential-shellcheck#real-life-examples-of-usage

nhosoi:
> Not sure what `Code scanning results` is - it just seems to have been added with Differential ShellCheck

Yeah, you are right. I tested with the original `shellcheck.yml` with https://github.com/linux-system-roles/logging/pull/321 and saw no `Code scanning results`. I.e., differential-shellcheck is much better and useful.

.github/workflows/shellcheck.yml
nhosoi: I removed the entire permissions section from the `lint job`, but the test ran and failed as expected...
```
    permissions:
      # required for all workflows
      security-events: write

      # only required for workflows in private repositories
      actions: read
      contents: read
```

richm: I'm not exactly sure how permissions work - but I think the general idea is that the default permissions could allow too much - better to follow the model of the other github actions that use a `permission: contents: read` https://github.com/linux-system-roles/logging/pull/320/files#diff-eace62a8ebcdafd0f6975200d68cbf851300952aae47db07e71f383038a368b3R8 for the whole action - then relax that permission for individual jobs where needed (using `security-events: write` - perhaps we don't need the others)

richm: Here are the types of REST actions allowed by `security-events: read|write` https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#security-events
It may be that the action cannot update `Code scanning` if it doesn't have `security-events: write` permission

richm (approved):

PR #326, opened 2023-04-07
ansible-lint - use changed_when even for tasks with conditionals

richm:
ansible-lint now requires the use of changed_when even for tasks
that are conditional


richm:
[citest]

richm:
[citest]

PR #333, opened 2023-05-24
fix: work with ansible-core 2.15

richm:
Variables set by a task `vars` with an `include_role` task
are no longer exported to the play with a `public: true`.
Use `set_vars` to set the variables globally so they can
be used both by the role and the test.

### DIFF
--- a/tests/tests_basics_files.yml
+++ b/tests/tests_basics_files.yml
@@ -19,11 +19,14 @@
 
   tasks:
     # TEST CASE 0
+    - name: Set firewall and selinux to true
+      set_fact:
+        logging_manage_firewall: true
+        logging_manage_selinux: true
+
     - name: "TEST CASE 0; Test the simplest configuration,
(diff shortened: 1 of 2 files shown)


nhosoi:
> Variables set by a task vars with an include_role task are no longer exported to the play with a public: true.

Wow, that’s a surprise…

> Error: Variables names from within roles should use role_name_ as a prefix. (vars: __logging_forwards_tls)

And even `__role_name_` is complained by ansible lint… :(

richm:
> > Variables set by a task vars with an include_role task are no longer exported to the play with a public: true.
> 
> Wow, that’s a surprise…

It happened in a few other roles.

> 
> > Error: Variables names from within roles should use role_name_ as a prefix. (vars: __logging_forwards_tls)
> 
> And even `__role_name_` is complained by ansible lint… :(

Yes - this is an ansible-lint bug which should be fixed shortly

richm:
[citest]

PR #342, opened 2023-07-18
refactor: Use selinux role instead of selinux cli

richm:
The motivation was getting rid of the need to use the `ansible_selinux` fact.
The configuration of the selinux policy for the collectd port was using
a `shell`.  This has been changed to use the selinux role.
In addition - the logic would cause rsyslogd to restart every time, so that
logic has been preserved.

I don't think anyone is using the ovirt functionality, so it probably doesn't
matter, but in the future we should make the configuration of the port
conditional on `logging_manage_selinux`, possibly also configure firewall,
and make the `notify` conditional upon changes.


richm:
[citest]

nhosoi (approved):
lgtm
Thank you, @richm!

PR #344, opened 2023-08-09
ci: Add markdownlint, test_html_build, and build_docs workflows

spetrosi:
- markdownlint runs against README.md to avoid any issues with
  converting it to HTML
- test_converting_readme converts README.md > HTML and uploads this test
  artifact to ensure that conversion works fine
- build_docs converts README.md > HTML and pushes the result to the
  docs branch to publish dosc to GitHub pages site.
- Fix markdown issues in README.md

Signed-off-by: Sergei Petrosian <spetrosi@redhat.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,31 +2,6 @@
 
 ![CI Testing](https://github.com/linux-system-roles/logging/workflows/tox/badge.svg)
 
-## Table of Contents
-
-<!--ts-->
-  * [Background](#background)
-  * [Definitions](#definitions)
-  * [Logging Configuration Overview](#logging-configuration-overview)
-  * [Variables](#variables)
-    * [Logging_inputs options](#logging_inputs-options)
-    * [Logging_outputs options](#logging_outputs-options)
-    * [Logging_flows options](#logging_flows-options)
-    * [Security options](#security-options)
-    * [Server performance optimization options](#server-performance-optimization-options)
-    * [Other options](#other-options)
-  * [Update and Delete](#update-and-delete)
-  * [Configuration Examples](#configuration-examples)
-    * [Standalone configuration](#standalone-configuration)
-    * [Client configuration](#client-configuration)
-    * [Server configuration](#server-configuration)
-    * [Client configuration with Relp](#client-configuration-with-relp)
-    * [Server configuration with Relp](#server-configuration-with-relp)
-  * [Port and SELinux](#port-and-selinux)
-  * [Providers](#providers)
-  * [Tests](#tests)
-<!--te-->
-
 ## Background
 
 Logging role is an abstract layer for provisioning and configuring the logging system. Currently, rsyslog is the only supported provider.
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -275,7 +259,7 @@ Allowed values are `traditional`, `syslog`, and `modern`. Default to `modern`.
 
 **Note:** Unless the above options are given, these local file outputs are configured.
 
-```
+```bash
```
spetrosi, line 262: @richm markdownlint requires some syntax language to be specified with code blocks. Can you think of something other than bash in this case? The list of supported languages: https://docs.readme.com/rdmd/docs/code-blocks

README.md
```diff
@@ -286,71 +270,74 @@ Allowed values are `traditional`, `syslog`, and `modern`. Default to `modern`.
...
                        `/path/to/output/dir/%FROMHOST%/%PROGRAMNAME:::secpath-replace%.log`
-- `remote_sub_path`: Relative path to logging_system_log_dir to store the filtered logs.
+* `remote_sub_path`: Relative path to logging_system_log_dir to store the filtered logs.
 
 **Note:** Selector options and property-based filter options are exclusive. If Property-based filter options are defined, selector options will be ignored.
 
 **Note:** If both `remote_log_path` and `remote_sub_path` are _not_ specified, the remote_file output configured with the following settings.
 
-```
+```java
```
spetrosi, line 340: And here I guess java is the one?

README.md
```diff
@@ -52,25 +27,25 @@ satisfied.
 
 Otherwise, please run the following command line to install the collection.
 
-```
+```bash
```
spetrosi, line 30: @richm this Collection requirements section comes from template's README, right? We do not copy it automatically. It's identical in many roles so better to ask.

richm: > @richm this Collection requirements section comes from template's README, right?

Right.

> We do not copy it automatically.

Right.

> It's identical in many roles so better to ask.

If it is identical, that means it didn't need to be changed.

README.md
richm: syntax language is required?  The original markdown spec says you can have fenced code blocks without specifying the syntax - https://www.markdownguide.org/extended-syntax/#fenced-code-blocks

README.md
richm: Since it works - yes - but the rsyslog config language is not generally like java

README.md
```diff
@@ -397,7 +384,7 @@ When TLS connection is configured, `ca_cert_src` and/or `ca_cert` is required.
 To configure the certificate of the logging system, `cert_src` and/or `cert` is required.
 To configure the private key of the logging system, `private_key_src` and/or `private_key` is required.
 
-```
+```yaml
```
richm, line 387: This needs to be reformatted entirely - see https://github.com/linux-system-roles/logging/pull/345

README.md
```diff
@@ -906,16 +899,20 @@ associated TLS value.
...
-```
+
+```bash
 semanage port --list | grep "syslog"
 ```
+
 The newly specified port will be added to this default set.
-```
+
+```bash
```
richm, line 915: would be better to not have a syntax here - not sure why the processor does not allow code blocks without a syntax

PR #361, opened 2023-10-24
fix: check that logging_max_message_size is set, not rsyslog_max_message_size

MrZmurf:
Enhancement: Check that `logging_max_message_size` is set, not `rsyslog_max_message_size`

Reason: The play does not check if the correct variable is defined

Result: Changed so that the test and action use the same variable name

### DIFF
--- a/roles/rsyslog/templates/global.j2
+++ b/roles/rsyslog/templates/global.j2
@@ -27,7 +27,7 @@ global(
   defaultNetstreamDriverKeyFile="{{ key_path }}"
 {%   endif %}
 {% endif %}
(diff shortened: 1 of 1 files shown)


spetrosi:
Thank you for the fix @MrZmurf, please fill in PR template with Reason and Result of this to provide more info and to have an informative changelog entry.

spetrosi:
[citest]

MrZmurf:
Reason: The play does not check if the the correct variable is defined.
Result: Changed so that the test and action use the same variable name.

PR #362, opened 2023-10-24
feat: Add support for the global config option preserveFQDN with a new logg…

MrZmurf:
…ing_preserve_fqdn variable

Enhancement:

Reason:

Result:

### DIFF
--- a/roles/rsyslog/templates/global.j2
+++ b/roles/rsyslog/templates/global.j2
@@ -30,7 +30,7 @@ global(
 {% if rsyslog_max_message_size is defined %}
   maxMessageSize="{{ logging_max_message_size }}"
 {% endif %}
-{% if logging_preserve_fqdn is defined %}
-  preserveFQDN="{{ logging_preserve_fqdn }}"
+{% if logging_preserve_fqdn is defined and logging_preserve_fqdn is sameas true %}
+  preserveFQDN="on"
 {% endif %}
 )
--- a/roles/rsyslog/templates/global.j2
+++ b/roles/rsyslog/templates/global.j2
@@ -30,7 +30,7 @@ global(
 {% if rsyslog_max_message_size is defined %}
   maxMessageSize="{{ logging_max_message_size }}"
 {% endif %}
-{% if logging_preserve_fqdn is defined and logging_preserve_fqdn is sameas true %}
+{% if logging_preserve_fqdn | d() %}
   preserveFQDN="on"
 {% endif %}
 )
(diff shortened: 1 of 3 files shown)


spetrosi:
Thank you for your contributions @MrZmurf. Please fill in the PR description template in this PR to provide more info about the change so that this change is described in CHANGELOG properly.

Please also update README.md to list `logging_preserve_fqdn` and add a test for setting `logging_preserve_fqdn` similar to how `logging_mark: true` is tested in https://github.com/linux-system-roles/logging/blob/main/tests/tests_basics_files.yml#L32

MrZmurf:
Enhancement: Add support to set the preserveFQDN option in rsync configuration.

Reason: There was previously no way to configure the preserveFQDN config option with this role.

Result: Add the optional logging_preserve_fqdn variable to set the preserveFQDN configuration option in syslog. This will make it configurable to use the full FQDN instead of short name in syslog entries.

roles/rsyslog/templates/global.j2
```diff
@@ -30,4 +30,7 @@ global(
 {% if rsyslog_max_message_size is defined %}
   maxMessageSize="{{ logging_max_message_size }}"
 {% endif %}
+{% if logging_preserve_fqdn | d() %}
```
richm, line 33: Since `logging_preserve_fqdn` is in the public API of the role (which are the parameters that users can set), it should be defined in `defaults/main.yml` as `logging_preserve_fqdn: false` - then it will always be defined and you can omit the `| d()`

MrZmurf: Ok, I have set it in defaults.

MrZmurf: But then again, should it be in globa defaults or should i change it to rsyslog_preserve_fqdn and put it in the defaults of the rsyslog role since it is rsyslog specific.

There are also other options in specified at the logger level that probably should be rsyslog specific.

richm: You are correct that this is an rsyslog specific config.  However, we broke that API convention about 4 years ago when we decided not to add fluentd support to the logging role, and ever since then, we've been adding rsyslog specific parameters to the top level defaults/main.yml - https://github.com/linux-system-roles/logging/blob/main/defaults/main.yml#L41 - starting here, most of these are rsyslog specific.

So I would prefer putting this in the global defaults as a `logging_` parameter, and if/when we ever refactor this role to support some logging provider other than rsyslog, we'll have to figure out how to do that.

richm:
[citest]

richm (approved):

PR #364, opened 2023-11-01
feat: Add support for general queue and general action parameters

MrZmurf:
Enhancement: Add support for general queue and general action parameters in output modules

Reason: The role is missing the ability to configure general queue parameters and general action parameters

Result: This PR implements a way to add general queue and general action parameters to output modules


roles/rsyslog/templates/general_action_params.j2
```diff
@@ -0,0 +1,7 @@
+{% if __rsyslog_output.action is defined and __rsyslog_output.action is mapping %}
+{%   for option, value in __rsyslog_output.action.items() %}
+{%     if option | lower in rsyslog_valid_general_action_params %}
+action.{{ option }}="{{ value }}"
```
richm, line 4: some of the action parameters have types other than `string` - e.g. https://www.rsyslog.com/doc/v8-stable/configuration/actions.html#general-action-parameters
Some of them are integers - is it ok to have `action.resumeInterval="30"`?  or does it have to be `action.resumeInterval=30`?  I guess the same applies to the queue parameters below.

MrZmurf: Fair, I guess a better approach would be 

{% if __rsyslog_output.action is defined and __rsyslog_output.action is mapping %}
{%   for option, value in __rsyslog_output.action.items() %}
{%     if option | lower in rsyslog_valid_general_action_params and value is number %}
action.{{ option }}={{ value }}
{%     elif option | lower in rsyslog_valid_general_action_params  %}
action.{{ option }}="{{ value }}"

richm: Looks like `action.resumeInterval="30"` is fine - https://github.com/rsyslog/rsyslog/blob/master/tests/es-basic-abort.sh#L19-L20

```
		action.resumeRetryCount="-1"
		action.resumeInterval="1"
```

so, sorry for the noise - your approach looks good

MrZmurf: Or split it in to two dicts

{% if __rsyslog_output.action is defined and __rsyslog_output.action is mapping %}
{% for option, value in __rsyslog_output.action.items() %}
{% if option | lower in rsyslog_valid_general_action_params_as_int %}
action.{{ option }}={{ value }}
{% elif option | lower in rsyslog_valid_general_action_params_as_string %}
action.{{ option }}="{{ value }}"

richm:
@MrZmurf This is a really cool patch - thanks!  I would really like to have some sort of test for this - do you see any of the tests in the `tests` directory that we could add some `queue` or `action` parameters to?

MrZmurf:
> @MrZmurf This is a really cool patch - thanks! I would really like to have some sort of test for this - do you see any of the tests in the `tests` directory that we could add some `queue` or `action` parameters to?

tests_basics_forwards.yml could probably test both queue and action parameters like queue.spooldirectory="/var/spool/rsyslog", queue.filename="forwards", queue.saveonshutdown="on" and action.resumeretrycount=50 and action.resumeInterval=300 or something like that

MrZmurf:
In tests_remote.yml the action.reportsuspension could be set to "on" and action.reportsuspensioncontinuation to "off" just to test the other way the template is used when it writes all options on the same line

richm:
@MrZmurf https://github.com/MrZmurf/logging/pull/1

roles/rsyslog/templates/output_remote_files.j2
```diff
@@ -13,16 +13,25 @@ template(
...
+{%   if __rsyslog_output.queue.type is not defined %}
         queue.type="{{ logging_server_queue_type }}"
+{%   endif %}
+{%   if __rsyslog_output.queue.size is not defined %}
         queue.size="{{ logging_server_queue_size }}"
-        queue.workerThreads="{{ logging_server_threads }}") {
+{%   endif %}
+{%   if __rsyslog_output.queue.workerthreads is not defined %}
+        queue.workerThreads="{{ logging_server_threads }}"
+{%   endif %}
```
MrZmurf, line 24: Not really happy with this code tho, it should not test for the parameter in lower case, the test should be case insensitive.

MrZmurf: % if __rsyslog_output.queue is defined %}
{%   set __queue_options = {} %}
{%   for option in __rsyslog_output.queue.keys() %}
{%     set _none = __queue_options.update({ option | lower : true }) %}
{%   endfor %}
{% endif %}
.....
{%   if __queue_options.type is not defined %}
        queue.type="{{ logging_server_queue_type }}"
{%   endif %}
{%   if  __queue_options.type.size is not defined %}
        queue.size="{{ logging_server_queue_size }}"
{%   endif %}
{%   if __queue_options.workerthreads is not defined %}
        queue.workerThreads="{{ logging_server_threads }}"
{%   endif %}

This would make the test case insensitive.

richm: nice - lgtm

MrZmurf: Well this is a special case just for the output_remote_files.j2 template since that template was already using some custom queue options to get better performance than with default settings.

tests/tests_remote.yml
```diff
@@ -134,6 +134,9 @@
             client_count: 20
             io_buffer_size: 8192
             async_writing: true
+            action:
+              reportsuspension: "on"
+              reportsuspensioncontinuation: "off"
```
richm, line 139:
```suggestion
              reportsuspensioncontinuation: "off"
            queue:
              Size: 49999
```

tests/tests_remote.yml
```diff
@@ -172,6 +175,12 @@
         __check_systemctl_status: true
       include_tasks: tasks/check_daemon_config_files.yml
 
+    - name: Check action parameters in remote0 output
+      command: >-
+        grep 'authpriv.none action.* action.reportsuspension="on"
+        action.reportsuspensioncontinuation="off"' {{ __test_output_remote0 }}
+      changed_when: false
+
```
richm, line 183:
```suggestion
    - name: Check queue parameters in remote0 output
      command: >-
        grep ' queue.Size="49999"' {{ __test_output_remote0 }}
      changed_when: false

```

richm:
[citest]

richm (approved):

PR #398, opened 2024-07-16
feat: add support for reopen_on_truncate for files input

richm:
Feature: Add support for reopen_on_truncate for files input.  This is
a boolean value which sets the rsyslog reopenOnTruncate option for
the imfile module.

Reason: Users need to be able to configure rsyslog to reopen
input file when it was truncated.

Result: Users can configure rsyslog reopenOnTruncate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,11 +115,16 @@ Available options:
 
 Available options:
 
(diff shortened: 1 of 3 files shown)


richm:
[citest]

PR #399, opened 2024-07-16
feat: support custom config files with logging_custom_config_files

richm:
Feature: Allow users to provide custom logging configuration files
using the new variable `logging_custom_config_files` which is a list
of config files to copy to the default logging configuration directory.
For example, this directory for rsyslog is `/etc/rsyslog.d/`.  This
assumes the logging configuration will load all files matching `*.conf`
in this directory as in the default rsyslog directive
`$IncludeConfig /etc/rsyslog.d/*.conf`.

Reason: Users need customized configuration not provided by the role.
The role allowed this previously by the use of the variable
`rsyslog_custom_config_files`, but that variable was intended to be used
privately and is now marked DEPRECATED.

Result: Users can provide custom configuration in a supported manner.


richm:
[citest]

PR #401, opened 2024-07-16
test: use selinux for relp to fix test AVC

richm:
The tests_relp is causing AVCs.  Use selinux in the test to
avoid the AVCs.


richm:
[citest]

PR #406, opened 2024-07-30
test: workaround a systemd issue in el10 restarting journald

richm:
Due to https://github.com/systemd/systemd/pull/33874
we need to clear the SubState=dead-resources-pinned of journald
when restarting it after the test.


richm:
[citest]

richm:
centos 7 failure is unrelated

PR #407, opened 2024-07-30
test: sort files to compare

richm:
On some platforms the parameters are stored in a `dict`, and the ordering
of the elements is not guaranteed to be stable.  So when comparing
the expected results to the actual file, sort both.


richm:
[citest]

PR #416, opened 2024-09-10
test: restart dbus broker after restarting journald

richm:
Restarting journald seems to break the ability of dbus to
configure policy for newly installed services.  Restarting
dbus broker seems to fix that.


richm:
[citest]

spetrosi (approved):

PR #418, opened 2024-09-12
test: restart logind and test

richm:
Restarting journald and dbus also causes problems for systemd-logind, so
restart it too and test with loginctl


richm:
[citest]

spetrosi (approved):

PR #425, opened 2024-11-09
feat: support custom templates

richm:
Feature: Allow users to specify custom templates to use for `files` and
`forwards` outputs.  Users can specify a list of templates to use with
`logging_custom_templates`.  Users can specify the default template to use for
all `files` outputs with `logging_files_template_format`, or specify on a per
output basis by using `template`.  Users can specify the default template to
use for all `forwards` outputs with `logging_forwards_template_format`, or
specify on a per output basis by using `template`.

Reason: Users need the ability to format log entries in different ways other
than the built-in defaults, using custom templates.

Result: Users can specify custom templates to use for files and forwards
outputs.

### DIFF
--- a/roles/rsyslog/templates/output_files.j2
+++ b/roles/rsyslog/templates/output_files.j2
@@ -42,12 +42,19 @@ ruleset(name="{{ __rsyslog_output.name }}"
 ruleset(name="{{ __rsyslog_output.name }}") {
 {%   endif %}
 {{ print_file_attrs(__rsyslog_output) -}}
+{%   set template = " ;RSYSLOG_TraditionalFileFormat"
+       if __rsyslog_output.template | d("") == "traditional"
+       else " ;RSYSLOG_SyslogProtocol23Format"
+       if __rsyslog_output.template | d("") == "syslog"
+       else " ;" ~ __rsyslog_output.template
+       if __rsyslog_output.template | d("") not in ["", "modern"]
+       else "" %}
 {%   if __rsyslog_output.property | d() %}
-    :{{ __rsyslog_output.property }}, {{ __rsyslog_output.property_op | d('contains') }}, "{{ __rsyslog_output.property_value | d('error') }}" {{ __rsyslog_output.path }}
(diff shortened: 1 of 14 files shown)


richm:
[citest]

roles/rsyslog/templates/output_files.j2
```diff
@@ -42,12 +42,19 @@ ruleset(name="{{ __rsyslog_output.name }}"
 ruleset(name="{{ __rsyslog_output.name }}") {
 {%   endif %}
 {{ print_file_attrs(__rsyslog_output) -}}
+{%   set template = " ;RSYSLOG_TraditionalFileFormat"
```
spetrosi, line 45: Is it possible to write it to make it more readable? Something like this but with all logic?
```
{% if __rsyslog_output.template | d("") == "traditional" %}
{%   set template  =   " ;RSYSLOG_TraditionalFileFormat" %}
{% endif %}
```
Same for roles/rsyslog/templates/output_remote_files.j2.

tests/tests_basics_files.yml
```diff
@@ -373,6 +384,18 @@
...
+    - name: Check that the custom template is in the templates file
+      command: grep BasicTestFormat /etc/rsyslog.d/20-templates.conf
+      changed_when: false
+
+    - name: Check that the custom template has an output action defined
+      command: cat /etc/rsyslog.d/30-output-files-custom_template.conf
+      changed_when: false
+
+    - name: Check that the custom template is being used
+      command: grep ThisIsBasicTestFormat$ /var/log/logging_custom_template.log
```
spetrosi, line 396: I like this test.

roles/rsyslog/templates/output_files.j2
richm: > Is it possible to write it to make it more readable?

Unfortunately no, due to the way jinja does scoping.  https://jinja.palletsprojects.com/en/stable/templates/#list-of-builtin-filters
"Scoping Behavior

Please keep in mind that it is not possible to set variables inside a block and have them show up outside of it."

spetrosi: Okay, I was looking at what I did in mssq, but indeed there I am doing it within a single block https://github.com/linux-system-roles/mssql/blob/main/templates/configure_ag.j2

PR #436, opened 2025-02-17
fix: Add a variable for rsyslog base package on Debian

TheLastChosenOne:
On RedHat, it's named "iproute" but on Debian based systems it's named "iproute2"

Enhancement: Make this role functional again for Debian based systems

Reason: Newer version of this role (see commit 64f915a5c) included the package "iproute" which is named "iproute2" on Debian based systems.

Result: Define an own __rsyslog_base_packages list for Debian based systems

### DIFF
--- a/roles/rsyslog/vars/Debian.yml
+++ b/roles/rsyslog/vars/Debian.yml
@@ -1,0 +1,7 @@
+# __rsyslog_base_packages
+#
+# List of default rpm packages to install.
+# NOTE: iproute2 is needed for the ip command which
+# is needed for the default_ipv4 fact
+# package is named iproute2 for Debian/Ubuntu systems
+__rsyslog_base_packages: ['iproute2', 'rsyslog']


roles/rsyslog/vars/main.yml
```diff
@@ -1,12 +1,5 @@
 # SPDX-License-Identifier: GPL-3.0-only
 ---
-# __rsyslog_base_packages
-#
-# List of default rpm packages to install.
-# NOTE: iproute is needed for the ip command which
-# is needed for the default_ipv4 fact
-__rsyslog_base_packages: ['iproute', 'rsyslog']
```
spetrosi, line 8: Return this file.

roles/rsyslog/vars/RedHat.yml
```diff
@@ -0,0 +1,6 @@
+# __rsyslog_base_packages
+#
+# List of default rpm packages to install.
+# NOTE: iproute is needed for the ip command which
+# is needed for the default_ipv4 fact
+__rsyslog_base_packages: ['iproute', 'rsyslog']
```
spetrosi, line 6: Remove this file.
Ansible sets vars from roles/rsyslog/vars/main.yml for every role, then https://github.com/linux-system-roles/logging/blob/main/roles/rsyslog/tasks/set_vars.yml#L22 sets os-specific variables.
So `__rsyslog_base_packages` is set for every OS, then for Debian it is reset from roles/rsyslog/vars/Debian.yml.

TheLastChosenOne: You're right, that's a cleaner way.

spetrosi:
@TheLastChosenOne did you test it on Debian?

TheLastChosenOne:
> @TheLastChosenOne did you test it on Debian?
Tested it on Debian and Ubuntu (up to 24.04).

spetrosi:
[citest]

spetrosi (approved):

PR #358, opened 2023-10-07
ci: This PR is to trigger periodic CI testing

systemroller:
This PR is for the purpose of triggering periodic CI testing. We don't currently have a way to trigger CI without a PR, so this PR serves that purpose.

PR #439, opened 2025-04-08
ci: use tox-lsr 3.6.0

richm:
tox-lsr 3.6.0 will guarantee order of qemu test execution, which should
help make tests reproducible and help debug some test failures.


martinpitt:
First look at the [initial fedora-42 failures](https://github.com/linux-system-roles/logging/actions/runs/14340662724/job/40198825610?pr=439): They look like [this](https://github.com/linux-system-roles/logging/actions/runs/14340662724/job/40198825610?pr=439#step:9:9681):
```
  fatal: [/home/runner/.cache/linux-system-roles/fedora-42.qcow2]: FAILED! => {
      "changed": false
  }
  
  MSG:
  
  Missing log message "testMessage0" in "/var/log/messages"
```

or [here](https://github.com/linux-system-roles/logging/actions/runs/14340662724/job/40198825610?pr=439#step:9:48068):
```
logger: socket /dev/log: No such file or directory
```

F42 does not have `rsyslog` installed by default. systemd/journald have been the main/default logging system for many years at least in Fedora, and I find it a bit annoying that in RHEL everything still gets logged twice by default by installing rsyslog by default.

[This one is different but also interesting](https://github.com/linux-system-roles/logging/actions/runs/14340662724/job/40198825610?pr=439#step:9:32641):
```
          "Problem: problem with installed package\n  - installed package dhcp-client-12:4.4.3-16.P1.fc42.x86_64 requires iproute, but none of the providers can be installed\n  - package dhcp-client-12:4.4.3-16.P1.fc42.x86_64 from fedora requires iproute, but none of the providers can be installed\n  - conflicting requests"
      ],
```

The standard image contains dhcp-client-4.4.3-16.P1.fc42.x86_64 and iproute-6.12.0-3.fc42.x86_64, so at first sight it's not a problem with Fedora.

So this is all certainly fixable, but it's a strategic decision: Does it actually make sense to support this role on Fedora? We could also just declare it "RHEL only" in the README and disable the Fedora tests (also in tmt).

To get some comparison, I'll trigger the tmt tests.

martinpitt:
[citest]

martinpitt:
The Fedora 41 failure in #358 shows the same depsolve errors, so that failure isn't something new. Search for "fatal:" in e.g. [this log](https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/tf_logging-358_Fedora-41-2.17_20250405-081531/artifacts/tests_enabled-ANSIBLE-2.17-test_playbooks_parallel-FAIL.log).

martinpitt:
Some of the f42 tests now pass locally for me with the #440 fix applied. To clear up the debris and look at the remaining failures, I pushed the fix also here in the meantime (we'll rebase it away again after #440 lands, of course). 

However, what is wrong with tmt here? [CentOS-7](https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/tf_logging-439_CentOS-7-latest-2.9_20250409-050014/artifacts/) is shown as failure, but all ansible playbook tests succeeded. It would be really useful to get a link to the actual Testing Farm artifact/run URL, to see pipeline.log. Likewise, the [CentOS 9 failure](https://dl.fedoraproject.org/pub/alt/linuxsystemroles/logs/tf_logging-439_CentOS-Stream-9-2.17_20250409-050015/artifacts) is a 404, but without a TF artifact link it's not possible to investigate why. @spetrosi @richm any idea how to debug these?

martinpitt:
[citest]

martinpitt:
@richm your most recent force push merged the two commits from the previous version of #440 and the github.git changes. I untangled them again, and pushed the updated fix from #440 to give it some more testing.

martinpitt:
@richm I [changed the test logging](https://github.com/linux-system-roles/logging/pull/439/commits/6568b4d34393382c60c2bf5a0acc4587d5eac83a) -- now the [workflow](https://github.com/linux-system-roles/logging/actions/runs/14374926015?pr=439) produces a zip artifact with all logs, and the [Show test log failures step](https://github.com/linux-system-roles/logging/actions/runs/14374926015/job/40305002261?pr=439#step:10:12) only shows the interesting bit -- they are fast and small now.

As for the actual failures, one part is the same "parameter 'workdirectory' specified more than once -" thing I'm already  fighting in #440, plus another one which is specific to the tox tests:

```
  TASK [/home/runner/work/logging/logging/tests/roles/linux-system-roles.logging/roles/rsyslog : Check certs - tls is true, but triplets are not given] ***
  task path: /home/runner/work/logging/logging/roles/rsyslog/tasks/set_certs.yml:39
  Thursday 10 April 2025  07:48:33 +0000 (0:00:00.038)       0:02:58.095 ******** 
  fatal: [/home/runner/.cache/linux-system-roles/fedora-42.qcow2]: FAILED! => {
      "changed": false
  }
  
  MSG:
  
  Error: you specified tls: true; you must specify all 3 of ca_cert, cert, private_key, or all 3 of ca_cert_src, cert_src, private_key_src, or set tls: false in the configuration named elasticsearch_output
  
  TASK [Report error] ************************************************************
  task path: /home/runner/work/logging/logging/tests/tests_files_elasticsearch.yml:439
  Thursday 10 April 2025  07:48:33 +0000 (0:00:00.048)       0:02:58.144 ******** 
  ok: [/home/runner/.cache/linux-system-roles/fedora-42.qcow2] => {}
  
  MSG:
  
  Caught an expected error - {'failed': True, 'msg': 'Error: you specified tls: true; you must specify all 3 of ca_cert, cert, private_key, or all 3 of ca_cert_src, cert_src, private_key_src, or set tls: false in the configuration named elasticsearch_output', '_ansible_no_log': False, 'changed': False}
```

If somebody has an idea about that, I'd appreciate -- but I'll look at the workdirectory thing first.

roles/rsyslog/tasks/main_core.yml
```diff
@@ -10,25 +10,38 @@
...
 
-    - name: Reset original confs - logging package is absent
-      package:
-        name: "{{ __rsyslog_base_packages }}"
+    - name: Clean up old rsyslog main configuration
+      file:
+        path: /etc/rsyslog.conf
+        state: absent
+
+    - name: Scan for old rsyslog configuration drop-ins
```
richm, line 27: The problem here is that other packages may provide /etc/rsyslog.d drop-in files.  For example, cloud-init.  We don't want to remove files provided by other packages, only files provided by the rsyslog* packages.

martinpitt: This is a change from #440 and got addressed there.

PR #441, opened 2025-04-21
ci: use tox-lsr 3.6.0; improve qemu test logging

richm:
tox-lsr 3.6.0 will guarantee order of qemu test execution, which should
help make tests reproducible and help debug test failures.

Improve qemu test logging - this will help debug the qemu test
failures.


martinpitt:
This supersedes #439. That one is itself blocked on #440, unless we accept red tests.

PR #444, opened 2025-04-28
feat: Support this role in container builds

martinpitt:
Feature: Support running the cockpit role during container builds.

Reason: This is particularly useful for building bootc derivative OSes.

Result: These flags enable running the bootc container scenarios in CI, which ensures that the role works in buildah build environment. This allows us to officially support this role for image mode builds.

Do *not* enable the role for system containers (the `container`) flag. `rsyslog.service` is not designed to work in containers, and fails in all RHEL/CentOS and older Fedora releases (it finally works in Fedora 42).

Fixes https://issues.redhat.com/browse/RHEL-88689

---

 - [x] Land initial test and infra, then do final sync: https://github.com/linux-system-roles/sudo/pull/52

### DIFF
--- a/tests/tests_enabled.yml
+++ b/tests/tests_enabled.yml
@@ -22,9 +22,16 @@
   hosts: all
 
   tasks:
+      # rsyslog package may not be installed
+    - name: Check if /etc/rsyslog.d exists
+      stat:
+        path: /etc/rsyslog.d
+      register: rsyslog_d_dir
+
     - name: Check /etc/rsyslog.d
       command: ls /etc/rsyslog.d
       changed_when: false
+      when: rsyslog_d_dir.stat.exists
 
     - name: Default run
       vars:
--- a/tests/tests_server.yml
+++ b/tests/tests_server.yml
@@ -11,6 +11,9 @@
 - name: "Test the server configuration containing tls tcp,
   plain tcp and udp connection"
   hosts: all
+  tags:
+    # certmonger does not work in container builds
+    - tests::booted
   vars:
     __test_cert_name: logging_cert
     __test_ca_cert: "/etc/pki/tls/certs/{{ __test_cert_name }}.crt"
--- a/.github/workflows/ansible-lint.yml
+++ b/.github/workflows/ansible-lint.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Install tox, tox-lsr
         run: |
           set -euxo pipefail
-          pip3 install "git+https://github.com/linux-system-roles/tox-lsr@3.6.0"
+          pip3 install "git+https://github.com/linux-system-roles/tox-lsr@3.7.1"
 
       - name: Convert role to collection format
         id: collection
--- a/.github/workflows/ansible-managed-var-comment.yml
+++ b/.github/workflows/ansible-managed-var-comment.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Install tox, tox-lsr
         run: |
           set -euxo pipefail
-          pip3 install "git+https://github.com/linux-system-roles/tox-lsr@3.6.0"
+          pip3 install "git+https://github.com/linux-system-roles/tox-lsr@3.7.1"
 
       - name: Run ansible-plugin-scan
         run: |
--- a/.github/workflows/ansible-test.yml
+++ b/.github/workflows/ansible-test.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Install tox, tox-lsr
         run: |
           set -euxo pipefail
-          pip3 install "git+https://github.com/linux-system-roles/tox-lsr@3.6.0"
+          pip3 install "git+https://github.com/linux-system-roles/tox-lsr@3.7.1"
 
       - name: Convert role to collection format
         run: |
--- a/.github/workflows/qemu-kvm-integration-tests.yml
+++ b/.github/workflows/qemu-kvm-integration-tests.yml
@@ -74,7 +74,7 @@ jobs:
           python3 -m pip install --upgrade pip
           sudo apt update
           sudo apt install -y --no-install-recommends git ansible-core genisoimage qemu-system-x86
-          pip3 install "git+https://github.com/linux-system-roles/tox-lsr@3.6.0"
+          pip3 install "git+https://github.com/linux-system-roles/tox-lsr@3.7.1"
 
       - name: Configure tox-lsr
         if: steps.check_platform.outputs.supported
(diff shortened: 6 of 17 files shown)


roles/rsyslog/tasks/set_vars.yml
```diff
@@ -7,6 +7,21 @@
   when: __logging_required_facts |
     difference(ansible_facts.keys() | list) | length > 0
 
+- name: Determine if system is booted with systemd
+  when: not __logging_is_booted is defined
```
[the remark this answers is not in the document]
martinpitt: Right. That's the approach that I use in https://github.com/linux-system-roles/cockpit/pull/212 and https://github.com/linux-system-roles/firewall/pull/264 to tell apart "booted" from "some chroot-like environment such as a container build". The `when:` condition specifically here is for caching, as it's not super cheap to determine (similar to `__logging_is_ostree` below).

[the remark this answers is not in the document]
richm: ok - note:
```bash
>podman run --rm -it centos:stream9 bash
bash-5.1# systemctl is-system-running
bash: systemctl: command not found
```
is systemctl always available in `bootc` containers?  do we need to install `systemd` first?
Is there some other way we can determine if we are running in a container, at build-time or at runtime?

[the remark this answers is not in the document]
martinpitt: systemd is always available in bootc base images --- they are intended to be OSes, so they all need an init system. More importantly, a lot of rules would fail if systemd isn't installed, as they also do `service: enable:`. We could perhaps fail with a more explicit error message like "this role only supports bootc base containers" or so, but it seems quite clear already.

We could also check what /proc/1/exe points to, but we don't truly care about that -- we want to know if systemd is running. I'm quite convinced that `systemctl is-system-running` is the right thing to call. We can perhaps rename this variable, like `_runs_systemd`, or `_systemd_booted` or so?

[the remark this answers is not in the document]
richm: so - systemctl executable exists + `systemctl is-system-running` in `['offline', 'degraded']` means "running in a bootc build environment"

I think that's fine.  And I'm not worried about other cases where the user tries to run the role and systemctl is not present.

roles/rsyslog/tasks/set_vars.yml
```diff
@@ -7,6 +7,21 @@
...
+  block:
+    - name: Run systemctl
+      # noqa command-instead-of-module
+      command: systemctl is-system-running
+      register: __is_system_running
+      changed_when: false
+      failed_when: false
```
richm, line 18: I would prefer not to use `failed_when: false` because it masks all errors - would it be possible to do something like `failed_when: not __is_system_running.rc in [0, 1]` assuming that `systemctl is-system-running` returns an error code of `1` when `__is_system_running.stdout` is `offline` or `degraded`?

tests/tests_enabled.yml
```diff
@@ -25,6 +25,8 @@
     - name: Check /etc/rsyslog.d
       command: ls /etc/rsyslog.d
       changed_when: false
+      # rsyslog package may not be installed
+      failed_when: false
```
richm, line 29: would prefer not to use `failed_when: false`

tests/tests_server.yml
```diff
@@ -11,6 +11,9 @@
 - name: "Test the server configuration containing tls tcp,
   plain tcp and udp connection"
   hosts: all
+  tags:
+    # certmonger does not work in container builds
```
richm, line 15: ok for now - but we need to think about what it means to use the certificate role at container build time

"As a user deploying services configured at container build time, I want to configure my services to be HTTPS servers with certs and private keys such that the services start serving HTTPS at first boot without additional configuration or file retrieval."

How to get the certs/keys placed on the system for this at bootc build time?  One way would be for the user to get the cert/key files, outside of the role/build process, and place them in the build directory and use `COPY mycert.crt /etc/tls/certs/....` in the Containerfile.  Hopefully there is a better way.

roles/rsyslog/tasks/set_vars.yml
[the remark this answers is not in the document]
martinpitt: I added an extra error message, looks like this:

```
TASK [/var/home/martin/upstream/lsr/logging/tests/roles/linux-system-roles.logging/roles/rsyslog : Require installed systemd] ******************
task path: /var/home/martin/upstream/lsr/logging/roles/rsyslog/tasks/set_vars.yml:20
redirecting (type: connection) ansible.builtin.buildah to containers.podman.buildah
fatal: [sut]: FAILED! => {"changed": false, "msg": "Error: This role requires systemd to be installed."}
```

roles/rsyslog/tasks/set_vars.yml
martinpitt: Unfortunately not -- the [manpage](https://www.man7.org/linux/man-pages/man1/systemctl.1.html#:~:text=is-system-running%20output) only guarantees "exit code > 0". It seems to be 0 *right now*, but TBH I'd rather not rely on that.

tests/tests_enabled.yml
martinpitt: Heh, for Ansible I need to unlearn "don't do TOCTOU" :grin: I added a separate stat check instead.

tests/tests_server.yml
martinpitt: I'm not a big fan of that "copy static certs" approach. We intentionally use certmonger to have automatic refresh set up, and these days certs only last 1 year. Also, such secrets don't belong into container images, which tend to be copied around registries and deployed to multiple systems.

 I agree we need to fix this properly, that requires some design and coding work in the certificate role, and testing a bootc build in QEMU.

I reworked this to be more explicit: There is now an error message when you try to build a certificate during container build:

```
TASK [linux-system-roles.logging : Certificates are only supported in a booted system] *********************************************************
task path: /var/home/martin/upstream/lsr/logging/tests/roles/linux-system-roles.logging/tasks/certificate.yml:2
fatal: [sut]: FAILED! => {"changed": false, "msg": "The logging_certificates option is only supported in booted systems, not container builds."}
```
and a new test to cover this.

This approach gradually enables the role on bootc, until we figure out certificates.

richm (approved):

tests/tests_server.yml
richm: > Also, such secrets don't belong into container images, which tend to be copied around registries and deployed to multiple systems.

True - container runtimes like kubernetes and podman have special ways to handle secrets so that you do not have to build them into the image, and they can be safely and securely used by the deployed images.

plus - secrets like this will be specific to *instances* of the image, not the image itself - so it really doesn't make sense to create a host specific secret like a x509 cert/key in the image build.

PR #499, opened 2026-04-24
feat: add role fingerprints to syslog

richm:
Feature: Add a fingerprint string to the system log to indicate when the role began
successfully, and when the role finished successfully.  The fingerprint string indicates
the role name, a timestamp, and the platform.

Reason: Users can see when the role was used and if it was used successfully.  This
information from the system log can be collected by log scanners and aggregators
for further analysis.

Result: The role logs fingerprints to the system log.

This also adds a test to check if the fingerprints were written upon a successful
role invocation.

## Summary by Sourcery

Add role fingerprint logging to syslog and verify it via system journal tests.

New Features:
- Introduce sr_fingerprint Ansible module to write fingerprint messages with timestamps to syslog.
- Record begin and success fingerprint messages for the logging system role including Ansible version and OS distribution metadata.

Enhancements:
- Ensure required Ansible facts are gathered before emitting role fingerprint messages.
- Add sanity ignore entries and a role-specific library directory to support the custom module across Ansible versions.

Tests:
- Extend default test playbook to verify that syslog/journal contains the expected begin and success fingerprints when the role runs successfully.


richm:
[citest]

